PR #8, opened 2025-08-08
draft init flow, component definition and responsibilities

maleck13:
Adding diagram outline the request flow for MCP init calls to the gateway. Also an attempt at components and responsibilities that I am sure will shift as we evolve


maleck13:
Will layer on the tool call flow which may impact on this initial diagram

esnible (approved):

PR #9, opened 2025-08-08
Add initial structure

david-martin:


huang195 (commented):
@david-martin for the mcp-broker, instead of a hello-world http-server, should we use a hello-world mcp server? It would be able to at least handle initialize calls, and we can leave the session management stuff for later

huang195:
Let's get this merged so we have an initial structure for the repo. We can work on the rest as we go

PR #13, opened 2025-08-11
simple MCP server for testing

esnible:
A simple MCP server based on https://github.com/modelcontextprotocol/go-sdk

Provides four tools
- The "hi" tool from the library sample
- A "time" tool that returns the current time
- A "slow" tool that waits N seconds, notifying the client of progress
- A "headers" tool that returns all HTTP headers it received


huang195 (changes requested):
@esnible is it possible to give a few examples about how to test for slow response and headers, and their expected response in the README.md file?

PR #14, opened 2025-08-11
MCP test server based on https://github.com/mark3labs/mcp-go

esnible:
- MCP test server based on https://github.com/mark3labs/mcp-go
- fix for progress notification on the other Go test server


esnible:
> this is the same as test server1 but using a different mcp client library?

yes

PR #15, opened 2025-08-12
add some more attempts at flows

maleck13:
Added some attempt at what the Auth flow could look like with Authorino in the mix. These have not been proven out yet. 
Goal here is prompt discussions and thought on some of these flows


esnible (approved):
These flows are good enough to get discussion started.  It makes sense to merge them rather polish this PR.
The first diagram shows the /init going to both the MCP Router and MCP Broker, but the second one doesn't.
It would be good to show the second flow to @maia-iyer to see if the security story fits into the plans she is working on for Kagenti.

Issue #6, opened 2025-08-07
Small set of test MCP servers


esnible:
- One using github.com/modelcontextprotocol/go-sdk/mcp
- One using github.com/mark3labs/mcp-go/mcp
- Python one using import FastMCP

Some servers with long-running tools.

- At least one tool with a static response
- At least one tool with an argument (add, or echo)
- At least one tool with no args (time)
- At least one long-running tool (outputs percentage, with input be the number of inputs)
- At least two tools with the same name

Security ideas
- One tool that needs an API key
- One tool that needs an OAuth token

Document the existing test servers

PR #19, opened 2025-08-13
Test server written in Python

esnible:
For #6


esnible:
> @esnible I see the removal of the messages about not seeing long lived connection in MCP Inspector. Does it mean you fixed it?

Yes.  https://github.com/kagenti/mcp-gateway/pull/14 fixed the notification problem by using the progressToken correctly.  The documentation was not fully fixed in pull 14.

PR #31, opened 2025-08-19
Add poc code, with placeholders, for ext-proc impl

### LINKED ISSUE #3: ext-proc routing code
Copy key code from other repo, and leave TODOs and placeholders for other issues.

david-martin:
Closes #3 

This is a drop in of the poc code in https://github.com/david-martin/mcp-helper/tree/main/ext-proc,
with any hardcoded server prefix/tool mappings removed (as the poc only allowed for 2 hardcoded mcp servers).
The hardcoded logic has been replaced with placeholders for now.

The intention is to have this be a base for adding in proper logic for server/tool mapping when the discovery piece is added,
and serve as a base for where session header mapping logic will land.


jasonmadigan:
I think this is mergeable as-is, thanks @david-martin

jasonmadigan (approved):

PR #32, opened 2025-08-19
Add pre-commit GitHub Action to run go vet on PRs

sbekkerm:
This PR introduces a new GitHub Action workflow named Pre-commit, which runs on all pull requests.

The workflow performs the following:
- Checks out the repository
- Sets up Go 1.24
- Runs `go vet ./...` to catch potential issues in the codebase


esnible:
I see this has already been merged.  I would have requested that we run `make lint`, which already does the `go vet`.

PR #35, opened 2025-08-19
initial dev env w/ kind

jasonmadigan:
Re: https://github.com/kagenti/mcp-gateway/issues/1

work in progress

based on some of Craig's work in https://github.com/david-martin/mcp-helper/pull/6

TODO:

 - [x] local dev loop: i.e. Run binaries locally on host (ext_proc/router + broker), w/ resources in-kind cluster

later:

 - replace mock-mcp with the ones from `tests/servers`
 - images + deployments for the broker/router (images tackled in the CI issue)

For commands, see `make help`


esnible:
I have been deploying _kind_ clusters using the Kagenti demo code https://github.com/kagenti/kagenti/blob/main/docs/demos.md which also installs a lot of pieces.  It might be worthwhile to consider aligning with that demo and using that demo's installer.

Makefile
```diff
@@ -55,4 +69,103 @@ lint: fmt vet golangci-lint
 test-unit:
 	echo "blah blah blah test"
 
-
+.PHONY: local-env-setup
```
R-Lawton, line 72: do you think we should check to see if the tools are installed here and run the tool make targets if there not like kustomise yq etc. i didnt have them installed in my bin and it failed. Could add it to the make info or help too as a fyi

R-Lawton:
im getting a status 500
```
curl -i -X POST \
                -H "Host: mcp.example.com" \
                -H "Content-Type: application/json" \
                -d '{"jsonrpc":"2.0","method":"initialize","params":{},"id":1}' \
                http://localhost:8888/mcp
HTTP/1.1 500 Internal Server Error
date: Thu, 21 Aug 2025 09:55:26 GMT
server: istio-envoy
connection: close
content-length: 0
```

Steps ive taken:
1. make local-env-setup
2. make dev
3. make run-router (sep terminal) 
4. make run-broker (sep terminal) 
5. make dev-gateway-forward (sep terminal) 
6. make dev-test (sep terminal)

jasonmadigan:
@R-Lawton I don't think the broker is fleshed out yet, so you'd expect a 500 (https://github.com/kagenti/mcp-gateway/blob/main/cmd/mcp-broker/main.go#L12)

this'll be fixed over time, main aim for this PR is just to deploy stuff in whatever state things are in right now (incomplete at the mo)

jasonmadigan:
@maleck13 prob close enough to merge something like this you reckon? more to do of course, but can do in further PRs (this is getting large now)

jasonmadigan:
OK let's merge for now, and keep at it.

PR #36, opened 2025-08-19
Add calls from gateway to wasm

david-martin:
A few small extra lines to show where the wasm bit gets kicked off from (i.e. from the gateway)


pdettori (approved):
/lgtm

PR #45, opened 2025-08-20
Moving mcp router into mcp broker to create the one binary

R-Lawton:
MCP router and broker are in the one binary now. Updated make files to reflect this, also removed duplicate linters.

Can be ran with `run-mcp-broker-router`


twghu (approved):
Some comments on code, but as a base this is ok

R-Lawton:
> * .github/workflows/code-style.yaml , andstor/file-existence-action current version is 5

Could you explain a little bit more im not fully following

> .github/workflows/images.yaml
> * does the `find` on line 47 need some error check?

Do you mean in code style? Thats the only find i could see. but yes i can add error check to that

> * .github/workflows/tests.yaml
>   * go version mismatch  in `unit-tests`
>   * will we need testing on other platforms

Spotted that,  should be fixed in another pr i created 

> * cmd/mcp-broker-router/main.go
>   * Missing grpcSrv.GracefulStop()
>   * Env vars need to be sanitized before use
>   * health check endpoints
> * internal/mcp-router/response.go
>   * general comment, we should ensure any injected text (body for example)
>     is not echoed out in log files (security).
> * internal/mcp-router/server.go
>   * for loop in Process only exits on read or send error, are these tests sufficient to exit loop?

All very valid but im hesitant to change the code to much as part of this pr, i think when work on these files gets going properly would be a better place.

PR #47, opened 2025-08-21
Initial Broker interface

### LINKED ISSUE #5: Session ID mapping - where will it be handled?
Document the flow for

- intialize
- initialized
- tools/list
-

### LINKED ISSUE #34: MCP Broker: Discovery and store list of tools per MCP Server
Related to #33 

This issue is for making the necessary calls to each newly discovery MCP server to get a list of available tools, and storing that list in the server to tools mapping.

See in the poc how this was done for a hardcoded list of servers & using the mcp-go library.
https://github.com/david-martin/mcp-helper/blob/96e4c64108c5c3e7f9edb32b92a6fcac2335eb10/main.go#L449-L461

esnible:
Resolves #5
Resolves #34

This offers a simple interface to register and unregister MCP servers.  Please pick it apart so that we are good enough to start building upon it.

There is also an implementation that is good enough to start with, that actually works, but can be improved later.

Currently, there are no mocks and the tests don't run unless an environment variable is defined.  To run the tests:

- Follow https://github.com/kagenti/mcp-gateway/tree/main/tests/servers/server1 to create test image
- `docker tag mcp-test1 mcp-test1:v1`
- `kind load docker-image mcp-test1:v1 --name mcp-gateway`
- `oc -n mcp-system run mcp-test1 --image=mcp-test1:v1 -- /mcp-test-server --http localhost:9090`
- `oc -n mcp-system port-forward mcp-test1 9090`
- `RUN_TESTS=TRUE go test ./...`

I think this code is good enough to merge, and we can take care of testing concerns and refactoring later.  I am creating it as a draft because I want everyone to get a chance to tell me if I am following our coding practices correctly.

I would appreciate advice on how we would like the test servers to run.  Unless I here otherwise, I will refactor test-mcp1 into the main code so that we can run it as part of unit tests, and run that server every time.


twghu:
Maybe the 
- `upstreamMCP`  could have "host" rather than "upstreamMCP" duplicated name.  
- "upstreamSessionState" "initialized" would this be more readable as "isIntialized"  an "isReady" func that tests isInitialized and session not null.
- Places where a "host" is used would benefit from hostname sanitizing.   In a final version we could have a white/black list configured from xDS?
- "Register/UnregisterServer" "upstreamMCPHost(mcpHost)" has multiple use.
- "CallTool" could use helper func for session creation to separate concerns.
- "Close" overwrites errors, these errors could be added to "var errorMessages [] string" for example so the possible original error can be tracked.
- The various maps should have synchronization guards?
- An error from "discoverTools" in "RegisterServer" could leave session open?

twghu:
Given this is base initial version it seems ok to merge for a base, noting the code comments ^^.

PR #76, opened 2025-08-27
Caching package

R-Lawton:
Adding caching package that adds a:
*  check if MCP session id exists based on key on mcp server name and gateway session id. 
* function type to be used by broker initalisation functionality to create  mcp client initalisation request if session id is not found and will add mcp session to map when found

### DIFF
--- a/internal/cache/session-caching.go
+++ b/internal/cache/session-caching.go
@@ -1,3 +1,4 @@
+// Package cache provides session caching functionality for gateway session IDs and MCP session IDs.
 package cache
 
 import (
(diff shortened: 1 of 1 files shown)


internal/cache/session-caching.go
```diff
@@ -0,0 +1,66 @@
...
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	// Double-check if it was created while waiting
+	if sessionID, exists := c.sessions[k]; exists {
+		return sessionID, nil
+	}
+
+	// Create new session
+	sessionID, err := c.initSession(ctx, serverName, gwSessionID)
```
maleck13, line 51: Interesting. I see why you added this. But not sure on this approach. Is this function making a network call to the  MCP backend? If there is an issue here we might be locking the cache for quite a long time poss until a timeout is reached for example with a bad MCP Server or a MCP server under load. This could become a bottleneck.

The scenario where the mcp session can be overwritten is the following I believe:

- client does init get gateway session but has no mcp-session-id
- client does tools list 
- client invokes multiple tools from the **same** MCP server concurrently

I think we can capture this is a potential issue and discuss how we want to resolve. 

off the top of my head, I think we might want to queue the concurrent requests (responding with 409 if too many) for that client (ie gateway-session) until a mcp-session-id was retrieved by the first request and stored. Only the read and write to the cache should lock the cache so allowing all other requests to quickly pass through and not get held up. Once a session is  present  then use that session on the remaining queued requests to complete the calls.   This way on the specific client sees any slow down and not all clients. 

Its late here so perhaps my understanding is wrong let me know what your thoughts are

maleck13: Looking at https://pkg.go.dev/sync#Map

R-Lawton:
the golint is very strict right now, complaining about things not i my pr.Jason has a pr to fix this either we can merge this with the lint failing or wait until his pr is merged https://github.com/kagenti/mcp-gateway/pull/75

maleck13 (approved):
Approving to allow us to move forward, but we need to capture that initializing sessions in the cache package should be removed . With the sync.Map I don't think there is any need for it to happen as it does in the cache package and can happen as part of the router tools call

PR #81, opened 2025-08-28
Fixes to build errors

twghu:
The `Wandalen/wretry.action@v3.8.0` doesn't seem to work with docker/build-push-action@v5

The issue was that retry action is having trouble parsing the multiline YAML values (like the `labels` output from the metadata step). The simplest solution is to avoid using the retry wrapper altogether for the Docker build step, since Docker builds are generally reliable and the retry adds unnecessary complexity. `wretry.action`

Updated golang version and fixed Dockerfile.


twghu:
changes out of date

PR #83, opened 2025-08-28
Broker MCP listener endpoint

### LINKED ISSUE #51: MCP Broker: handle tools/list call
**What**

Add an endpoint that will handle the the tools/list call from the MCP clients 

In the POC, this was handled by the mcp-go library as aggregated tools were registered with the library
https://github.com/david-martin/mcp-helper/blob/96e4c64108c5c3e7f9edb32b92a6fcac2335eb10/main.go#L476-L489

esnible:
Resolves #51

This creates an MCP server which does everything including listing federated tools.  It is essentially a federated MCP without Envoy.  The idea is that its tools/list endpoint is exactly what is needed for the higher performance Envoy version.  I am not sure if we also want the broker to offer the list of tools without MCP initialization as a pure REST endpoint.

There are currently no unit tests; that will come in a follow-up.

To test, start a test server, e.g. `MCP_TRANSPORT=http PORT=9090 go run internal/tests/server2/main/main.go`

To create a configuration so that the broker uses the test server, create a _/tmp/config.yaml_ file 

```YAML
servers:
  - name: test
    url: http://localhost:9090/mcp
    toolPrefix: t2
    enabled: true
    hostname: t2.example.com
```

To run the broker, do `go run cmd/mcp-broker-router/main.go -mcp-broker-address localhost:8081 -mcp-gateway-config /tmp/config.yaml`

To interactively use the Broker as an MCP server, do `DANGEROUSLY_OMIT_AUTH=true npx @modelcontextprotocol/inspector` and connect to http://localhost:8081/mcp

### DIFF
--- a/internal/broker/broker.go
+++ b/internal/broker/broker.go
@@ -159,13 +159,19 @@ func (m *mcpBrokerImpl) RegisterServer(
 		slog.Info("Federating tool", "mcpHost", mcpHost, "federated name", newTool.Name)
 		tools = append(tools, server.ServerTool{
 			Tool: newTool,
+			Handler: func(_ context.Context, _ mcp.CallToolRequest) (*mcp.CallToolResult, error) {
+				return mcp.NewToolResultError("Kagenti MCP Broker doesn't forward tool calls"), nil
+			},
+			/* UNCOMMENT THIS TO TURN THE BROKER INTO A STAND-ALONE GATEWAY
 			Handler: func(ctx context.Context, request mcp.CallToolRequest) (*mcp.CallToolResult, error) {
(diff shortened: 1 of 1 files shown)


esnible:
mcp-inspector-on-broker

jasonmadigan (approved):
`kubectl -n mcp-system get configmap mcp-gateway-config -o jsonpath='{.data.config\.yaml}' > /tmp/config.yaml`

tested w/ controller config, seems to work good. may need to look at options for with/without ingress of cluster for hostnames and stuff, but does work.

think this is mergeable as-is

Issue #62, opened 2025-08-26
Initialize mcp server for client if there's no upstream mcp server session id


david-martin:
In the router.

Depends on https://github.com/kagenti/mcp-gateway/issues/49 first

pseudocode

```
mcpSession, err := cache.lookUpSession(serverName, gatewaySession)(string, error)
if isNotFound(err){
   mcpSession, err := mcpClient.initializeMCP(serverName)(string, error)
   cache.StoreSession(serverName, gatewaySession, mcpSession)
}
```

In the poc, the [mcp-go library was used](https://github.com/david-martin/mcp-helper/blob/96e4c64108c5c3e7f9edb32b92a6fcac2335eb10/main.go#L355-L371), and a client connection kept around.
For this issue, we should *not* keep the client around, just the session ID.

esnible:
This logic is implemented in https://github.com/kagenti/mcp-gateway/blob/main/internal/broker/broker.go#L157

When the Broker is asked to invoke a call, it initializes if there is no existing upstream session ID for the current session ID.

Issue #49, opened 2025-08-25
MCP Router: Add routing logic to MCP Router ext_proc service


maleck13:
**What**

Component:

- MCP Router

parsing the body
setting the `authority:` header based on the host in the config (created mcp controller)
setting the `x-mcp-tool` header
calling cache to get a MCP-session-id
initializing the MCP server on behalf of the user if no session id for that server
storing the MCP-session-id in the cache
- set the `mcp-session-id` header
- remove prefix from tool name in the json rpc payload
- allow the request to continue 

As part of this work we should add a config that will work locally:

```yaml 
  apiVersion: v1
  kind: ConfigMap
  metadata:
    name: mcp-gateway-config
    namespace: mcp-system
    labels:
      app: mcp-gateway
  data:
    config.yaml: |
      logLevel: info
      port: 8080
      servers:
      - name: mcp-system/weather-route
        url: http://weather-service.mcp-system.svc.cluster.local:8080
        hostname: weather.example.com
        toolPrefix: weather_
        enabled: true
      - name: mcp-system/calendar-route
        url: http://calendar-service.mcp-system.svc.cluster.local:8080
        hostname: calendar.example.com
        toolPrefix: cal_
        enabled: true
      - name: kb-system/kbase-route
        url: http://kbase-service.kb-system.svc.cluster.local:8080
        hostname: kbase.example.com
        toolPrefix: ai_  # Used default prefix
        enabled: true
```

is the current way it looks based on @jasonmadigan work

PR #86, opened 2025-08-28
 Gh 49 Routing piece

R-Lawton:
#62 & #49 

### Verification
Set up a local kind cluster with the Broker, Router & Controller running.
These components are built during the make target into a single image and loaded into the cluster.
Also sets up an Istio Gateway API Gateway, and HTTPRoutes for test mcp servers, which are added to the broker/router.

```bash
make local-env-setup
```

Port forward to the gateway

```bash
make dev-gateway-forward
```

Run the mcp-inspector and connect to the gateway

```bash
DANGEROUSLY_OMIT_AUTH=true npx @modelcontextprotocol/inspector
```

Open http://localhost:6274/?transport=streamable-http&serverUrl=http://mcp.127-0-0-1.sslip.io:8888/mcp

You should be able to make tool calls for the available tools.


david-martin:
Steps to verify locally have been added the readme. 
Posting them here as well:

Set up a local kind cluster with the Broker, Router & Controller running.
These components are built during the make target into a single image and loaded into the cluster.
Also sets up an Istio Gateway API Gateway, and HTTPRoutes for test mcp servers, which are added to the broker/router.

```bash
make local-env-setup
```

Port forward to the gateway

```bash
make dev-gateway-forward
```

Run the mcp-inspector and connect to the gateway

```bash
DANGEROUSLY_OMIT_AUTH=true npx @modelcontextprotocol/inspector
```

Open http://localhost:6274/?transport=streamable-http&serverUrl=http://mcp.127-0-0-1.sslip.io:8888/mcp

You should be able to make tool calls for the available tools.


An issue will be created for server3 as that's failing to init at this time.

README.md
```diff
@@ -4,49 +4,31 @@ An Envoy-based gateway for Model Context Protocol (MCP) servers, enabling aggreg
...
-make dev-stop               # Stop local processes
-make local-env-teardown     # Destroy cluster
+make dev-gateway-forward
 ```
 
-Run `make help` to see all available commands.
+Run the mcp-inspector and connect to the gateway
+
+```bash
+DANGEROUSLY_OMIT_AUTH=true npx @modelcontextprotocol/inspector
```
jasonmadigan, line 28: can probably just run one of the inspect.mk targets

config/mcp-system/config.yaml
```diff
@@ -0,0 +1,16 @@
+servers:
```
jasonmadigan, line 1: Prob don't need this:

`make dev-broker-local` from the other PR will grab config from cluster & rewrite:

https://github.com/kagenti/mcp-gateway/pull/90/files#diff-a42ef31db17cb47677b726741c26c6ffb6afdd5af936255d6744d92f6d97c5c7R69

config/mcp-system/broker-service.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: v1
+kind: Service
```
jasonmadigan, line 2: already exists no?

https://github.com/kagenti/mcp-gateway/blob/main/config/mcp-system/service.yaml#L3

jasonmadigan: The existing service https://github.com/kagenti/mcp-gateway/blob/main/config/mcp-system/service.yaml#L6 already serves both the broker (port 80->8080) and router (50051). The HTTPRoute in this PR could just reference mcp-gateway on port 80 instead of creating a new mcp-broker service, similar to how   https://github.com/kagenti/mcp-gateway/blob/main/config/istio/gateway/httproute.yaml#L17-L18 does it. does that make sense?

jasonmadigan (approved):

PR #91, opened 2025-09-01
MCPGateway -> MCPServer - resource rename

jasonmadigan:
`MCPGateway` -> `MCPServer` for clarity.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -72,7 +72,7 @@ deploy-example: install-crd ## Deploy example MCPServer resource
 	@echo "Waiting for test servers to be ready..."
 	@kubectl wait --for=condition=ready pod -n mcp-test -l app=mcp-test-server1 --timeout=60s
 	@kubectl wait --for=condition=ready pod -n mcp-test -l app=mcp-test-server2 --timeout=60s
-	@kubectl wait --for=condition=ready pod -n mcp-test -l app=mcp-test-server3 --timeout=60s
+	@kubectl wait --for=condition=ready pod -n mcp-test -l app=mcp-test-server3 --timeout=90s
 	@echo "All test servers ready, deploying MCPServer resource..."
 	kubectl apply -f config/samples/mcpserver-test-servers.yaml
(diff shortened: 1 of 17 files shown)


esnible:
The renaming seems fine.

It would be good if this PR or a follow-up could document what an MCPServer is.  A paragraph or two in a README and also `kubectl explain MCPServer` could provide an explanation.

jasonmadigan:
> It would be good if this PR or a follow-up could document what an MCPServer is. A paragraph or two in a README and also kubectl explain MCPServer could provide an explanation.

sounds reasonable, will add some changes here to `describe` & the README

jasonmadigan:
Added some notes to the README on the CRD, and added some descriptions to the CRD:

```bash
kubectl explain MCPServer
GROUP:      mcp.kagenti.com
KIND:       MCPServer
VERSION:    v1alpha1

DESCRIPTION:
    MCPServer defines a collection of MCP (Model Context Protocol) servers to be
    aggregated by the gateway. It enables discovery and federation of tools from
    multiple backend MCP servers through HTTPRoute references.

FIELDS:
  apiVersion	<string>
    APIVersion defines the versioned schema of this representation of an object.
    Servers should convert recognized schemas to the latest internal value, and
    may reject unrecognized values. More info:
    https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#resources

  kind	<string>
    Kind is a string value representing the REST resource this object
    represents. Servers may infer this from the endpoint the client submits
    requests to. Cannot be updated. In CamelCase. More info:
    https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#types-kinds

  metadata	<ObjectMeta>
    Standard object's metadata. More info:
    https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#metadata

  spec	<Object>
    Spec defines the desired state of the MCPServer resource

  status	<Object>
    Status represents the observed state of the MCPServer resource
```

PR #92, opened 2025-09-01
Add well-known oauth resource server and keylcoak auth server interation

david-martin:
Ability to use keycloak via an AuthPolicy to protect the mcp-gateway via the OAuth Resource Server and Authorization Server mechanisms as per https://modelcontextprotocol.io/specification/2025-06-18/basic/authorization

TODO:

- [ ] make target for setting up as much as possible in a local dev environment
- [ ] remove hardcoding of well-known endpoint response in mcp-broker - probably pull in config from MCPServer CRD
- [ ] AuthPolicy validation of token on mcp endpoint (currently denying all traffic with 401 hardcoded)
- [ ] Configure and check scope(s)/claims in AuthPolicy
- [ ] Avoid hardcoded rewrite in the keycloak HTTPRoute. Needed until https://github.com/keycloak/keycloak/issues/40923 is solved, but could use a regex in an EnvoyFilter for now.

Steps to try out:

- Follow the [quickstart in the README](https://github.com/kagenti/mcp-gateway?tab=readme-ov-file#quick-start-with-mcp-inspector)
- Expose keycloak `kubectl apply -f ./config/keycloak/httproute.yaml`
- Login as admin/admin to the admin realm at http://keycloak.127-0-0-1.sslip.io:8888/ and set up a new realm, user and anonymous client registration
  - Create a new realm called 'mcp'
  - Create a new user called 'mcp`, with password `mcp` in the new realm
  - From 'Clients' > 'Client Registration' > 'Anonymous access polices' - delete the 'Trusted Hosts' policy
- Create the AuthPolicy `kubectl apply -f ./config/mcp-system/authpolicy.yaml`
- From mcp-inspector UI with the mcp server url set as http://mcp.127-0-0-1.sslip.io:8888/mcp, go through the OAuth flow.


config/istio/gateway/oauth-envoyfilter.yaml
```diff
@@ -0,0 +1,46 @@
...
+      operation: INSERT_FIRST
+      value:
+        name: mcp-well-known-direct
+        match:
+          path: "/.well-known/oauth-protected-resource/mcp"
+        direct_response:
+          status: 200
+          body:
+            inline_string: >
+              {"resource_name":"MCP Server","resource":"http://mcp.127-0-0-1.sslip.io:8888/mcp","authorization_servers":["http://keycloak.127-0-0-1.sslip.io:8888/realms/mcp"],"bearer_methods_supported": ["header"],"scopes_supported":["basic"]}
```
maleck13, line 27: This is pretty ugly. Kinda would prefer it hard coded in the broker than here tbh.  Do we have a follow up issue to capture the required configuration and aggregation of the Oauth values so that it can be properly handled by the broker?  This isn't going to work for kagentai as it is

david-martin: Yes, I agree it's pretty ugly.
At this stage I would prefer it outside the broker than hardcoded in there. 
At least if it's outside (in an EnvoyFilter), it can be tweaked more easily until functionality is added to pass in this config to serve it via the broker.

> Do we have a follow up issue

Once this PR lands, i'll capture what's needed in an issue to bring in this response from broker config.

> This isn't going to work for kagentai as it is

Why is that? I would think it's at least configurable compared to a hardcoded response.
Maybe it doesn't matter if auth is enabled in kagenti only after the follow up?

README.md
```diff
@@ -24,6 +24,27 @@ make inspect-gateway
 
 Open http://localhost:6274/?transport=streamable-http&serverUrl=http://mcp.127-0-0-1.sslip.io:8888/mcp
 
+## Example OAuth setup
```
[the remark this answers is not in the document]
david-martin: As it stands, this section follows on from the quickstart, so it's all local.
I'd need to give it more thought how examples could be made useful outside the local dev environment, if that's what you're getting.

maleck13 (approved):
This works well. As I am capturing the auth in a doc I will add some issues based on that to make the configuring of external auth servers and the MCP Response.

PR #99, opened 2025-09-02
Fixes to work with Podman

esnible:
I needed to do these things to get the environment to set up with Podman.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,6 +8,8 @@ ifeq ($(ARCH),aarch64)
     ARCH = arm64
 endif
 
+# Use `export BUILD_FLAGS=--load` for Podman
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -8,6 +8,8 @@ ifeq ($(ARCH),aarch64)
     ARCH = arm64
 endif
 
+# Use `export BUILD_FLAGS=--load` for Podman
+BUILDFLAGS = $(BUILD_FLAGS)
```
david-martin, line 12: fyi @maleck13 I know you had issues with this flag not being present

david-martin (approved):
Thanks for fixing this @esnible

PR #111, opened 2025-09-03
Remove individual tool prefix from targetRefs to reduce complexity

jasonmadigan:
After a chat with @maleck13 

- Removed individual tool prefix for now form individual servers in `targetRefs[]`
- Prefix can be set for all servers in an `MCPServer`

If we think we want this again, we can add back later.


huang195 (approved):
/lgtm. Yeah, I like this change. It makes things easier to use, and if needed, we can add per-tool prefix back in.

PR #116, opened 2025-09-04
handle zero mcpserver resource at startup time

huang195:
The current implementation seems to only handle the case where MCPServer resources are all deleted, and it would then write an empty Configmap. However, if there is zero MCPServer to start with, it doesn't write an empty Configmap. This PR fixes this use case.


jasonmadigan (approved):
lgtm

PR #118, opened 2025-09-04
add different MCPServer resources

maleck13:


jasonmadigan (approved):
gets the cow out of that ditch while we think a bit more in https://github.com/kagenti/mcp-gateway/issues/117

PR #131, opened 2025-09-05
Allow configuring oauth-protected-resource response via env vars

### LINKED ISSUE #129: Allow configuring oauth-protected-resource response in mcp-broker via env var
Currently, the mcp-broker has no built-in way to serve a /.well-known/oauth-protected-resource response. For proof of concept we [hacked this with an EnvoyFilter](https://github.com/kagenti/mcp-gateway/blob/a6bb549b6c351507deb3a110c7b39de1cbc82fc7/config/istio/gateway/oauth-envoyfilter.yaml) returning a static direct response, but this needs to be configurable at runtime so that the MCP Gateway can front multiple backend MCP servers under a single auth provider (e.g. Keycloak).

The proposal is to add support in mcp-broker for serving a JSON document at /.well-known/oauth-protected-resource, with the contents configured via an environment variable. This allows users to configure the broker’s OAuth discovery response without code changes or Envoy patching.

Example minimal response shape:
```json
{
  "resource_name": "MCP Server",
  "resource": "http://mcp.127-0-0-1.sslip.io:8888/mcp",
  "authorization_servers": [
    "http://keycloak.127-0-0-1.sslip.io:8888/realms/mcp"
  ],
  "bearer_methods_supported": ["header"],
  "scopes_supported": ["basic"]
}
```

Some considerations:

* The MCPServer resource already describes multiple backend servers (via the mcp-controller → mcp-broker/mcp-router), but this discovery response is about fronting all of them through one OAuth entrypoint.
* At minimum, the env var should allow configuring:
    * authorization_servers (list of issuer URLs, likely just 1 in most setups)
    * resource (the URL of the broker/gateway endpoint, e.g. /mcp)
* Defaults can be applied for bearer_methods_supported and scopes_supported, unless operators want to override.
* Implementation could parse a JSON string from an env var and serve it directly on /.well-known/oauth-protected-resource.

This feature would remove the need for manual EnvoyFilter hacks and make the broker deployable with different Keycloak realms or OIDC providers in a clean, configurable way.
The [example auth setup](https://github.com/kagenti/mcp-gateway/blob/a6bb549b6c351507deb3a110c7b39de1cbc82fc7/README.md#example-oauth-setup) in the README should be updated accordingly as part of this change.

david-martin:
Closes #129 

Steps to try this out have been updated in the readme.
The previous method that uses an EnvoyFilter have been removed.


README.md
```diff
@@ -26,15 +26,35 @@ Open http://localhost:6274/?transport=streamable-http&serverUrl=http://mcp.127-0
...
 kubectl apply -f ./config/mcp-system/authpolicy.yaml
 ```
 
+The mcp-broker now serves OAuth discovery information at `/.well-known/oauth-protected-resource`.
+
+### Keycloak Setup
+
 Set up a new 'mcp' realm in keycloak:
 
 * Open http://keycloak.127-0-0-1.sslip.io:8888/
```
jasonmadigan, line 60: not sure if it's just me, but I got some weird embiggened UI that looked like some resources like CSS, e.g. 500s from http://keycloak.127-0-0-1.sslip.io:8888/resources/0ypou/admin/keycloak.v2/assets/main-DmoViYol.css

To get around, ran:

`make keycloak-forward`

and instead went to http://localhost:8095/

jasonmadigan (approved):
Worked good, but did see some funnies with loading keycloak assets. It seemed to be intermittent when using the http://keycloak.127-0-0-1.sslip.io:8888 URL

README.md
jasonmadigan: (appreciate that isn't part of this PR btw)

README.md
```diff
@@ -26,15 +26,35 @@ Open http://localhost:6274/?transport=streamable-http&serverUrl=http://mcp.127-0
...
+
+# Apply AuthPolicy for token validation
 kubectl apply -f ./config/mcp-system/authpolicy.yaml
 ```
 
+The mcp-broker now serves OAuth discovery information at `/.well-known/oauth-protected-resource`.
+
+### Keycloak Setup
+
 Set up a new 'mcp' realm in keycloak:
```
jasonmadigan, line 58: wonder if we could/should automate this in setup for demo-ability.

david-martin: I'll look into options for having a script to do this

README.md
jasonmadigan: yeah, I noticed with a few refreshes, it goes back to normal. very weird.

david-martin:
@jasonmadigan would you mind giving this another spin.
I've updated the readme as there's now a make target to configure the keycloak realm and user.

Note the keycloak port has changed to 8889 instead of 8888 when forwarding to the gateway. However you shouldn't really have to visit it directly unless you're debugging something now (but you do get redirected there during the oauth flow, so port forwarding is still important for oauth flow to work).

The problem with the 500's was that all requests through the Gateway (including keycloak) were going through the ext-proc. The switch to a different listener port (8081) for the keycloak listener allows us to easily only enable the ext-proc for the mcp listeners.
When forwarding, it's still 8888 for 8080, but now also 8889 for 8081.

jasonmadigan:
👀

PR #132, opened 2025-09-05
Rename internal type to reflect meaning

esnible:
PR to rename an internal type from `upstreamMCPHost` to `upstreamMCPURL`, because that type is used to hold URLs, not hostnames.

A parameter is also renamed for the same reason.

This is only renaming, no logic has changed.


huang195 (approved):
/lgtm no logic change

PR #135, opened 2025-09-05
Option for broker to send per-upstream Authorization API key

esnible:
Related #139
For https://github.com/kagenti/mcp-gateway/issues/69#issuecomment-3257644837

This allows the Broker to supply per-upstream API key values when initializing and listing tools.

The configuration of the Broker is done via an env var rather than the config file.  This is for security.

This code is good enough, but it could perhaps be improved by using JSON format for the server-to-apikey relationship.  Or perhaps another implementation.

To test, start test server 2 using the instructions in https://github.com/kagenti/mcp-gateway/pull/133
Then, start the system using

```bash
cat <<EOF | > /tmp/config.yaml
servers:
  - name: test
    url: http://localhost:9091/mcp
    toolPrefix: t2
    enabled: true
    hostname: t2.example.com
EOF

KAGENTAI_test_CRED="Bearer 1234" go run cmd/mcp-broker-router/main.go -mcp-broker-address localhost:8081 -mcp-gateway-config /tmp/config.yaml
```

The idea is that when the Broker sees the upstream http://localhost:9091/mcp it looks it up in the env variable and gets the Authorization header value `Bearer 1234`.

### DIFF
--- a/internal/broker/broker.go
+++ b/internal/broker/broker.go
@@ -4,8 +4,11 @@ package broker
 import (
 	"context"
 	"fmt"
+	"log"
 	"log/slog"
+	"os"
 	"slices"
+	"strings"
 	"time"
 
 	"github.com/kagenti/mcp-gateway/internal/config"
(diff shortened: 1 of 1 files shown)


internal/broker/broker.go
```diff
@@ -429,3 +441,37 @@ func (m *mcpBrokerImpl) Close(_ context.Context, downstreamSession downstreamSes
...
+	for _, auth := range strings.Split(authorizations, "+") {
+		authParts := strings.SplitN(auth, "=", 2)
+		if len(authParts) != 2 {
+			// It would be better to print the incorrect fragment, but we don't want
+			// to log a possibly important bearer key
+			log.Printf("Invalid auth encoding (%d parts)", len(authParts))
+			continue
+		}
+
+		if authParts[0] == string(upstream.upstreamMCP) {
```
maleck13, line 469: I wonder would it make sense to use the MCP name from the config as the key here?

maleck13 (approved):
@jasonmadigan if you begin looking at loading creds via a secret from the controller take a look at this to know how to structure the secret

PR #152, opened 2025-09-10
MCPServer: support credentials for MCP servers in CRD

jasonmadigan:
Re: https://github.com/kagenti/mcp-gateway/issues/139

Implements authentication support for `MCPServer` resources, allowing the broker to authenticate with upstream servers that require credentials.

### Some changes

- Authentication tokens passed via `Authorization: <token>` header to upstream servers (e.g. `Bearer 1234`)
- new `credentialRef`:
```
credentialRef:
  key: token
  name: server4-auth-credentials
```
- Credentials are injected via env vars with this pattern: `KAGENTAI_{MCP_NAME}_CRED`
- Broker deployment uses `envFrom` with `secretRef` to bulk inject all credentials
- Controller aggregates credentials into a `mcp-aggregated-credentials` secret
- Added a "server 4" test server with authentication support (`EXPECTED_AUTH`) - it's server 2 with auth (from https://github.com/kagenti/mcp-gateway/pull/133)
  - Tools with from server 4 (`s4_auth_`) prefix require Bearer token authentication, which the broker will provide to the downstream mcp server
- Note: existing tools work without authentication (i.e. `s1`, `s2` etc)

Bugs/other stuff:

- Was using `strings.Split(toolName, "_")[0]` which failed for multi-underscore prefixes like `s4_auth_` changed to `strings.HasPrefix()`, any prefix should be fine now
- A bug (?) that I went around in circles on: `setUpRouter()` was creating its own broker instance, causing issues with session lookups
- Router adds auth headers (tool calls bypass the broker)

## verifying

Deploy:

`make local-env-setup`

Check creds:
```bash
# ConfigMap has credentialEnvVar key (not value) for server4
kubectl get configmap -n mcp-system mcp-gateway-config -o yaml | grep credentialEnvVar

# env-var set in broker env
kubectl exec -n mcp-system deployment/mcp-broker-router -- printenv | grep KAGENTAI
```

Test with MCP Inspector:

`make inspect-gateway`

Connect to the broker at http://mcp.127-0-0-1.sslip.io:8888/mcp & `List Tools`

In the MCP Inspector UI:

1. Test non-authenticated tool:
   Call: test3_add with arguments a: 5, b: 3
   Expected: Returns 8
2. Test authenticated tool:
   Call: test4_hello_world with name: "test"
   Expected: Returns "Hello, test!"

Logs to check:

```bash
# router is adding auth headers
kubectl logs -n mcp-system deployment/mcp-broker-router | grep "Adding Authorization header"
# should see: "Adding Authorization header for routing server=mcp-test/mcp-server4-route"

# check server4 is receiving authenticated requests
kubectl logs -f -n mcp-test deployment/mcp-test-server4
# you will see an initial rejection like Rejecting "Bearer test-server4-secret-token", got ""
# this is during tool discovery, which doesn't use auth
# After you call the test4_hello_world tool, you should not see rejection
# you should see "Processing tools/call request"
```

## Appendix

Found it useful to verify that the upstream server is actually configured with auth.

Verify test server 4 is _actually_ using auth:

Open a tunnel directly to server 4 (no gateway):

`make inspect-server4`

In MCP Inspector:

1. Connect to http://localhost:9093/mcp
    You'll see a bunch of `Rejecting "Bearer test-server4-secret-token", got ""`  messages in the test server 4 logs.
2. Configure Authentication > Bearer Token: `test-server4-secret-token`
3. Reconnect
    Init and tool list / calls work

## examples

```yaml
apiVersion: mcp.kagenti.com/v1alpha1
kind: MCPServer
metadata:
  name: test-server4
  namespace: mcp-test
spec:
  credentialRef:
    key: token
    name: server4-auth-credentials
  targetRef:
    kind: HTTPRoute
    name: mcp-server4-route
  toolPrefix: test4_
```

```yaml
apiVersion: v1
data:
  token: Bearer test-server4-secret-token
kind: Secret
metadata:
  name: server4-auth-credentials
  namespace: mcp-test
type: Opaque
```


Makefile
```diff
@@ -141,6 +143,74 @@ golangci-lint:
 .PHONY: lint
 lint: fmt vet golangci-lint
 
+# Code style checks (matches CI)
+.PHONY: check-style
+check-style: check-gofmt check-goimports check-newlines
+
+.PHONY: check-gofmt
```
jasonmadigan, line 150: > I could pull this out into a separate PR if it's a bit noisy.

this too

build/inspect.mk
```diff
@@ -22,54 +22,39 @@ urls-impl:
 .PHONY: inspect-broker
 inspect-broker: inspect-gateway
 
-# Open MCP Inspector for mock server implementation
```
jasonmadigan, line 25: > I could pull this out into a separate PR if it's a bit noisy.
and all of this

PR #167, opened 2025-09-11
Auth phase 1 draft

maleck13:
I haven't yet proven out these interactions. But I wanted to share the ideas


docs/design/auth.md
```diff
@@ -0,0 +1,108 @@
...
+        - email
+        - role
+        - user
+      rbac:
+        - role: accounting
+          tools: 
+            - "payroll"
+        - role: admin
+          tools: 
+            -"*"
```
guicassolato, line 56: If this info can be compiled dynamically from multiple (hypothetical) config files of each tool, then perhaps we want here some kind of selector of the tools/config files themselves instead, as opposed to duplicating the info into the MCPServer resource?

On the other hand, if it's a static list (that occasionally may diverge from the actual tools running/being served behind), then I wonder if the info doesn't belong into the AuthPolicy.

In fact, why having an `auth` section in the `MCPServer` resource et all? Couldn't this (shouldn't this?) all be in the AuthPolicy? Unless we're thinking on hiding the creation/ownership of the AuthPolicy resource behind a MCPServer one as a high level API?

docs/design/auth.md
```diff
@@ -0,0 +1,108 @@
...
+
+```
+
+> Note In a future phase these defined scopes may be used to define a token exchange to scope down the token to only the required scopes. 
+
+### Registering MCP Auth Severs
+
+Part of the response from the `oauth-protected-resource` endpoint is an array of `authorization_servers`. For phase 1 the gateway will only support a single auth server. The expectation is each of the MCPServers using OAuth is using the same auth server and will independantly validate the token against that auth server. The MCP Broker will be configured with this shared server separately via a flag or EnvVar.
+
+> Note this doesn't include URL elicitation where an MCP server may request a user provide it access to a resource not protected by the defined auth server. Example a github resource/API. This is consisdered seaparate from the general Gateway authentication and authorization requirement.
```
guicassolato, line 79: This could be achieved using [**Keycloak Authorization Services**](https://www.keycloak.org/docs/latest/authorization_services/index.html).

Example in an AuthConfig/AuthPolicy: https://github.com/Kuadrant/authorino/blob/main/docs/user-guides/keycloak-authorization-services.md

guicassolato (commented):
I'm sure I'm missing something, but my general feeling is that we may here be overcomplicating the auth setup a little, with steps for the broker and router that may not be not needed given a tool like Authorino has been pointed.

From my POV, scope could be reduced to:
1. store use role bindings ("role mappings") in Keycloak – they will be added to the JWT automatically by the auth server;
2. push the static `tool → required roles` config into the AuthPolicy and write a pattern matching authorisation rule that calculates the intersection of it against the role bindings retrieved from the JWT – this could be a CEL expression or using OPA in case the list of tools is too big;
3. enable caching in the auth service;
4. (Optional) for more limited token permissions, set some default client scopes in Keycloak, so that varying the client ID results in a filtered list of user role bindings inside the JWT at the source.

docs/design/auth.md
```diff
@@ -0,0 +1,108 @@
...
+
+### Protecting the MCP Gateway HTTP endpoint(s)
+
+In order to enforce authenticated access, the gateway listeners exposing the MCP gateway to clients, can use a kuadrant `AuthPolicy` resource. This resource can define what authentication must be met and also enforce Authz requirements at the gateway (Authorization Covered Below).
+
+[Example AuthPolicy](./../../config/mcp-system/authpolicy.yaml)
+
+[Example Oauth Setup](../../README.md#example-oauth-setup)
+
+With this policy in place any unauthenticated request not going to the /.well-known endpoints will get a 401 with a resource_metadata url specified. It is then up to the client to leverage this information to retrieve the resource metadata to know how to authenticate to access the resource (example resource would be /mcp). This resource metadata will be served via the MCP Broker component.
```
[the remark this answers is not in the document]
david-martin: > This will make all requests to go through whenever it passes authentication

yeah, this was very much intentional to just do an authN check (no authZ) with that example.
We'll need to evolve the expression for authZ when we work through a good example e2e with keycloak where scopes ultimately map to some permissions on tool calls.

> Authorino will automatically inject the authentication challenge header in the following format

Am I reading into this too much, or does authorino have some built in support for https://datatracker.ietf.org/doc/html/rfc9728#name-www-authenticate-response to satisfy https://modelcontextprotocol.io/specification/2025-06-18/basic/authorization#authorization-server-location, 
so you wouldn't have to add something like this to the AuthPolicy?

```yaml
        unauthenticated:
          code: 401
          headers:
            'WWW-Authenticate':
              value: Bearer resource_metadata=http://mcp.127-0-0-1.sslip.io:8888/.well-known/oauth-protected-resource/mcp
          body:
            value: |
              {
                "error": "Forbidden",
                "message": "Access denied."
              }
```

docs/design/auth.md
david-martin: Just confirming, your comment is about the elicitation text, not the text above that for the `oauth-protected-resource` endpoint?

For elicitation (or avoiding it), if I'm understanding what you're suggesting, there would be backchannel communication between authorino and the 3rd party server to request and get some level of access on behalf of the user, but without the user (or the client) being part of that flow?

docs/design/auth.md
```diff
@@ -0,0 +1,108 @@
...
+      - ext_auth_data
+```                  
+
+
+Note> We will use AuthPolicy to enforce the tool access check as a reference of how you can achieve this integration with the gateway and Kuadrant. 
+
+
+### Filtering of tools
+
+The broker, when receiving a tools/list call, will validate the JWT token if present in the request, and load the RBAC data from the config. As it has access to the JWT it can identify which tools should be returned based on this configuration.
```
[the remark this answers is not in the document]
david-martin: So this comes down to if the filtering of the tool list should be part of the 'mcp gateway',
or should the 'mcp gateway' support the ability to allow something to filter that list response per user.

The allow/deny of an individual tool call per user feels like its firmly in the realm of an auth integration that the mcp-gateway supports, not part of the mcp-gateway itself.
However, when filtering the response body, it's a little different than an allow/deny filter. That's where I'm hesitating. 
I'm also aware of not introducing hard dependency on kuadrant for any of the basic mcp-gateway functionality.
So an important question here is if filtering of the tool list response is basic functionality.
If not, then how do we allow plugging in something that can do that filtering (including kuadrant as an example) in a decoupled manner.

docs/design/auth.md
guicassolato: Yes, the elicitation part.

In the example I shared (Authorino user guide), the backchannel is for Authorino to open a ticket with the auth server, on behalf of the requestor, for the owner of the resource to decide on authorisation. The owner of the resource still needs to go to the Keycloak UI and grant/deny the requested access. If access is granted, then the request can be repeated and it goes through.

There are a few caveats. One is that the resource needs to be tracked by the auth server, with the proper ownership association stored there. From my experience, that's easily said than done, though not impossible.

Another caveat, at least with the example I shared, is that there's no support for Keycloak Authorization Services built-in in a Kuadrant AuthPolicy or Authorino AuthConfig today. The example was crafted using OPA, leaking some client secrets into the spec of the policy. That's not ideal.

maleck13:
Example implementation https://github.com/kagenti/mcp-gateway/pull/196/files

docs/design/auth.md
```diff
@@ -0,0 +1,108 @@
...
+### Authenticated Calls
+
+Once a client has obtained a token, it can then make requests to the MCP Gateway. When a request comes to the gateway, the Kuadrant WASM plugin intercepts this request and based on configuration, will decide whether or not to call to the Authorino component. With the linked AuthPolicy, Authorino will receive the request and then validate the token with the configured auth server before allowing the request to continue. 
+
+> Note not covered here is message signinging as specified under https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1415 . This will be covered at a later phase. 
+
+
+### Authorization of tools/calls
+
+When the router receives a tools/call. It will load from the config the RBAC section for the specified MCP server. It will then set this data into the dynamic metadata via the `ext_auth_data` namespace. Optionally we could also encode this data and set it as a request header. The reason for this is to allow downstream filters to access this information to make authorization decisions.
```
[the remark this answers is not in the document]
maleck13: I have done this now. AuthPolicy requests this metadata from a separate "server" in the example it is just a static file but obv could be anything that Authorino can interact with https://github.com/kagenti/mcp-gateway/pull/196/files#diff-80c5f77bc4ee52cd5426f9b1fda44c143c58084313203dc47354e59ecc5cddbeR17

docs/design/auth.md
```diff
@@ -0,0 +1,108 @@
+## Auth with the MCP Gateway (Phase 1)
+
+To provide auth intetgration, the MCP gateway uses (Kuadrant)[https://kuadrant.io] and Gateway API in combination with Keycloak as an identity provider.
```
esnible, line 3:
```suggestion
To provide auth integration, the MCP gateway uses (Kuadrant)[https://kuadrant.io] and theGateway API in combination with Keycloak as an identity provider.
```

docs/design/auth.md
```diff
@@ -0,0 +1,108 @@
...
+> Note while we use keycloak in our examples, in theory other IDPs could also be used.
+
+
+### Authentication Implementation
+
+![](./images/MCPgateway-auth.jpg)
+
+### Protecting the MCP Gateway HTTP endpoint(s)
+
+In order to enforce authenticated access, the gateway listeners exposing the MCP gateway to clients, can use a kuadrant `AuthPolicy` resource. This resource can define what authentication must be met and also enforce Authz requirements at the gateway (Authorization Covered Below).
```
esnible, line 16:
```suggestion
In order to enforce authenticated access, the gateway listeners exposing the MCP gateway to clients, can use a Kuadrant `AuthPolicy` resource. This resource can define what authentication must be met and also enforce AuthZ requirements at the gateway (Authorization covered below).
```

docs/design/auth.md
esnible, line 22:
```suggestion
With this policy in place any unauthenticated request not going to the /.well-known endpoints will get a 401 with a resource_metadata URL specified. It is then up to the client to leverage this information to retrieve the resource metadata to know how to authenticate to access the resource (example resource would be /mcp). This resource metadata will be served via the MCP Broker component. 
```

docs/design/auth.md
```diff
@@ -0,0 +1,108 @@
...
+  "scopes_supported":["email","role","user"]
+}
+
+```
+
+> Note In a future phase these defined scopes may be used to define a token exchange to scope down the token to only the required scopes. 
+
+### Registering MCP Auth Severs
+
+Part of the response from the `oauth-protected-resource` endpoint is an array of `authorization_servers`. For phase 1 the gateway will only support a single auth server. The expectation is each of the MCPServers using OAuth is using the same auth server and will independantly validate the token against that auth server. The MCP Broker will be configured with this shared server separately via a flag or EnvVar.
```
esnible, line 77:
```suggestion
Part of the response from the `oauth-protected-resource` endpoint is an array of `authorization_servers`. For phase 1 the gateway will only support a single auth server. The expectation is that each of the MCPServers using OAuth are using the same auth server and will independently validate the token against that auth server. The MCP Broker will be configured with this shared server separately via a flag or EnvVar.
```

docs/design/auth.md
```diff
@@ -1,6 +1,6 @@
 ## Auth with the MCP Gateway (Phase 1)
 
-To provide auth intetgration, the MCP gateway uses (Kuadrant)[https://kuadrant.io] and Gateway API in combination with Keycloak as an identity provider.
+To provide auth integration, the MCP gateway uses (Kuadrant)[https://kuadrant.io] and theGateway API in combination with Keycloak as an identity provider.
```
esnible, line 3:
```suggestion
To provide auth integration, the MCP gateway uses (Kuadrant)[https://kuadrant.io] and the Gateway API in combination with Keycloak as an identity provider.
```

docs/design/auth.md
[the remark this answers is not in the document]
maleck13: going to remove tool filtering from this and move to phase 2

docs/design/auth.md
maleck13: So we have 2 options we can show. One the ACL server and 2 via keycloak. I am moving tool filtering to phase 2 doc

david-martin (approved):
@maleck13 this is good to merge.

Let's capture issues for anything we haven't provided yet in examples, like fetching from vault or using keycloak resource roles as an acl.

maleck13:
I believe we have it all here https://github.com/orgs/kagenti/projects/6/views/1?filterQuery=-type%3AFeature+label%3Aauth

Issue #53, opened 2025-08-25
MCP Broker: Notifications support for tools/list_changed


david-martin:
https://modelcontextprotocol.io/specification/2025-06-18/server/tools#list-changed-notification

As part of server discovery (#21), an init call is made to each mcp server (followed by a tools list).
For streaming/sse support, the client is expected to make a GET /mcp 'notifications/initialized' call
https://modelcontextprotocol.io/specification/2025-06-18/basic/lifecycle, which is a long running sse connection where communication happens in one direction only (server to client).

There are a number of different notifications that can be sent from the server to the client. 
However, to minimise the scope and focus on getting a working flow,  only 'notifications/tools/list_changed' will be  supported first. This notification doesn't need to be client/session aware. That is, all clients are interested in this notification. This reduces the complexity of multiplexing sse connections right now.

A flow diagram would be good to add to https://github.com/kagenti/mcp-gateway/blob/main/docs/design/flows.md to better understand the components at play.

david-martin:
@esnible I like the idea of having endpoints to modify the tool list.
In #17, I have a use case where the tool name changes.
That would be good to have as well

jasonmadigan:
Parking this for a bit while looking at https://github.com/kagenti/mcp-gateway/issues/89

david-martin:
@esnible how is progress on this issue?
Do you need any help or additional guidance on it?

david-martin:
Capturing some of our discussion from standup yesterday.

> I dropped this Issue for now. I now notice this Issue is tagged MVP. How important is it? Do we know of any popular MCP servers that add and remove tools over their lifetime?

Notifications and streaming support is an important part of MCP as it allows for server to client comms for things like tool list changes or progress updates in a long running tool call.
Although there isn't a hard use case for this yet, priority wise it was something we thought important enough to make some headway on early, and build on it to add eventual full support (i.e. the whole story #24 )
So it is not a blocker for the MVP milestone.

> My thought is that the broker is using the mark3labs library correctly, but until we implement Issue https://github.com/kagenti/mcp-gateway/issues/20 we can't connect the Broker's tool change notifications through the gateway. What do you think?

The original thinking on this issue was to implement a slimmed down version of that for a 'broadcast' approach to tools list changed notifications. For any connected clients, they all get that notification as it's not client specific.
The more permanent solution is definitely captured in #20, but perhaps what we do here could be a building block?

[...]

PR #182, opened 2025-09-15
fix: make toolPrefix immutable

jasonmadigan:
re: https://github.com/kagenti/mcp-gateway/issues/150

- Makes `toolPrefix` immutable
- `controller-gen` regen of CRDs + rename (mcpserver was originally hand-carved)


david-martin (approved):
I'm in favour of this as it prevents accidental breaking of a client due to tool name changes, while also simplifying reconcile logic.
That being said, tool name changes due to the tool name itself changing in the mcp backend (not the prefix) is still something we will handle, but separately in #53 

Looking at the reconcile function in https://github.com/kagenti/mcp-gateway/blob/main/pkg/controller/mcpserver_controller.go, confirming that logic doesn't need to change as the validation check is enough to prevent prefix changes getting through. After that, only creates will get through where the prefix is being used.

This all seems reasonable.
I'm aware of some ad-hoc discussions about generating a toolPrefix.
That can progress to it's conclusion separate from this PR, which fixes/prevents some potentially annoying problems that exist already.

PR #186, opened 2025-09-16
doc: Support External MCP Servers

jasonmadigan:
- Add a guide on connecting `mcp-gateway` with external MCP servers, using Gateway API and Istio
- Controller now detects `ExternalName` services and use their external hostnames for routing instead of internal Kube service DNS names

Refs:

- https://github.com/Kuadrant/kuadrant-operator/blob/db8e2d1747cfcd4be69769d2a627bdfd5256814c/doc/user-guides/misc/external-api.md
- https://istio.io/latest/docs/tasks/traffic-management/egress/egress-control/

### DIFF
--- a/docs/guides/external-mcp-server.md
+++ b/docs/guides/external-mcp-server.md
@@ -51,9 +51,9 @@ export GITHUB_PAT="ghp_YOUR_GITHUB_TOKEN_HERE"
 Before proceeding, verify all dependencies are met:
 
 ```bash
-# MCP Gateway is deployed
-kubectl get mcpservers -A
+# Verify MCP Gateway components are deployed
 kubectl get pods -n mcp-system
+kubectl get gateway -n gateway-system mcp-gateway
 
 # verify egress to GitHub's MCP server from a test pod
 kubectl run -it --rm debug --image=nicolaka/netshoot --restart=Never -- \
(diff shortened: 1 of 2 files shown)


pkg/controller/mcpserver_controller.go
```diff
@@ -383,12 +407,27 @@ func (r *MCPReconciler) discoverServersFromHTTPRoutes(
 		}
 	}
 
-	// to think about: service vs ingress via GW API
+	// external services use https on port 443
+	if isExternal && backendRef.Port != nil && *backendRef.Port == 443 {
```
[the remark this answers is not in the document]
jasonmadigan: right to question. that is better. went with `appProtocol` from the [service](https://kubernetes.io/docs/concepts/services-networking/service/)

docs/guides/external-mcp-server.md
```diff
@@ -0,0 +1,299 @@
...
+### Network Requirements
+- Egress access from your cluster to `api.githubcopilot.com` (port 443)
+
+## Verify Prerequisites
+
+Before proceeding, verify all dependencies are met:
+
+```bash
+# MCP Gateway is deployed
+kubectl get mcpservers -A
```
huang195, line 55: why do we want to run this command?

docs/guides/external-mcp-server.md
```diff
@@ -0,0 +1,299 @@
...
+        "namespaces": {
+          "from": "All"
+        }
+      }
+    }
+  }
+]'
+```
+
+## Step 2: Create ExternalName Service
```
huang195, line 103: Are the 7 steps here corresponding to the 7 steps above? If so, step 2 is supposed to be creating the HTTPRoute.

docs/guides/external-mcp-server.md
```diff
@@ -0,0 +1,299 @@
...
+# Delete the MCPServer and Istio resources
+kubectl delete mcpserver github -n mcp-test
+kubectl delete httproute github-mcp-external -n mcp-test
+kubectl delete service api-githubcopilot-com -n mcp-test
+kubectl delete serviceentry github-mcp-external -n mcp-test
+kubectl delete destinationrule github-mcp-external -n mcp-test
+kubectl delete secret github-token -n mcp-test
+
+# Remove the listener from the Gateway
+kubectl get gateway mcp-gateway -n gateway-system -o json | \
```
huang195, line 296: I wonder if we could make the gateway more permissive by default. It looks like for every external host we support, we would need to add new listeners to the gateway, and when we're done, we remove it. This is more error prone than adding/removing other types of objects like Httproutes.

PR #190, opened 2025-09-16
Close MCP server on shutdown

esnible:
This causes our server to explicitly shut down the MCP server implementation.

In interactive testing, this seems to cause shutdowns of uncontacted MCP servers by a few seconds.

This code isn't really needed now, as we always exit, but it is good practice to shut down the MCP server, and we should keep this as an example if we refactor the code so that `setUpBroker()` is called more than once during the lifetime of our server.


huang195 (approved):
/lgtm

Issue #195, opened 2025-09-17
Keycloak helm chart updates


jasonmadigan:
Suffering from a rolling [brownout](https://github.com/keycloak/keycloak/issues/37636) of the KC helm chart's images.

Look to switch to an alternative.

Upstream [keycloak issue](https://github.com/keycloak/keycloak/issues/37636), but no consensus it looks like on if/when they'll have a helm chart.

PR #196, opened 2025-09-17
added config endpoint to get acl . added two separate auth policies

maleck13:
**What**

This adds a few key things

1) a config/{id} endpoint
2) two separate authpolicies  (one for authnz access to none tool calls and one specific for tool calls)
3) a change to the EnvoyFilter to ensure it is called before the WASM plugin

**Try it**

**EDIT**
See README changes.

Then Follow OAuth flow

You should be able to initialize and do a tools list and see all tools
you should only be able to use the test2_headers and test_headers tools

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,7 +12,7 @@ COPY pkg/ pkg/
 
 RUN CGO_ENABLED=0 GOOS=linux go build -o mcp_gateway cmd/mcp-broker-router/main.go
 
-FROM alpine:latest
+FROM alpine:3.22.1
 
 RUN apk --no-cache add ca-certificates
 


david-martin:
got it working locally.
few things:

- the bitnamilegacy image change didn't work. for now i commented out the pull of keycloak & postgres image as i had them cached locally. #195 should fix that
- had to `make inspect-gateway` to get port forwarding up before i could run `make keycloak-setup-mcp-realm` and access the keycloak UI & API
- saw the keycloak css issue has returned. Looks like the portnumber was removed from the EnvoyFilter. Comment above.
- had to `kubectl apply -f ./config/mcp-system/tool-call-auth.yaml` for the tool call authZ check

I'm wondering about the tool name in the acl having the prefix. 
That may become an issue if that prefix is generated in future.
I don't have a good suggestion how to overcome this yet.

I'll push up an update to the `keycloak-setup-mcp-realm` target to set up the client scopes and mapper.

Makefile
```diff
@@ -317,8 +317,16 @@ istioctl: ## Download and install istioctl
 	@$(MAKE) -s -f build/istio.mk istioctl-impl
 
 .PHONY: keycloak-install
+<<<<<<< HEAD
```
guicassolato, line 320: ☝️

config/mcp-system/authpolicy.yaml
```diff
@@ -2,24 +2,27 @@ apiVersion: kuadrant.io/v1
...
-  namespace: mcp-system
+  namespace: gateway-system
 spec:
   targetRef:
     group: gateway.networking.k8s.io
-    kind: HTTPRoute
-    name: mcp-route
+    kind: Gateway
+    name: mcp-gateway
+    sectionName: mcp
```
maleck13, line 11: important this is applying only to the mcp route (IE the gateway)

config/example-access-control/deployment.yaml
```diff
@@ -0,0 +1,25 @@
...
+    matchLabels:
+      app: config-server
+  template:
+    metadata:
+      labels:
+        app: config-server
+    spec:
+      containers:
+        - name: server
+          image: quay.io/maleck13/config_server:latest
```
david-martin, line 18: where's the source of this?

config/istio/envoyfilter.yaml
```diff
@@ -19,7 +19,7 @@ spec:
               subFilter:
                 name: 'envoy.filters.http.router'
       patch:
-        operation: INSERT_BEFORE
+        operation: INSERT_FIRST
```
david-martin, line 22: there's another file when deploying the broker on your host rather than in cluster that needs updating too in config/dev/envoyfilter.yaml

david-martin:
Something has changed since yesterday morning and the oauth flow isn't working for me anymore from fresh setup.
I'm trying to automate the keycloak setup, but the manual setup isn't working either now.
Failing with a 405 on the mcp inspector calling http://keycloak.127-0-0-1.sslip.io:8889/realms/mcp/clients-registrations/openid-connect at the Client Registration phase of oauth flow.
I'll keep digging to understand why this is.

david-martin:
@maleck13 I've pushed up another update.
The steps to try this out have been simplified.

Now it's the quickstart steps (like before)

```
make local-env-setup
make inspect-gateway
```

followed by a new target that sets up keycloak and all other auth stuff

```
make oauth-example-setup
```

Good to go after that.
README has been updated with this info.

I've also pushed up an update to the AuthPolicy cel for the group tool check (as mentioned on chat)

cmd/mcp-broker-router/main.go
```diff
@@ -307,7 +307,8 @@ func setUpConfigServer(address string) *http.Server {
 	}
 
 	configHandler := broker.NewConfigUpdateHandler(mcpConfig, authToken, logger)
-	mux.Handle("POST /config", configHandler)
+	mux.HandleFunc("POST /config", configHandler.UpdateConfig)
+	mux.HandleFunc("GET /config/{serverID}", configHandler.GetConfig)
```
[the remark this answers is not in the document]
maleck13: will remove no need to really keep it

Makefile
```diff
@@ -329,6 +329,44 @@ keycloak-forward: ## Port forward Keycloak to localhost:8090
...
+	@echo "Prerequisites: make local-env-setup should be completed"
+	@echo ""
+	@echo "Step 1: Setting up Keycloak realm..."
+	@$(MAKE) -s -f build/keycloak.mk keycloak-setup-mcp-realm
+	@echo ""
+	@echo "Step 2: Configuring mcp-broker with OAuth environment variables..."
+	@kubectl set env deployment/mcp-broker-router \
+		OAUTH_RESOURCE_NAME="MCP Server" \
+		OAUTH_RESOURCE="http://mcp.127-0-0-1.sslip.io:8888/mcp" \
+		OAUTH_AUTHORIZATION_SERVERS="http://keycloak.127-0-0-1.sslip.io:8889/realms/mcp" \
```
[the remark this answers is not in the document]
maleck13: I think this is now in main not in the make file. But I agree we could default them

internal/broker/config_handler.go
```diff
@@ -28,8 +29,33 @@ func NewConfigUpdateHandler(cfg *config.MCPServersConfig, authToken string, logg
...
+		}
+	}
+	if server == nil {
+		w.WriteHeader(404)
+		return
+	}
+
+	enc := json.NewEncoder(w)
+	w.Header().Add("content-type", "application/json")
+	if err := enc.Encode(server); err != nil {
```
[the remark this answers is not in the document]
maleck13: removed entirely

esnible (approved):

david-martin (commented):
Verified all again locally 👍

PR #202, opened 2025-09-18
updating MCPServer to add a new path field

jasonmadigan:
Re: https://github.com/kagenti/mcp-gateway/issues/124

Adds support for custom MCP endpoint paths in `MCPServer` resources, allowing servers to use non-standard paths instead of the default `/mcp`.

### Some changes

- New optional `path` field in MCPServer spec to specify custom MCP endpoint path
- Default remains `/mcp`, when path is not specified
- Controller includes path in ConfigMap for broker configuration
- Broker uses custom path when connecting to upstream servers
- Router correctly routes requests to custom path endpoints
- Adds a new test server with a custom path
- We weren't building some of the new test servers, added to CI
- Fixed an e2e test flake (I think). The test scanned logs to confirm things, but because of that SSE bug spamming the logs, that became unreliable. Switched the test to use the new /status endpoint, which will be much more reliable.

### Example

```yaml
apiVersion: mcp.kagenti.com/v1alpha1
kind: MCPServer
metadata:
  name: custom-path-server
  namespace: mcp-test
spec:
  path: /v1/special/mcp  # custom path instead of default /mcp
  toolPrefix: custompath_
  targetRef:
    kind: HTTPRoute
    name: mcp-custom-path-server
```

## Verifying

Deploy:
```bash
make local-env-setup
```

Check custom path in ConfigMap:
```bash
kubectl get configmap -n mcp-system mcp-gateway-config -o yaml | grep path
# url: http://mcp-custom-path-server.mcp-test.svc.cluster.local:8080/v1/special/mcp
```

Test with MCP Inspector:
```bash
make inspect-gateway
```

Connect to broker and verify tools from custom path server are available and are callable with the `custompath_` prefix.

### Additional verification

Connect directly to the new test server with a custom path:

```bash
make inspect-custom-path
```


jasonmadigan:
@david-martin couple of suggested fixes

PR #223, opened 2025-09-26
remove invalid session for session cache

R-Lawton:


tests/servers/server1/main.go
```diff
@@ -252,3 +263,29 @@ func handleEmbeddedResource(
...
+	if r.Method == "POST" {
+		body, err := io.ReadAll(r.Body)
+		if err == nil && bytes.Contains(body, []byte("always_404")) {
+			log.Printf("Intercepting always_404 tool call - returning HTTP 404")
+			r.Body = io.NopCloser(bytes.NewReader(body))
+
+			sessionID := r.Header.Get("mcp-session-id")
+			if sessionID != "" {
+				w.Header().Set("mcp-session-id", sessionID)
+			}
```
david-martin, line 282: Is setting the session id in a response header for debugging purposes?
Looking at the protocol spec, https://modelcontextprotocol.io/specification/2025-06-18/basic/transports#session-management, the 404 status code is the only 'must' `after which it MUST respond to requests containing that session ID with HTTP 404 Not Found`.

I think it's fine if it doesn't cause any problems though

internal/mcp-router/response_test.go
```diff
@@ -0,0 +1,85 @@
...
+					RawValue: []byte("test-mcp-session-123"),
+				},
+			},
+		},
+	}
+
+	// Test that 404 detection works
+	resp, err := server.HandleResponseHeaders(responseHeaders)
+	require.NoError(t, err)
+	require.Len(t, resp, 1)
```
maleck13, line 59: doesn't it always return at least one response? Should we check something about the response like if it has a the mcp-session-id not set etc?

internal/mcp-router/response_test.go
```diff
@@ -0,0 +1,85 @@
...
+}
+
+func TestInvalidateByMCPSessionID(t *testing.T) {
```
maleck13, line 62: Not sure I am understanding this test well. Seems like whether the session exists or not we would get the same result, so not really testing that the session exists or was created ?

PR #224, opened 2025-09-26
token exchange example (auth phase 2)

### LINKED ISSUE #137: Support OAuth token exchange for upstream MCP servers in same trust domain
When an upstream MCP server is protected by OAuth (e.g. Keycloak) and shares the same trust domain as the MCP Gateway, the broker should support exchanging the client’s token for a new token scoped to the upstream MCP server.

* The gateway validates the inbound token.
* If the backend MCP server requires different scopes, the gateway performs an [RFC 8693 Token Exchange](https://datatracker.ietf.org/doc/html/rfc8693?utm_source=chatgpt.com)
 against the same authorization server (e.g. Keycloak).
* The new token is then attached to the upstream request (Authorization: Bearer <token>).

This allows fine-grained scoping per upstream server (server1:read, server2:read) while the client only needs a generic scope (e.g. gateway:access).

Notes:

* Implementation may reuse existing JWT validation logic, in AuthPolicy, if possible
* Requires configuration on the MCPServer resource to specify the scope required for that upstream.
* First step: support only same trust domain (same Keycloak realm).

guicassolato:
Add example resources for implementing token exchange between the gateway and targeted MCP server.

- Based on Keycloak's [Standard token exchange](https://www.keycloak.org/securing-apps/token-exchange#_standard-token-exchange).
- Leverage [Kuadrant](https://kuadrant.io) and [Authorino](https://github.com/kuadrant/authorino) for enforcing authN/authZ and performing the token exchange within the scope of the MCP gateway

Closes #137 

### End-to-end flow

1. Client obtains an access token token with the auth server by following a supported OAuth2 flow
    * The OAuth client must have Full scope allowed configured in the Keycloak server as well as the `roles` scope assigned
    * `aud` claim will be set to a list of all MCP servers the user has access to
    * `resource_access` claim is set to an object of format `{"<server-name>":{"roles":<list of tool names>}}`
2. Client sends a `tools/list` request to `/mcp`
3. Kuadrant intercepts the request and calls Authorino ext_authz to enforce the 'broker' AuthPolicy
4. Authorino verifies the access token
5. Authorino verifies `request.headers['x-mcp-method']` is one of: "tools/list", "initialize", "notifications/initialized"
6. Authorino injects a `x-authorized-tools` request header with value set to a stringified JSON of format: `{"<server-name>":<list of tool names>}`, mapped from the `resource_access` claim
7. If `request.headers['x-mcp-virtualserver']` is present, then the broker filters the list of tools to the specified MCP server only
8. If `request.headers['x-authorized-tools']` is present, then the broker filters the list of tools to the intersection with the authorised tools only
9. The broker responds to the `tools/list` request with the list of tools (filtered or unfiltered)
10. Client sends a request to an MCP server passing the same token obtained at (1)
11. Kuadrant intercepts the request and calls Authorino ext_authz to enforce the 'router' AuthPolicy
12. Authorino verifies the access token
13. Authorino performs a token exchange of the access token to obtain a new token, with more limited scope, where the `aud` claim is set to the targeted MCP server
14. Authorino verifies the new token has `aud` claim set to the targeted MCP server
15. Authorino verifies the new token has `resource_access` claim set to an object with an entry `"<targeted MCP server>":{"roles":<list of tools including the requested tool>}`
16. Authorino replaces the `Authorization` request header with the new token (obtained from the token exchange)
17. The router send the request to the targeted MCP server
18. The MCP server verifies the access token with `aud` claim set to its own name (and nothing else)

### Try the example

❶ Setup the local environment and deploy the example

```sh
make local-env-setup
make inspect-gateway
```

❷ Using the MCP Inspector, clean up any previous auth settings and then click on 'Connect', to obtain an access token. Use `mcp` / `mcp` for credentials.

This will create a client using OAuth2 [Dynamic Client Registration](https://datatracker.ietf.org/doc/html/rfc7591) and obtain a access token that looks like this:

```json
{
  "aud": [
    "server1.mcp.local",
    "mcp-gateway",
    "server3.mcp.local",
    "oidc-server.mcp.local"
  ],
  "auth_time": 1760092590,
  "azp": "cd0dbba7-052c-42d8-a2a4-7b9e0cd49004",
  "exp": 1760094390,
  "groups": [
    "accounting",
    "mcp-users"
  ],
  "iat": 1760092590,
  "iss": "http://keycloak.127-0-0-1.sslip.io:8889/realms/mcp",
  "jti": "onrtac:3f464dc1-97f5-a674-eec9-db638e439a12",
  "resource_access": {
    "mcp-gateway": {
      "roles": [
        "impersonator"
      ]
    },
    "oidc-server.mcp.local": {
      "roles": [
        "hello_world"
      ]
    },
    "server1.mcp.local": {
      "roles": [
        "greet"
      ]
    },
    "server3.mcp.local": {
      "roles": [
        "add"
      ]
    }
  },
  "scope": "groups",
  "sid": "81ce1d2e-35e5-4507-bd93-13f944d98dda",
  "sub": "9c949380-5811-4234-8f84-e021c211ce83",
  "typ": "Bearer"
}
```

❸ List the tools the user has access to

The returned list of tools will be filtered only to the tools authorised for user `mcp` according to the user's group membership and scopes associations.

❹ Run one of the tools – e.g.: server1's `greet` tool

The targeted server will receive in the `Authorization` header a scoped token that looks like this:

```json
{
  "aud": "server1.mcp.local",
  "azp": "mcp-gateway",
  "exp": 1759615077,
  "iat": 1759613277,
  "iss": "http://keycloak.127-0-0-1.sslip.io:8889/realms/mcp",
  "jti": "ntrtte:4d238013-b3ec-eaf4-4102-7d63b9e122fe",
  "resource_access": {
    "server1.mcp.local": {
      "roles": [
        "greet"
      ]
    }
  },
  "scope": "openid",
  "sid": "81ce1d2e-35e5-4507-bd93-13f944d98dda",
  "sub": "9c949380-5811-4234-8f84-e021c211ce83",
  "typ": "Bearer"
}
```

### Also in this PR

- Fix kind load-image for Podman
- Bootstrap mcp realm from a configmap at the deployment of keycloak using realm import

### DIFF
--- a/internal/broker/broker.go
+++ b/internal/broker/broker.go
@@ -148,14 +148,14 @@ type mcpBrokerImpl struct {
 // this ensures that mcpBrokerImpl implements the MCPBroker interface
 var _ MCPBroker = &mcpBrokerImpl{}
 
-// BrokerOpts is a set of optional key values to configure the broker, if not present the broker will have a sensible default
-type BrokerOpts struct {
+// Opts is a set of optional key values to configure the broker, if not present the broker will have a sensible default
+type Opts struct {
 	EnforceToolFilter       bool
 	TrustedHeadersPublicKey string
 }
 
 // NewBroker creates a new MCPBroker
-func NewBroker(logger *slog.Logger, opts BrokerOpts) MCPBroker {
+func NewBroker(logger *slog.Logger, opts Opts) MCPBroker {
 
 	mcpBkr := &mcpBrokerImpl{
 		// knownSessionIDs: map[downstreamSessionID]clientStatus{},
--- a/internal/broker/filtered_tools_handler.go
+++ b/internal/broker/filtered_tools_handler.go
@@ -24,7 +24,7 @@ const (
 
 // FilterTools will reduce the tool set down to those passed based the authorization lay via the x-authorized-tools header
 // The header is expected to be signed as a JWT if we cannot verify the JWT then nothing will be returned
-func (broker *mcpBrokerImpl) FilteredTools(_ context.Context, id any, mcpReq *mcp.ListToolsRequest, mcpRes *mcp.ListToolsResult) {
+func (broker *mcpBrokerImpl) FilteredTools(_ context.Context, _ any, mcpReq *mcp.ListToolsRequest, mcpRes *mcp.ListToolsResult) {
 	originalTools := make([]mcp.Tool, len(mcpRes.Tools))
 	copy(originalTools, mcpRes.Tools)
 	// set to empty by default
@@ -120,7 +120,7 @@ func (broker *mcpBrokerImpl) filterTools(authorizedTools map[string][]string) []
 
 // validateJWTHeader validates the JWT header expects ES256 alg.
 func validateJWTHeader(token string, publicKey string) (*jwt.Token, error) {
-	return jwt.Parse(token, func(t *jwt.Token) (any, error) {
+	return jwt.Parse(token, func(_ *jwt.Token) (any, error) {
 		block, _ := pem.Decode([]byte(publicKey))
 		pubkey, err := x509.ParsePKIXPublicKey(block.Bytes)
 		if err != nil {
--- a/cmd/mcp-broker-router/main.go
+++ b/cmd/mcp-broker-router/main.go
@@ -211,7 +211,7 @@ func setUpBroker(address string, toolFiltering bool) (*http.Server, broker.MCPBr
 		WriteTimeout: 10 * time.Second,
 	}
 
-	mcpBroker := broker.NewBroker(logger, broker.BrokerOpts{
+	mcpBroker := broker.NewBroker(logger, broker.Opts{
 		EnforceToolFilter:       toolFiltering,
 		TrustedHeadersPublicKey: os.Getenv("TRUSTED_HEADER_PUBLIC_KEY"),
 	})
--- a/internal/broker/broker_test.go
+++ b/internal/broker/broker_test.go
@@ -30,7 +30,7 @@ const (
 )
 
 var logger = slog.New(slog.NewTextHandler(os.Stdout, nil))
-var brokerOpts = BrokerOpts{
+var brokerOpts = Opts{
 	EnforceToolFilter: false,
 }
 
--- a/internal/broker/filtered_tools_handler_test.go
+++ b/internal/broker/filtered_tools_handler_test.go
@@ -250,7 +250,7 @@ VEiyi/nozagw7BaWXmzbOWyy95gZLirTkhUb1P4Z4lgKLU2rD5NCbGPHAA==
 			if tc.AllowedToolsList != nil {
 				headerValue := createJWTHeader(tc.AllowedToolsList)
 				request.Header = http.Header{
-					authorizedToolsHeader: {string(headerValue)},
+					authorizedToolsHeader: {headerValue},
 				}
 			}
 			mcpBroker.mcpServers = tc.RegisteredMCPServers
--- a/internal/mcp-router/request_test.go
+++ b/internal/mcp-router/request_test.go
@@ -159,7 +159,7 @@ func TestHandleRequestBody(t *testing.T) {
 				},
 			},
 		},
-		Broker: broker.NewBroker(logger, broker.BrokerOpts{}),
+		Broker: broker.NewBroker(logger, broker.Opts{}),
 	}
 
 	data := &MCPRequest{
--- a/internal/mcp-router/server_test.go
+++ b/internal/mcp-router/server_test.go
@@ -24,7 +24,7 @@ func TestSetupSessionCache(_ *testing.T) {
 				},
 			},
 		},
-		Broker: broker.NewBroker(logger, broker.BrokerOpts{}),
+		Broker: broker.NewBroker(logger, broker.Opts{}),
 	}
 
 	// We can't test the internals of this, because it returns nothing on error...


config/mcp-system/tool-call-auth.yaml
```diff
@@ -9,38 +9,59 @@ spec:
...
-          code: 403
-          body:
-            value: |
-              {
-                "error": "Forbidden", 
-                "message": "MCP Tool Access denied. Insufficient permissions for this tool."
-              }
-        unauthenticated:
-          code: 401
+  rules:
```
maleck13, line 12: does it make sense to have multiple examples here? Perhaps for some users token exchange is a step they don't want to take yet so having a stepped approach to show examples of different options might be useful?

config/samples/oauth-token-exchange/tools-call-auth.yaml
```diff
@@ -9,38 +9,59 @@ spec:
...
-          code: 403
-          body:
-            value: |
-              {
-                "error": "Forbidden", 
-                "message": "MCP Tool Access denied. Insufficient permissions for this tool."
-              }
-        unauthenticated:
-          code: 401
+  rules:
```
guicassolato: Let me know if https://github.com/kagenti/mcp-gateway/pull/224/commits/f3ee5681a8ff56090f1b7b471fad7e5f0d8d4889 works

config/samples/oauth-token-exchange/tools-call-auth.yaml
```diff
@@ -0,0 +1,67 @@
...
+          url: http://keycloak.keycloak.svc.cluster.local/realms/mcp/protocol/openid-connect/token # FIXME: this will make the token exchange request fail with 'Invalid Token' due to mismatching Keycloak hostname compared to what's in the `iss` claim of the subject token
+          method: POST
+          headers:
+            Authorization:
+              value: 'Basic YXV0aG9yaW5vOnNlY3JldA==' # authorino:secret
+          bodyParameters:
+            grant_type:
+              value: urn:ietf:params:oauth:grant-type:token-exchange
+            subject_token:
+              expression: request.headers['authorization'].split('Bearer ')[1] # remove "Bearer "
```
esnible, line 29: This line upsets me because I don't know what this language will do with `[1]` if the split generates 0 or 1 items.  The authorization header comes from outside the system and therefore can be controlled by an attacker.  Perhaps this is no problem, but I couldn't figure out what language is used here (Rego expression?) and how to test various inputs in a playground to ensure safety.

internal/broker/broker.go
```diff
@@ -126,13 +137,30 @@ type mcpBrokerImpl struct {
...
 	logger *slog.Logger
+
+	//allowUnfilteredList
+	enforceToolFilter bool
 }
 
 // this ensures that mcpBrokerImpl implements the MCPBroker interface
 var _ MCPBroker = &mcpBrokerImpl{}
 
+type BrokerOpts struct {
```
esnible, line 148: This type is capitalized, please add a comment.

internal/broker/filtered_tools_handler.go
```diff
@@ -0,0 +1,69 @@
...
+			slog.Debug("full list not allowed")
+			mcpRes.Tools = filteredTools
+			return
+		}
+		return
+	}
+	slog.Debug("filtering tools based on header", "value", allowedToolsValue)
+	authorizedTools := map[string][]string{}
+	if err := json.Unmarshal([]byte(allowedToolsValue), &authorizedTools); err != nil {
+		slog.Error("failed to unmarshall authorized tools header returning empty tool set", "error", err)
```
esnible, line 37:
```suggestion
		slog.Error("failed to unmarshall authorized tools header returning empty tool set", "error", err, "allowedToolsValue", allowedToolsValue)
```

internal/broker/filtered_tools_handler.go
```diff
@@ -0,0 +1,69 @@
...
+		mcpRes.Tools = filteredTools
+		return
+	}
+
+	for server, allowedToolNames := range authorizedTools {
+		slog.Debug("checking tools for server ", "server", server, "allowed tools for server", allowedToolNames, "mcpServers", broker.mcpServers)
+		// we key off the host so have to iterate for now
+		upstreamServer := broker.mcpServers.findByHost(server)
+		if upstreamServer == nil {
+			slog.Debug("failed to find registered upstream ", "server", server)
```
esnible, line 47:
```suggestion
			slog.Debug("failed to find registered upstream", "server", server)
```

There are several other places a log message or log key ends with a space, all should be cleaned up.

internal/broker/broker.go
```diff
@@ -126,13 +137,30 @@ type mcpBrokerImpl struct {
...
+	EnforceToolFilter bool
+}
+
 // NewBroker creates a new MCPBroker
-func NewBroker(logger *slog.Logger) MCPBroker {
+func NewBroker(logger *slog.Logger, opts BrokerOpts) MCPBroker {
```
esnible, line 153: Consider making BrokerOpts be an interface and the constructor `NewBroker(logger *slog.Logger, opts ...BrokerOpt)` so that we can add options and don't need to change so much code?  (Or keep it this way, just a suggestion).

config/samples/oauth-token-exchange/tools-call-auth.yaml
guicassolato: This is [Common Expression Language (CEL)](https://cel.dev).

In general, an expression like this in CEL could be a problem if the split returned less than 2 items. However, in this particular instance, it is not a problem and it will never be a problem.

The reason is because the CEL expression is evaluated in the metadata phase of the AuthPolicy, which means it passed the authentication phase. The authentication phase won't ever succeed if the value of the `Authorization` header is not exactly equal to `Bearer <valid jwt>`.

guicassolato:
**A note about using `*.sslip.io` host names (and `*.svc.cluster.local` internally) in the demo:**
  
Keycloak requires the issuer URL in the subject access token (`iss` claim) to match the one used to request the token exchange. Otherwise, the server will reject the token exchange with an _'Invalid Token'_ error. In other words, we cannot use `keycloak.127-0-0-1.sslip.io:8889` externally and `keycloak.keycloak.svc.cluster.local` internally, because the URLs wouldn't match. By default, Keycloak determines the issuer URLs from the request headers.  
  
There are multiple ways to try to workaround this. One of them is by setting a static *frontend URL* for the Keycloak realm. A similar approach, though with effects for the entire auth server, is by specifying the `--hostname` option when starting up the server. These approaches do make Keycloak consistently use the specified host:port in all exposed configurations (such as the well-known OpenID metadata) and in the `iss` claim of every issued token—rather than inferring it from the request. As a result, all tokens and configuration documents would reference the same issuer URL.  
  
However, setting a static Keycloak host name at the server side introduces its own problems. Many OAuth/OIDC client libraries enforce strict checks: the issuer URL in the requested configuration must exactly match the one returned, and the same rule often applies when validating tokens as well. This means the workaround could easily cause clients to fail.  
  
The safer approach is to route internal workloads through the gateway to reach Keycloak, by setting internal host:port equal to the external one. In production, the auth server is usually exposed under a valid, resolvable domain name that works both externally and internally. While this adds an extra hop, it ensures consistent policy enforcement on Keycloak routes and keeps external and internal traffic aligned.  
  
To implement the safer approach, there are still a couple of different options to consider, represented in the matrix below, with my own personal opinion about the pros and cons of each option:

<table>
<thead>
<tr>
<td>Component to patch \ Wiring of the Gateway IP</td>
<td>
<i>Gateway IP statically set in the patch</i>
<br/><br/>
<b>Pros:</b> less resources, simple<br/>
<b>Cons:</b> brittle – changing the IP of the gateway beaks the configuration | Use a Kuadrant DNSPolicy for dynamic reconciliation of the DNS records
</td>
<td>
<i>Use a Kuadrant DNSPolicy for dynamic reconciliation of the DNS records</i>
<br/><br/>
<b>Pros:</b> more robust / less error-prone, closer to a production setup<br/>
<b>Cons:</b> more complex, requires a DNS server (e.g. a dedicated CoreDNS instance or Route53 zone, etc), possibly overkill for a demo
</td>
<tr>
</thead>
<tbody>
<tr>
<td><i>Authorino pods</i> only – by patching the Authorino Deployment with additional DNS resolution configs (<code>hostAliases</code> or <code>dnsPolicy</code>/<code>dnsConfig</code>), to point Keycloak's hostname to the Gateway's IP</td>
<td colspan="2">
<b>Pros:</b> less intrusive / reduced scope<br/>
<b>Cons:</b> brittle – touching the Authorino CR could lose the patch due to overwriting by the Authorino Operator
</td>
</tr>
<tr>
<td><i>Entire cluster</i> – by patching kube-system CoreDNS instance to hijack Keycloak's or even <code>*.sslip.io</code> host names for the Gateway IP</td>
<td colspan="2">
<b>Pros:</b> consistent across all workloads, resilient to modifying the spec of the pods<br/>
<b>Cons:</b> pervasive
</td>
</tr>
</tbody>
</table>

Right now, I have implemented for this demo the option of patching for Authorino pods only, with static wiring of the Gateway IP in the DNS settings.

PR #233, opened 2025-10-01
Refactor to improve ability to test and read logic in the request handler

maleck13:
This is an attempt to improve the code here beyond the POC level to something a little more structured. 

Current State:

there is still a bunch of commented out code while I add more tests and test with mcp-inspector.


jasonmadigan (approved):
thanks for adding obs pattern for config changes, was on my laundry list

PR #244, opened 2025-10-03
Fixup helm workflow

david-martin:
Couple of helm fixups while testing out the workflow here https://github.com/kagenti/mcp-gateway/actions/runs/18226597349


jasonmadigan:
Style check failure

PR #258, opened 2025-10-07
oAuth + API key

jasonmadigan:
re: https://github.com/kagenti/mcp-gateway/issues/201

credentialsRef secrets - append as x-mcp-api-key header. If used with oAuth + AP, copy credentialRef into Authorization _post_ JWT validation


jasonmadigan:
We should nudge this into the realm of the AuthPolicy instead

PR #265, opened 2025-10-07
Fix keycloak port fowarding & env vars

david-martin:
Fixes a regression from https://github.com/kagenti/mcp-gateway/pull/225 where the keycloak port forwarding args were removed from a couple of targets.
I hit this when trying to run `make inspect-gateway` followed by `make oauth-example-setup`, which expected to be able to reach keycloak then.

Also renames the 2 newly introduced variables from that PR to something more appropriate as both ports are http listeners, not http & https.


jasonmadigan (approved):
thanks @david-martin

PR #283, opened 2025-10-17
Integration with Vault

### LINKED ISSUE #140: Support credential brokering and lookup from external store (e.g. Vault)
The MCP Gateway should support brokering credentials for upstream MCP servers by looking them up from a secure credential store (e.g. Vault, Kubernetes Secrets, external secrets manager).

Currently, the gateway can validate incoming client tokens (via the configured auth provider such as Keycloak). After validation, it must be able to retrieve the correct credential to use when calling the target MCP server/tool. These credentials may vary depending on the scenario:

* Per MCP server: a static credential is configured for the broker to authenticate to the server (similar to API key or service account).
* Per tenant / user: credentials are scoped to the authenticated client. This ensures that the upstream MCP server/tool (e.g. Slack, GitHub, ServiceNow) only grants access aligned with that specific user’s permissions.

Considerations:

* Support different credential types (OAuth tokens, API keys, client secrets, etc).
* Define lookup strategy (per user vs. per server).
* Integrate with external secrets management systems (Vault initially?).
* Ensure isolation between tenants and prevent leakage of cross-user credentials.

guicassolato:
Integration with Vault for MCP servers whose authentication is not based on OAuth2.

Closes #140

### How it works

- Agent obtains an OAuth2 token with Keycloak and uses this token for tools/list and tools/call
- On tools/call, Kuadrant validates the Keycloak-issued bearer token; then
  - if an entry `{username}/{server-host}` exists in Vault, the `Authorization:` header of the request is replaced with the secret stored in Vault;
  - otherwise, there are 2 cases:
    - if the bearer token is scoped to multiple audiences (default), Kuadrant performs an OAuth token exchange with Keycloak to scope the token down to the target MCP server only (single audience), and then replaces the `Authorization:` header with the scoped token;
    - if the bearer token is already scoped to the target MCP server only, Kuadrant keeps the original `Authorization:` header of the request.
  - In all cases, Kuadrant checks in the access token the user permissions to call the tool – either in the original bearer token (if no token exchange was performed) or in the scoped token (if token exchange was performed), this means all access control permissions must be stored in Keycloak as roles of the OAuth2 server clients, whether the server is an MCP server that supports OAuth or an MCP server with authentication based on secret stored in Vault.

### Try the example

```sh
make local-env-setup
make inspect-gateway
```

Then:

1. Create a Vault secret for MCP test-server2

```sh
kubectl port-forward -n vault svc/vault 8200:8200 2>&1 >/dev/null &

curl -s -H "X-Vault-Token: root" -H 'Content-Type: application/json' -X POST \
  --data '{"data":{"token":"s3cr3t"}}' \
  http://localhost:8200/v1/secret/data/mcp/server2.mcp.local
```

2. In the Keycloak UI: add `server2.mcp.local`'s `headers` role to the `accounting` group of the `mcp` realm.

3. In the MCP Inspector: connect with credentials `mcp` / `mcp`, list the tools, and run the test2_headers tool.

The response from the `headers` tool should echo back the received header `Authorization: Bearer s3cr3t`. For any other user or server, the value of the `Authorization:` header will be the scoped token obtained by OAuth token exhange handled automatically by the MCP gateway.

### DIFF
--- a/build/auth.mk
+++ b/build/auth.mk
@@ -52,7 +52,7 @@ oauth-token-exchange-example-setup: ## Setup auth example of enabling OAuth2 aut
 		OAUTH_RESOURCE="http://mcp.127-0-0-1.sslip.io:8888/mcp" \
 		OAUTH_AUTHORIZATION_SERVERS="http://keycloak.127-0-0-1.sslip.io:8889/realms/mcp" \
 		OAUTH_BEARER_METHODS_SUPPORTED="header" \
-		OAUTH_SCOPES_SUPPORTED="basic,groups,roles" \
+		OAUTH_SCOPES_SUPPORTED="basic,groups,roles,profile" \
 		-n mcp-system
 	@echo "✅ OAuth environment variables configured"
 	@echo ""
--- a/config/samples/oauth-token-exchange/tools-call-auth.yaml
+++ b/config/samples/oauth-token-exchange/tools-call-auth.yaml
@@ -18,7 +18,7 @@ spec:
       vault:
         http:
           urlExpression: |
-            "http://vault.vault.svc.cluster.local:8200/v1/secret/data/" + auth.identity.username + "/" + request.host
+            "http://vault.vault.svc.cluster.local:8200/v1/secret/data/" + auth.identity.preferred_username + "/" + request.host
           method: GET
           credentials:
             customHeader:


maleck13:
Test headers tool doesn't seem to be available , had to add it manually

guicassolato:
Added to the instructions

Issue #208, opened 2025-09-22
Investigate support for Resources and Prompts


esnible:
The MVP is focused on tool calls.

The next version should also gateway Resources and Prompts.

esnible:
Removing my assignment because this is a 0.6 item and should not be held.  When 0.5 is finished I can take it again if needed.

esnible:
@david-martin The "everything" sever has Resources and Prompts, and is available now, and doesn't require Kubernetes.

david-martin:
Split into two follow-up issues:
- #787 — Prompts federation
- #788 — Resources federation

PR #288, opened 2025-10-22
Broker prompts, resources, and resource templates

esnible:
For #208

This adds resources, resource templates, and prompts to the broker.
(It does not add them to the router.  The gateway'ed resources and prompts appear in the MCP inspector, but cannot be invoked.)

To test this PR locally, first start the "everything MCP server" with

```bash
npm install -g @modelcontextprotocol/server-everything@latest
npm run start:streamableHttp
```

Next, run the MCP gateway against the everything server.

```bash
cat <<EOF > /tmp/mcp-gateway-config.yaml
servers:
  - name: test
    url: http://localhost:3001/mcp
    toolPrefix: everything_
    enabled: true
    hostname: t2.example.com
EOF

go run cmd/mcp-broker-router/main.go -mcp-broker-public-address localhost:8081 -mcp-gateway-config /tmp/mcp-gateway-config.yaml
```

At this point, run `DANGEROUSLY_OMIT_AUTH=true npx @modelcontextprotocol/inspector` and compare the difference between connecting to http://localhost:3001/mcp and http://localhost:8081/mcp in terms of prompts, resources, and resource templates.

### DIFF
--- a/internal/broker/broker.go
+++ b/internal/broker/broker.go
@@ -291,7 +291,7 @@ func (m *mcpBrokerImpl) RegisterServerWithConfig(
 	// Add to map BEFORE discovering tools so createMCPClient can find it
 	m.mcpServers[upstreamMCPURL(mcpServer.URL)] = upstream
 
-	discovered, err := m.discoverTools(ctx, upstream)
+	discovered, err := m.discoverUpstream(ctx, upstream)
 	if err != nil {
 		slog.Info("Failed to discover tools, will retry with backoff", "mcpURL", mcpServer.URL, "error", err)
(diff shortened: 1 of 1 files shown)


esnible:
> Wondering should we add this to our test MCP servers https://hub.docker.com/r/mcp/everything permanently

We should put prompt/template/resource handling into our timeline.  I understand the Broker and had no problem adding it, but I don't understand if we are ready to add it to the router.

We can consider adding the `everything` image to our dev environment.

We SHOULD create an issue to add prompts/templates/resources to our Go test server, so that we can do unit tests on the broker.

PR #292, opened 2025-10-23
Service labels

### LINKED ISSUE #289: Service name hardcoded, causing issues when installed via helm
As found in #284, the service name is hardcoded in https://github.com/kagenti/mcp-gateway/blob/736eedf93fe871b3e3dd0e62ce09e7b63cfb38ab/pkg/controller/server_validator.go#L48

Let's use labels to find the service as it's more robust.
There is a `component` label already, https://github.com/kagenti/mcp-gateway/blob/main/charts/mcp-gateway/templates/service.yaml
but let's change this to a common label https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/

david-martin:
Fixes #289 by changing the labels for looking up endpointslices for the controller to talk to the broker.
The important thing is the service name isn't used when looking up endpointslices as the service name can be different depending on how you installed (kustomize vs helm with a different name used).

The e2e tests fail if you change the labelling, but no the code.
So re-running the e2e tests with `make test-e2e` is sufficient to validate,

however this can also be validated with the `sample_local_helm_setup.sh` script e.g.

```shell
export USE_LOCAL_CHART=true
./charts/sample_local_helm_setup.sh
```
You will need to build and deploy the controller code though to include the code fixes:

```
make build-and-load-image
```

### DIFF
--- a/charts/mcp-gateway/templates/service.yaml
+++ b/charts/mcp-gateway/templates/service.yaml
@@ -6,7 +6,7 @@ metadata:
   namespace: {{ include "mcp-gateway.namespace" . }}
   labels:
     {{- include "mcp-gateway.labels" . | nindent 4 }}
-    component: broker
+    app.kubernetes.io/component: mcp-broker
 spec:
   type: ClusterIP
   selector:
@@ -29,7 +29,7 @@ metadata:
   namespace: {{ include "mcp-gateway.namespace" . }}
   labels:
     {{- include "mcp-gateway.labels" . | nindent 4 }}
-    component: config-api
+    app.kubernetes.io/component: mcp-broker-config
 spec:
   type: ClusterIP
   selector:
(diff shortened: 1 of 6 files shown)


build/e2e.mk
```diff
@@ -15,7 +15,14 @@ test-e2e-deps: ginkgo ## Install e2e test dependencies
...
 		echo "Setting up cluster for e2e tests..."; \
-		$(MAKE) local-env-setup; \
+		$(MAKE) tools; \
+		$(MAKE) kind-create-cluster; \
+		$(MAKE) build-and-load-image; \
+		$(MAKE) gateway-api-install; \
+		$(MAKE) istio-install; \
+		$(MAKE) metallb-install; \
+		$(MAKE) deploy-namespaces; \
+		$(MAKE) deploy-gateway; \
```
david-martin: this change is to slim down the test environment to just what's needed.
More could maybe be removed , but for now I removed keycloak and kuadrant.

tests/e2e/happy_path_test.go
```diff
@@ -85,18 +87,29 @@ var _ = Describe("MCP Gateway Happy Path", func() {
 
 		By("Setting up port-forward to broker for status check")
 		statusPortForwardCmd := setupPortForward("mcp-broker-router", SystemNamespace, "18081:8080")
-		defer statusPortForwardCmd.Process.Kill()
+		defer func() {
```
david-martin: lots of additional debugging and error catching in this file while I was struggling with some failures related to my environment.

jasonmadigan (approved):
this looks good

PR #298, opened 2025-10-29
Use jwt instead of uuid for mcp-session-id

maleck13:
This removes the UUID default session and replaces it with a JWT instead. This improves security, allows us to extend its uses and invalidate the session

Validation

You should be able to use MCP inspector but see in the logs that we are now being sent a JWT token instead of an opaque string

When the session becomes invalid the client will be sent a 404 as per the MCP spec

### DIFF
--- a/cmd/mcp-broker-router/main.go
+++ b/cmd/mcp-broker-router/main.go
@@ -36,10 +36,11 @@ import (
 )
 
 var (
-	mcpConfig = &config.MCPServersConfig{}
-	mutex     sync.RWMutex
-	logger    = slog.New(slog.NewTextHandler(os.Stdout, nil))
-	scheme    = runtime.NewScheme()
+	mcpConfig            = &config.MCPServersConfig{}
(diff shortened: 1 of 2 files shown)


cmd/mcp-broker-router/main.go
```diff
@@ -127,9 +132,17 @@ func main() {
 		return
 	}
 	ctx := context.Background()
-	brokerServer, mcpBroker, mcpServer := setUpBroker(mcpBrokerAddrFlag, enforceToolFilteringFlag)
+	var jwtSessionMgr *session.JWTManager
+	if jwtSigningKeyFlag != "" {
```
david-martin: The signing key is optional?
What's the expected behaviour if no key is specified, and expected value for the session id header?

maleck13: It has a default value that shouldn't really be used other than locally. But I think we can fail if not set at all and log if set to the default 👍

Issue #263, opened 2025-10-07
Idea: Tool annotation headers for observability


esnible:
MCP-Gateway currently adds headers for observability:

- `x-mcp-toolname:` used to indicate what tool (including prefix) is being accessed
- `x-mcp-method:` used to indicate the json RPC method being called

This proposal is to add `x-mcp-annotation-hints` which are derived from the annotations in the MCP spec at https://modelcontextprotocol.io/legacy/concepts/tools#tool-annotations .  These are for observability --when using an observability tool that separates traffic by header to by able to measure and visualize the amount of readOnly, destructive, idempotent, and openWorld tool calls.

The hints are only hints, because they are not to be trusted if the server is not trusted.  The spec makes the hints optional, so they have three values: `true`, `false,` and unspecified.

- We could provide a `x-mcp-annotation-hints` header which includes hint and value, e.g.  `x-mcp-annotation-hints: destructive=true, idempotent=true, readOnly=false`.
- We could provide two headers, `x-mcp-annotation-true-hints` and `x-mcp-annotation-false-hints`, with values such as `x-mcp-annotation-true-hints: destructive, idempotent`  and `x-mcp-annotation-false-hints: openWorld`

The Gateway currently knows the hints for all tools through an in-memory mark3labs data structure.  The implementation of this feature is probably having the router keep a mapping from tool name to hint in memory.

d0w:
Hi I'm new to this project and if possible I can work on this! Are there any acceptance criteria or methods to test this to ensure it's working as expected?

esnible:
@david-martin , can you comment on if this idea is a good one or not, before @d0w begins to implement it?  I created this issue because it seems very useful for future debugging.
@david-martin can you comment on if we have any observability tests in the current environment?  I see that we import opentelemetry and prometheus indirectly, but how exactly do we do logging and are there tests for it?

@d0w usually when I install MCP-Gateway it is in a Kagenti kind environment.  In that environment there is a UI at http://kagenti-ui.localtest.me:8080/Observability .  I believe the headers should be visible there, but I am having difficulties with my env right now and can't check.

david-martin:
@esnible 
Can we close this as done with #311 ?

PR #311, opened 2025-11-03
Add tool annotations to MCP request headers

d0w:
Issue: https://github.com/kagenti/mcp-gateway/issues/263

# Description
This PR addresses #263 where mcp requests now have a `x-mcp-annotation-hints` header that abides by the MCP tool annotation spec [here](https://modelcontextprotocol.io/legacy/concepts/tools#tool-annotations).

# Changes
The broker now, on the discovery of tools, populates the `upstreamToolInfo` object with `tool.Annotations`, a field of `mcp.Tool`. This is stored in memory alongside tool name and url (`toolMapping` map). The router will then, on an mcp request, check the broker for tool annotations and append `true`, `false`, or `unspecified` to the headers (e.g. `readOnly=true,destructive=false,idempotent=true,openWorld=unspecified`).

# To test
I used a Kagenti installation and modified the mcp-gateway installation to use a locally built image. 

1. Run `make docker-build`
2. Go to `kagenti/kagenti` and modify https://github.com/kagenti/kagenti/blob/main/kagenti/installer/app/resources/gateway-deployment.yaml to use the same image that you just build, ensuring to set imagePullPolicy to IfNotPresent or Never.
```
containers:
  - name: mcp-broker-router
    image: kagenti/mcp-gateway:latest
    imagePullPolicy: Never
```
3. Run the kagenti installer using their [instructions](https://github.com/kagenti/kagenti/blob/main/docs/demos.md).
4. Import/create any combination of agent/mcp, ensuring the agent uses the MCPGateway route by applying the proper CRs. I followed the instructions [here](https://github.com/kagenti/kagenti/blob/main/docs/gateway.md) for the weather agent.
5. `kubectl logs -f svc/mcp-broker-router -n mcp-system` to follow the logs for the router and make a request with the agent.

You should see the headers be set like so

Screenshot 2025-11-03 at 4 58 12 PM

# Observability
To showcase an example where these headers can be used, you can follow the steps outlined in this added [doc](https://github.com/d0w/mcp-gateway/blob/tool-annotations/docs/guides/observability.md) that runs you through adding Loki + Grafana to the cluster and setting up visualizations for mcp tool annotation hints.

### DIFF
--- a/internal/mcp-router/headers_test.go
+++ b/internal/mcp-router/headers_test.go
@@ -9,7 +9,7 @@ func Test_Headers(t *testing.T) {
 
 	expected := map[string]string{
 		toolHeader:            "test_tool",
-		toolAnnotationsHeader: "detructive=true,indempotent=true,readOnly=false",
+		toolAnnotationsHeader: "destructive=true,idempotent=true,readOnly=false",
 		authorityHeader:       "mcp1.mcp.local",
 		"authorization":       "auth",
 		methodHeader:          "tools/call",


d0w:
Thanks for the review @david-martin

> Before merging the feature, I would still like to see how this helps with observability in a more real scenario. I wouldn't like downstream mcp servers to depend on these headers for example if they are only meant for observability, and we change them in some way.

For now I'll get a use case up and running while there's still deliberation on broker/router dependency. I can go off your [comment](https://github.com/kagenti/mcp-gateway/issues/263#issuecomment-3461608737) that suggested using Loki. I'll most likely run the gateway within the Kagenti stack and use an existing MCP server like the GitHub MCP to see if adding the header affects MCP calls.
- Please let me know if that's the type of scenario you're looking for and if you have a preferred stack instead of Loki like using Prometheus that's installed with the main UI 
- Did we want this scenario to be part of the code change somewhere or just shown in the description of the PR?


> Also, what if a client sets the header themselves. What's the knock on effect? Could it impact cardinality in some way on a backend observability system?

One potential work around could be to just delete the request header from the client and reset it in the router. Not sure if that's a viable solution.

d0w:
@david-martin I've added some instructions on setting up an observability stack with the tool annotations as one example of visualizations for log data coming from the mcp-broker-router. 

Here are some screenshots of some example metrics that show how often a tool with a certain hint is called.

Screenshot_20251109_160751

Screenshot_20251109_160737

----------------------------------------------------------------------------------------------------------------------------------

As for the headers from the client: To my best understanding, the router doesn't send any arbitrary headers back to envoy on a request body processing request. It does store the client headers in `s.requestHeaders` but the request handlers manually take values from this mapping. I've also verified this with a request where I added the annotation hints header manually.

Request debug log (random test value)
Screenshot_20251109_160459


You'll see that the request body sent back to Envoy ignores the request header and simply uses the tool annotations.
Screenshot_20251109_160431

Please let me know if you would like me to investigate further on showing the observability setup or anything else.

internal/mcp-router/request_handlers.go
```diff
@@ -116,6 +115,31 @@ func (s *ExtProcServer) HandleMCPRequest(ctx context.Context, mcpReq *MCPRequest
 		return calculatedResponse.Build()
 	}
 
+	// Get tool annotations from broker and set headers
+	if s.Broker != nil {
+		if annotations, hasAnnotations := s.Broker.ToolAnnotations(toolName); hasAnnotations {
+			// build header value (e.g. readOnly=true,destructive=false,openWorld=true)
```
david-martin: If a client stes a `x-mcp-annotation-hints` header, and a tool doesn't have any annotations,
will that impact logging/observability?
From my read of the code, I don't think this matters right now, but it would allow users to set headers that __may__ impact how observability is done in future. Perhaps we should strip the header in the router if the client has set it?

So this log statement is really important for observability:
```
s.Logger.Info(fmt.Sprintf("Sending MCP body routing instructions to Envoy: %+v", response))
```
Without this, there is no way to observe what annotations are attached to the tool being called.
It feels fragile as it could easily be changed or removed without realising the impact of that on observability.

What do you think of a bit more structure around the logging, so it's a deliberate decision to log mcp specific headers (`x-mcp-toolname`, `x-mcp-method`, `:authority`, `x-mcp-annotation-hints` etc...).
This could be a dedicated function that logs these in a structured way (which makes it easier to parse from loki queries), or using https://pkg.go.dev/go.opentelemetry.io/otel (which could add future tracing capabilities as well).

docs/guides/observability.md
```diff
@@ -0,0 +1,89 @@
...
+   kubectl get secret --namespace monitoring my-grafana -o jsonpath="{.data.admin-password}" | base64 --decode ; echo
+
+
+2. The Grafana server can be accessed via port 80 on the following DNS name from within your cluster:
+
+   my-grafana.monitoring.svc.cluster.local
+
+   Get the Grafana URL to visit by running these commands in the same shell:
+     export POD_NAME=$(kubectl get pods --namespace monitoring -l "app.kubernetes.io/name=grafana,app.kubernetes.io/instance=my-grafana" -o jsonpath="{.items[0].metadata.name}")
+     kubectl --namespace monitoring port-forward $POD_NAME 3000
```
esnible:
```suggestion
   We will port-forward to expose Grafana at http://localhost:3000/login :
   
   ```bash
     kubectl --namespace monitoring port-forward deployment/my-grafana 3000
     ```
```

david-martin (approved):
@d0w I'm going to merge this as it adds value, given the additional logging and the examples with loki.
Thanks for taking the time to do this.

This will be a great starting point for future observability pieces.

Issue #20, opened 2025-08-18
Spike & investigate adding notifications to the router & broker


david-martin:
When an MCP Client connects to the GET /mcp endpoint to establish an SSE connection with the mcp gateway,
the gateway needs to map that session to already active sessions with various backend mcp servers (e.g. where long running tool calls are happening).
Initially, the SSE endpoint could only handle the tools/list_changed notification to figure out the right pattern of multplexing connections.

Considerations:
- the number of open connections between the mcp gateway and backend mcp servers
- how that scales with the number of active users/mcp clients
- possible efficiencies, like only connecting to a backend mcp servers SSE endpoint if there's an active tool call right now (i.e. don't connect to all possible backend mcp servers 'just in case')
- what other notifications are possible to be send from those backend mcp servers, and which ones to support (and in which order)
- capabilities support for notifications

david-martin:
Spike closed by #315 , with deisgn doc

Follow up issue to implement:
* state change event notifications support 
* elicitation support

PR #315, opened 2025-11-04
Add Notifications design doc

david-martin:
Proposed architecture for MCP Notifications support of broadcast type and client-specific notifications.
Includes a sequence diagram for either flow.

There are some open questions, that may require progress in #20 before being able to answer.

### DIFF
--- a/docs/design/notifications.md
+++ b/docs/design/notifications.md
@@ -17,6 +17,18 @@ In the MCP Gateway architecture, clients connect to the gateway's `/mcp` endpoin
 
 The gateway uses a **lazy initialization** approach where backend sessions to MCP servers are only established when a client makes a tool call. However, some notifications (like `list_changed` notifications) are logically 'broadcast' notifications and should be sent to all connected clients, not just those who have made tool calls.
 
+### Non-Goals
+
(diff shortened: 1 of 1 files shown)


docs/design/notifications.md
```diff
@@ -0,0 +1,196 @@
...
+  Gateway ->> Client1: forward notification
+  Server2 ->> Gateway: tool/call response
+  Gateway ->> Client1: tool/call response
+  Gateway -->> Server2: SSE connection closed
+```
+
+1. **Lazy Session Establishment**: When a client makes a `tools/call` request, the broker performs lazy initialization to establish a backend session to the target MCP server using the client's authentication credentials (passed in the tool call request). This initialization includes:
+   - Sending an `initialize` request to the backend server *only* if a session hasn't already been established from a previous tool call.
+   - Sending a `notifications/initialized` notification
+   - Establishing a GET `/mcp` connection for receiving SSE notifications
```
maleck13: would this get be done with the clients session id? I assume yes?

david-martin: Yes. It has to to client specific

david-martin:
Implementation in a follow up PR.

david-martin:
@maleck13 this is good for review again.

No actual code changes here.

Changes to support the following in separate PRs:

* state change events, like tool list changed ('broadcast' to all connected clients)
* elicitation responses (mapping of elicitation request ids in elicitation/create messages from server, and back in the POST response)

maleck13 (approved):
Good stuff @david-martin I think we can  merge and break down into issues

PR #318, opened 2025-11-05
improve the routing to ensure it always controls the :authority header

maleck13:
This corrects an issue where the router would not set the :authority header for none tools/calls. It also updates the design doc with clearer explanations on how to expose an MCP server. 

**Note**

This change makes the flag `--mcp-gateway-public-host` a required flag to set when deploying the router 

**Validation**

Before this change a curl request such as 

```
curl -v 'http://mcp.127-0-0-1.sslip.io:8888/mcp' -H 'Host: mcp-server-1'   -H 'Accept-Language: en-GB,en-US;q=0.9,en;q=0.8'   -H 'Connection: keep-alive'   -H 'Origin: http://localhost:6274'   -H 'Referer: http://localhost:6274/'   -H 'Sec-Fetch-Dest: empty'   -H 'Sec-Fetch-Mode: cors'   -H 'Sec-Fetch-Site: same-site'   -H 'User-Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/141.0.0.0 Safari/537.36'   -H 'accept: application/json, text/event-stream'   -H 'content-type: application/json'   -H 'sec-ch-ua: "Google Chrome";v="141", "Not?A_Brand";v="8", "Chromium";v="141"'   -H 'sec-ch-ua-mobile: ?0'   -H 'sec-ch-ua-platform: "macOS"'   --data-raw '{"method":"initialize","params":{"protocolVersion":"2025-06-18","capabilities":{"sampling":{},"elicitation":{},"roots":{"listChanged":true}},"clientInfo":{"name":"inspector-client","version":"0.17.2"}},"jsonrpc":"2.0","id":0}'
```

Would bypass the broker and go straight to the MCPServer backend. Now this request will result in the request going to the MCPRouter and then the MCPBroker that should handle all initialize requests

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -125,19 +125,17 @@ define load-image
 endef
 
 .PHONY: build-and-load-image
-build-and-load-image: kind ## Build & load router/broker/controller image into the Kind cluster
+build-and-load-image: kind build-image load-image  ## Build & load router/broker/controller image into the Kind cluster and restart
 	@echo "Building and loading image into Kind cluster..."
-	$(CONTAINER_ENGINE) build $(CONTAINER_ENGINE_EXTRA_FLAGS) -t ghcr.io/kagenti/mcp-gateway:latest .
-	$(call load-image,ghcr.io/kagenti/mcp-gateway:latest)
 	kubectl rollout restart deployment/mcp-broker-router -n mcp-system 2>/dev/null || true
 	kubectl rollout restart deployment/mcp-controller -n mcp-system 2>/dev/null || true
 
 .PHONY: load-image
-load-image: kind ## Build & load router/broker/controller image into the Kind cluster	
+load-image: kind ## Load the mcp-gateway image into the kind cluster
 	$(call load-image,ghcr.io/kagenti/mcp-gateway:latest)	
 
 .PHONY: build-image
-build-image: kind ## Build & load router/broker/controller image into the Kind cluster	
+build-image: kind ## Build the mcp-gateway image
 	$(CONTAINER_ENGINE) build $(CONTAINER_ENGINE_EXTRA_FLAGS) -t ghcr.io/kagenti/mcp-gateway:latest .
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -130,6 +130,15 @@ build-and-load-image: kind ## Build & load router/broker/controller image into t
...
 	kubectl rollout restart deployment/mcp-broker-router -n mcp-system 2>/dev/null || true
+	kubectl rollout restart deployment/mcp-controller -n mcp-system 2>/dev/null || true
+
+.PHONY: load-image
+load-image: kind ## Build & load router/broker/controller image into the Kind cluster	
+	$(call load-image,ghcr.io/kagenti/mcp-gateway:latest)	
+
+.PHONY: build-image
+build-image: kind ## Build & load router/broker/controller image into the Kind cluster	
+	$(CONTAINER_ENGINE) build $(CONTAINER_ENGINE_EXTRA_FLAGS) -t ghcr.io/kagenti/mcp-gateway:latest .
```
david-martin: The comments on these 2 new targets looks like a copy/paste.
Also, the 'rollout restart' is left out of the load-image section. Maybe that was intentional?
If there's value in these standalone targets, we can call them from `build-and-load-image` to avoid duplicate cmds

internal/mcp-router/request_handlers.go
```diff
@@ -0,0 +1,174 @@
+package mcprouter
```
david-martin: The removal of request.go is actually moving existing logic into request_handlers.go, or is there more to it than that?

maleck13: It moves the existing logic and pulls out the ext_proc specific responses into the `ResponseBuilder` type. Hopefully making it all a bit more readable extendible and maintainable

PR #327, opened 2025-11-10
feat: Added MCP Everything server to test servers

### LINKED ISSUE #299: Add the mcp everything server to test servers
https://github.com/modelcontextprotocol/servers/tree/main/src/everything

If we added this test server to https://github.com/kagenti/mcp-gateway/tree/main/tests/servers, it adds a lot of mcp functionality that we have been developing ourselves in some of the test servers.
Let's leverage this everything server instead.

Let's carefully decide if this should be in addition to existing test servers, or instead of 1 or more.
I know we deliberately developed some servers with different libraries to test behaviour behind the mcp-gateway, such as session management.
We already have a lot of test servers.
Are we getting sufficient value from all of them?
Are they included in any automated tests?
Should they be moved out to a separate repo and referenced by image instead?

pzhao0:
Fixes #299 

## Description
Adds the official [MCP Everything Server](https://github.com/modelcontextprotocol/servers/tree/main/src/everything) as a test server. The everything server is a comprehensive test implementation written in TypeScript from the Model Context Protocol organization that exercises all features of the MCP protocol, including tools, resources, prompts, sampling, logging, and roots protocol support.

## Changes
- Dockerfile that builds the MCP everything server image by using the official repository as a dependency
- New deployment, service, HTTPRoute manifests for the everything server
- Modified Makefile test server targets to build and deploy new server with the rest of the test servers

## Running with MCP Gateway
Run `make local-env-setup`, which installs the complete local environment with the test MCP servers, including the everything server
Run `make inspect-gateway` to pull up the MCP inspector. Your list tools output should contain tools with the `everything_` prefix:
<br>
Screenshot 2025-11-10 at 2 53 35 PM

## Future Considerations and Notes
- The everything server has implementations for different MCP primitives such as Resources and Prompts, which are not currently supported
- The everything server has its own official published image: https://hub.docker.com/r/mcp/everything, but has not been updated in 6 months and only supports SSE
- Do we want to consolidate all these test servers in some way? What is the determining factor for overlap?
- How can we integrate [conformance frameworks](https://github.com/modelcontextprotocol/conformance) to automate the testing of these test servers? <- taking a look at this now

### DIFF
--- a/.github/workflows/test-images.yaml
+++ b/.github/workflows/test-images.yaml
@@ -59,6 +59,10 @@ jobs:
             context: tests/servers/oidc-server
             dockerfile: Dockerfile
             image-name: test-oidc-server
+          - server: everything-server
+            context: tests/servers/everything-server
+            dockerfile: Dockerfile
+            image-name: test-everything-server
 
     steps:
       - name: Check out code


maleck13:
@pzhao0 thanks for the PR 🥳  I think the conformance suite is a great addition thanks for pointing it out. I  have added an issue to track that work. 

@esnible What are your thoughts on the overlap with the everything server vs our individual MCP backends? Could we drop some of ours for this one? I know we will want multiple servers for some use cases and we may want to ensure that specific "bad behaviour" is also covered by some of these servers

PR #336, opened 2025-11-11
wildcard hostname for the `mcps` listener

### LINKED ISSUE #334: OAuth flow broken on main
The oauth flow with mcp-inspector is failing with CORs errors, and also seeing a 401 unauthenticated response from keycloak endpoint.
It looks like an AuthPolicy is affecting the keycloak listenere despite having a sectionName of mcp/mcps.

Example curl that should return a response from with oauth-authorization-server info, but is instead hitting authorino and getting denied.
The AuthPolicy shouldn't affect the keycloak listener and route.

```shell
curl -i   -X OPTIONS   'http://keycloak.127-0-0-1.sslip.io:8889/.well-known/oauth-authorization-server/realms/mcp'    -H 'Access-Control-Request-Method: GET'   -H 'Access-Control-Request-Headers: Authorization, Content-Type'
HTTP/1.1 401 Unauthorized
www-authenticate: Bearer resource_metadata=http://mcp.127-0-0-1.sslip.io:8888/.well-known/oauth-protected-resource/mcp
x-ext-auth-reason: credential not found
content-length: 84
content-type: text/plain
date: Tue, 11 Nov 2025 14:02:16 GMT
server: istio-envoy

{
  "error": "Forbidden",
  "message": "MCP Tool Access denied. Unauthenticated."
}
```

To reproduce I ran these locally

```shell
make local-env-setup
make inspect-gateway
```

and

```shell
``` 

Then tried connect in the mcp inspector UI. 
Using the browser developer tools shows the 401 responses, and also missing CORs headers (which should be applied via the keycloak HTTPRoute)

guicassolato:
fix #334

### DIFF
--- a/config/istio/gateway/gateway.yaml
+++ b/config/istio/gateway/gateway.yaml
@@ -61,6 +61,7 @@ spec:
         namespaces:
           from: All
     - name: mcps
+      hostname: '*.mcp.local'
       port: 8080
       protocol: HTTP
       allowedRoutes:


maleck13:
Actually I will move it back to use the `mcp.local` host in the example as it ties into several places. The change here was really to 1 ensure that we fully control the authority and 2 to show that you didn't have even use a domain / hostname for the routing to work.

PR #342, opened 2025-11-13
Remove port forwarding in favour of node port.

### LINKED ISSUE #345: bug everything MCP server tools return 404
We added the MCP everything server but it is returning 404 for tools calls. Possibly some path issue

```
"Error POSTing to endpoint (HTTP 404): Transport not found for sessionId 0b1a1fb1-d922-489a-b709-7bba240c6251"
```

david-martin:
This is a large set of changes, but a small surface area affected.

* removes any port-forwarding required for accessing the mcp gateway or keycloak (was previously 8888 & 8889)
* replaces that access with extraPortMappings in kind, that leverages a nodeport service
    * MCP Broker/Gateway 8001 (host) ->30080 (nodeport) ->8080 (gateway mcp listener)
    * Keycloak: 8002 (host) ->30089 (nodeport) -> 8002 (gateway keycloak listener)
* removes any make targets, echo statements and readme content that referenced port forwarding for this
* updates existing references, including docs & yaml, to use the nodeport forwarded ports instead. These are 8001 for the mcp gateway, and 8002 for keycloak.
    * http://mcp.127-0-0-1.sslip.io:8001/mcp
    * http://keycloak.127-0-0-1.sslip.io:8002/admin/master/console/

(Also fixes #345 )

To verify:

```
make local-env-setup
make inspect-gateway
```
Tools listing and calling should work as before, without any port forwarding needed.

Additional verification with auth (Not working yet)

```
```

* Connect via the inspector
* Login to keycloak as mcp/mcp
* List & use tools

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -48,7 +48,6 @@ Client → Gateway (Envoy) → Router (ext_proc) → Broker → Upstream MCP Ser
 ### Quick Start
 ```bash
 make local-env-setup     # Create Kind cluster with everything
-make dev-gateway-forward # Access gateway at localhost:8888
 make reload              # Build, load to Kind, and restart controller and broker
 ```
 
@@ -85,7 +84,8 @@ This prevents resource waste during rapid development/force-pushing.
 - 8080: Broker HTTP (/mcp endpoint)
 - 50051: Router gRPC (ext_proc)
 - 8081: Controller health probes
-- 8888: Gateway forward for dev
+- 8001: Gateway port mapping
+- 8002: Keycloak port mapping
 
 ## Known Issues & Solutions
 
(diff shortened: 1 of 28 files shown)


david-martin:
> it used to work and stopped working this suddenly

I think it was the modelcontextprotocol/sdk package updating a type in the response object.

david-martin:
Discussed with @trepel on a meet.
This PR is good to merge.

There are 2 bugs to create, unrelated to these changes:

> Then I tried with OAuth (the bug in WASMSHIM was fixed in Kuadant v1.3.1). That worked as well, what surprised me a bit was that clicking "Clear" button invalidated the session and one had to click "Reconnect" button to make it work again.

Tomas will try this again from main with and without auth to reproduce & understand the extent of the issue, creating a bug after.

> I could not make it work using docker, the docker build  -t ghcr.io/kagenti/mcp-gateway:latest . is failing for me, but it's failing on main too so it's unrelated to this PR. Hence approving this one.

Tomas will create an issue for this. I was unable to reproduce this (I'm using docker via orbstack on macos). The build logs can be attached to the bug.

trepel:
@david-martin I created https://github.com/kagenti/mcp-gateway/issues/380 - it complains about invalid session ID even without auth configured (using `main` branch)

Issue #355, opened 2025-11-17
MCP Everything server seems to have invalid external HTTPRoute


esnible:
After installing the MCP Gateway dev env with `make local-env-setup`, `oc -n mcp-test get httproute mcp-everything-server-route-ext -o yaml` shows a problem:

```YAML
status:
  parents:
  - conditions:
    - lastTransitionTime: "2025-11-17T13:53:46Z"
      message: no hostnames matched parent hostname "server2.127-0-0-1.sslip.io";
        no hostnames matched parent hostname "server3.127-0-0-1.sslip.io"; no hostnames
        matched parent hostname "api-key-server.127-0-0-1.sslip.io"; no hostnames
        matched parent hostname "oidc-server.127-0-0-1.sslip.io"; no hostnames matched
        parent hostname "mcp.127-0-0-1.sslip.io"; no hostnames matched parent hostname
        "*.mcp.local"; no hostnames matched parent hostname "keycloak.127-0-0-1.sslip.io";
        no hostnames matched parent hostname "server1.127-0-0-1.sslip.io"
```

PR #356, opened 2025-11-17
Fix everything server hostname

david-martin:
Fix for #345 ~and possibly #355~


esnible (approved):
This works -- I can now invoke `everything_echo` through the gateway -- I tested interactively with the MCP Inspector.

This WILL NOT fix #355 because 355 is about the external direct route to everything, everything-server.127-0-0-1.sslip.io , not the gateway internal route.

Issue #157, opened 2025-09-10
Support the HTTP Delete method in the broker


david-martin:
Clients that no longer need a particular session (e.g., because the user is leaving the client application) SHOULD send an HTTP DELETE to the MCP endpoint with the Mcp-Session-Id header, to explicitly terminate the session.

We can either implement this or return a 405

We should remove the session-id from the cache when the delete comes in, we should also consider iterating over any backend mcp-session-ids and sending on the DELETE to those servers to allow them opportunity to clean up

david-martin:
Let's investigate if this is partially handled by the mcp-go library/server for invalidating the in memory session id.
We should also clean up the gateway to mcp backend sessions mappings when this happens

esnible:
The broker uses the mark3labs MCP server implementation with the default session handler, [_InsecureStatefulSessionIdManager_](https://github.com/mark3labs/mcp-go/blob/main/server/streamable_http.go#L908).  This session handler is "insecure" because it doesn't do anything other than verify the format of the sessionID.

The implementation doesn't write a secure stateful sessionID manager.  That implementation would either be memory-based (thus not scalable and not surviving pod restarts), or it would use shared K8s state (thus being slow and hard to test).

We "support" delete in that the broker returns 20x on delete, but we don't support it in the sense that a client can delete a sessionID and then continue to use it, or just forge sessionIDs.  The only clever thing is that the sessionIDs are UUIDs, and thus it is difficult to guess/forge another tenant's ID.

PR #361, opened 2025-11-19
Session mgmt revamp

### LINKED ISSUE #50: MCP Router: Handle 404 invalid session id response from upstream MCP Server
See https://modelcontextprotocol.io/specification/2025-06-18/basic/transports#session-management for how mcp servers are expected to handle an invalid session id.

We should call out to the ~~broker~~ **router** to invalidate the session id when this happens, then let the client re-initialize

maleck13:
closes #50 

- Always ensure we send back the gateway session id and not the mcp backend session id 
- Allow for using redis as an external store for client sessions
- Ensure when we initialize the MCP backend we use the client auth and not the broker auth
- Ensure we remove sessions when a client gateway session is invalidated
- Ensure we close any open connections when a client gateway session is invalidated

- Adds a redis backend for local dev for testing
- adds a custom response tool to test different responses coming through the gateway

**Still todo**

- [x]  Update the docs
- [x] Add some verification steps
- [x]  Add a flow diagram on the session flow now
- [ ]  Add some e2e test coverage
- [x]  Test with AuthPolicy etc 
- [x] Add issues for 1) invalidating gateway session 2) invoking DELETE against MCP servers when we receive a DELETE from the client Relates to #157 

### Validation

- Use `make local-env-setup`
- make `oauth-token-exchange-example-setup`
- use stern to monitor logs `stern broker -n mcp-system` and `stern mcp-oidc-server -n mcp-test`
- use `make inspect-gateway`
- Initilalize and tools/list (should get 3 tools)
- Invoke the OIDC tool:

In the logs of the MCP Gateway should should see as there is nothing in cache

```
initializing target as no mcp-session-id found for client

```

In the `mcp-oidc-server` logs you should see an initialize call with a JWT token. If you take that token to jwt.io you can see it is the scoped down token that was sent to the backend 

```
it will have an aud:  "aud": "oidc-server.mcp.local",
```

```
mcp-oidc-server-6f7b8d6b8c-q6k9b server 2025/11/21 11:00:43 	Header: X-Mcp-Method: initialize

```

You should also see a toos/call log that also has a scoped down token with the same aud

- Invoke the oidc tool again

you should see in the broker logs

```
found session in cache
```

In the oidc mcp logs you should see only one call come through now and it should be a tool call and it should have  scoped down token

```
mcp-oidc-server-6f7b8d6b8c-q6k9b server 2025/11/21 11:05:56 	Header: X-Mcp-Method: tools/call
```

Finally disconnect your client. And re-invoke the tool. You should see the same result as the first tool call

Note:  currently the config is sent many times to gateway components so you see it go a bit nuts when you run `oauth-token-exchange-example-setup` but it will settle down #364


config/samples/mcpserver-test-servers.yaml
```diff
@@ -104,11 +104,26 @@ metadata:
...
-    "kagenti/mcp": "true"
+    'kagenti/mcp': 'true'
 spec:
   toolPrefix: everything_
   targetRef:
     # MCP Everything Server - Typescript SDK based (prompts, tools, resources, sampling)
     group: gateway.networking.k8s.io
     kind: HTTPRoute
-    name: everything-server-route
+    name: mcp-everything-server-route
```
maleck13: bad rebase remove mcp-

internal/mcp-router/request_handlers.go
```diff
@@ -148,30 +160,70 @@ func (s *ExtProcServer) HandleMCPRequest(ctx context.Context, mcpReq *MCPRequest
...
+			for _, h := range mcpReq.Headers.Headers {
+				if h.Key == "Authorization" || h.Key == "x-request-id" {
+					passThroughHeaders[h.Key] = string(h.RawValue)
+				}
+			}
+		}
+		s.Logger.Debug("initializing target", "server ", serverInfo.Name, "with passthrough headers", passThroughHeaders)
+		clientHandle, err := s.InitForClient(ctx, s.RoutingConfig.MCPGatewayInternalHostname, s.RoutingConfig.RouterAPIKey, serverInfo, passThroughHeaders)
 		if err != nil {
-			s.Logger.Error("Failed to get session from cache", "error", err)
```
maleck13: close on error

internal/mcp-router/server.go
```diff
@@ -109,19 +97,10 @@ func (s *ExtProcServer) Process(stream extProcV3.ExternalProcessor_ProcessServer
 					}
 				}
 			}
-			if _, err := mcpRequest.Validate(); err != nil {
```
maleck13: why is it removed?

internal/session/cache.go
```diff
@@ -0,0 +1,98 @@
...
+	}
+	for _, opt := range opts {
+		opt(c)
+	}
+	if c.connectionString != "" {
+		opt, err := redis.ParseURL(c.connectionString)
+		if err != nil {
+			return c, err
+		}
+		c.extClient = redis.NewClient(opt)
```
maleck13: reconnects handled automatically?

maleck13: I think it is likely we might want to set some other settings but at the moment this will take a credential from the connection string and connect using that. We could do a ping as I don't think the new connection will error if not available

david-martin:
fixes #380

maleck13:
I think this has enough value to merge as is.. The current e2e test is a bit of a disaster. So will follow up with that as a separate PR as it also found quite a few more bugs

Issue #408, opened 2025-12-02
Only deploy everything server for local dev


david-martin:
The list of test servers deployed when do run `make local-env-setup` is ever growing (~8 currently), with more being added (see #400).
This is slowing down the dev environment setup time.
Let's only deploy 1 well rounded mcp server, i.e. the everything server, so there's something with a reasonable set of functionality available straight away.
Let's keep the `make deploy-test-servers` target for deploying all of the test servers if developers want.

PR #412, opened 2025-12-02
Remove keycloak/kuadrant/cert-manager install from default dev environment

david-martin:
Small quality of life change for local dev.
Speeds up local dev environment setup with `make local-env-setup` by no longer installing keycloak.

Keycloak, Kuadrant and cert-manager is installed if you run either of the `oauth-token-exchange-example-setup` or `oauth-acl-example-setup` targets, where it's a dependency.

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,7 +102,7 @@ Finally, open MCP Inspector at http://localhost:6274/?transport=streamable-http&
 
 When you click connect with MCP Inspector, you should be redirected to Keycloak. There you will need to login as the MCP user with password mcp. You now should only be able to access tools based on the ACL configuration.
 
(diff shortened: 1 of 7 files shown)


esnible (commented):
As part of this work, could the undocumented `oauth-acl-example-setup` be documented?

As part of this work, can you change _mcp-gateway/README.md_ to mention that `make oauth-token-exchange-example-setup` will also install Keycloak.

david-martin:
> We need to check what happens with the OIDC-enabled test MCP server, as it depends on Keycloak.

Good catch.

We could add keycloak as a dep to deploy-test-servers target as part of #408

PR #415, opened 2025-12-03
Add instructions on how to configure AuthPolicy to get external Github use case working again

huang195:
Fixes https://github.com/kagenti/mcp-gateway/issues/390. After adding the headers, I'm able to get github tool successfully initialize and run with an appropriate Github PAT.

image

### DIFF
--- a/config/samples/remote-github/authpolicy.yaml
+++ b/config/samples/remote-github/authpolicy.yaml
@@ -1,0 +1,20 @@
+apiVersion: kuadrant.io/v1
+kind: AuthPolicy
+metadata:
+  name: mcps-auth-policy
+  namespace: mcp-test
+spec:
+  targetRef:
+    group: gateway.networking.k8s.io
+    kind: HTTPRoute
+    name: github-mcp-external
+  rules:
+    response:
+      success:
+        headers:
+          authorization:
+            plain:
+              expression: 'request.headers["authorization"]'
+          x-mcp-api-key:
+            plain:
+              expression: 'request.headers["authorization"].split(" ")[1]'
--- a/config/samples/remote-github/create_resources.sh
+++ b/config/samples/remote-github/create_resources.sh
@@ -1,0 +1,85 @@
+#!/bin/bash
+set -e
+
+# check for GITHUB_PAT environment variable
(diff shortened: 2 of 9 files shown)


internal/mcp-router/request_handlers.go
```diff
@@ -257,6 +257,18 @@ func (s *ExtProcServer) HandleToolCall(ctx context.Context, mcpReq *MCPRequest)
 	headers.WithMCPSession(remoteMCPSeverSession)
 	// reset the host name now we have identifed the correct tool and backend
 	headers.WithAuthority(serverInfo.Hostname)
+
+	// add authorization headers for backends that require auth
+	mcpServerConfig := s.RoutingConfig.GetServerConfigByName(serverInfo.Name)
```
maleck13: I don't think we want to do this. The credential provided in the MCPServer resource is intended for use by the broker to initialise , list tools and listen for global notifications. With this change we allow an external client to use that credential. I think it would be better to use an authpolicy for the route that loads the needed credential during the tools call. This change will overwrite any Authorization set by the client by an AuthPolicy if the backend mcpserver has specified a credential.

See example here of fetching a key from vault https://github.com/kagenti/mcp-gateway/blob/main/config/samples/oauth-token-exchange/tools-call-auth.yaml#L18

internal/mcp-router/request_handlers.go
```diff
@@ -362,6 +374,20 @@ func (s *ExtProcServer) HandleNoneToolCall(mcpReq *MCPRequest) []*eppb.Processin
 
 			s.Logger.Debug("HandleMCPBrokerRequest initialize request", "target", remoteInitializeTarget, "call", mcpReq.Method)
 			headers.WithAuthority(remoteInitializeTarget)
+
+			// add authorization header if the server has credentials configured
+			serverConfig := s.RoutingConfig.GetServerConfigByHostname(remoteInitializeTarget)
```
maleck13: None tool calls don't ever his the backend MCPServer, they only hit the broker

maleck13:
@guicassolato perhaps you can help @huang195 with how he could configure an AuthPolicy to solve this ?

david-martin:
@maleck13 Good to merge now?

PR #432, opened 2025-12-08
Add vision doc

david-martin:
This captures what I see the role of the project being over the next ~6 to 12 months.

Longer term it's a bit tricker to put that in words, but I see the MCP Gateway still providing glue code via the mcp-broker, even if the ext-proc is needed less (or even disappears in time).
There may be cases where some of the abstractions added to the MCP Gateway don't make sense or align with the upstream, and should still be provided here.

### DIFF
--- a/VISION.md
+++ b/VISION.md
@@ -57,7 +57,7 @@ Success is measured by reduced duplication across MCP servers, consistent policy
 Note, these are just examples at the time of writing, and are not intended to be goals.
 
 *Example 1:* The MCP Gateway project will provide a Kubernetes CRD, like MCPServer, that represents a running MCP Server somewhere.
-This CRD is expected to converge with or be replaced by a 'Backend' resource that is aligned with the outcome of the ai-gatweay working group or agentic-networking sub-project. This Backend resource may get implemented in the Gateway API provider, Istio, in time.
+This CRD is expected to converge with or be replaced by a 'Backend' resource that is aligned with the outcome of the ai-gateway working group or agentic-networking sub-project. This Backend resource may get implemented in the Gateway API provider, Istio, in time.
 
 *Example 2:* The MCP Gateway project has some very specific Kuadrant AuthPolicy examples around tool permissions based on integrating with Keycloak. The MCP Gateway project will provide a KeycloakToolRoleMappingPolicy that wraps an AuthPolicy, abstracting the detailed rules configuration required to parse and iterate tools, checking against Keycloak role mappings.
 
--- a/.cspell.json
+++ b/.cspell.json
@@ -1,5 +1,5 @@
 {
-	"language": "en",
+	"language": "en,en-GB",
 	"dictionaryDefinitions": [{
 		"name": "project-words",
 		"path": "./project-words.txt",
--- a/project-words.txt
+++ b/project-words.txt
@@ -1,4 +1,5 @@
 # New Words
+Agentic
 apikey
 authorino
 Authorino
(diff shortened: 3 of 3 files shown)


VISION.md
```diff
@@ -0,0 +1,38 @@
...
+When standards firm up, MCP Gateway adopts them. Until then, teams get working infrastructure today.
+
+### What does this mean in practice?
+
+Note, these are just examples at the time of writing, and are not intended to be goals.
+
+*Example 1:* The MCP Gateway project will provide a kubernetes CRD, like MCPServer, that represents a running MCP Server somewhere.
+This CRD may eventually evolve into or be replaced by a 'Backend' resource that is aligned with the outcome of the ai-gatweay working group or agentic-networking sub-project. This Backend resource may get implemented in the Gateway API provider, Istio, in time.
+
+*Example 2:* The MCP Gateway project has some very specific Kuadrant AuthPolicy examples around tool permissions based on integrating with Keycloak. The MCP Gateway project will provide a KeycloakToolRoleMappingPolicy that wraps an AuthPolicy, abstracting the detailed rules configuration required to parse and iterate tools, checking against keycloak role mappings.
```
esnible:
```suggestion
*Example 2:* The MCP Gateway project has some very specific Kuadrant AuthPolicy examples around tool permissions based on integrating with Keycloak. The MCP Gateway project will provide a KeycloakToolRoleMappingPolicy that wraps an AuthPolicy, abstracting the detailed rules configuration required to parse and iterate tools, checking against Keycloak role mappings.
```

david-martin:
@maleck13 @esnible @shaneutt I've pushed a few updates based on comment feedback,
and added a 'why not...' section to highlight at a high level how the MCP Gateway project is different than other solutions.

VISION.md
```diff
@@ -0,0 +1,87 @@
+# MCP Gateway Vision
+
+## Problem Statement
+
+Platform teams want to expose MCP servers to agents safely, but today this requires bespoke routing, auth, policy, and observability per server. There is no standard entrypoint, no shared policy model, and no production-grade way to operate MCP at scale.
+
+## Solution
+
+MCP Gateway is a production-grade MCP entrypoint that lets platform teams operate many MCP servers with shared routing, policy, and observability, while tracking upstream standards instead of inventing new ones.
```
maleck13: I am not sure I would call this out as production grade yet. Maybe `will be` ?

david-martin:
> * one of the main differentiators of our MCP Gateway is that we build on top of a mature proxy, Envoy. Do we want to give a few links on how it is battle tested, performant, scalable, etc.?

I think this is a good idea, if you have some link(s) in mind?

> can we be a bit more explicit on what it means by "...outside native Envoy mechanisms"

Updated to say "...outside native Envoy mechanisms like routing and filters."

> when comparing with Envoy AI Gateway, we say it's Go-based, thus, limiting reuse by downstream extensions. Ours is also based on Go, so what do we mean by this?

Maybe the wording can change. My intention was to focus on it being a new proxy, as opposed to it being a go proxy. As much as possible, in our MCP Gateway we try delegate to Envoy where it's best at solving a problem (like routing and filtering, for the actual tool calls and for auth). We use our go code when it's not possible to use more natively with envoy (for tool list and init), but we don't try to act as a proxy.
Is there a rewording you think would help convey this?

> Do we want to explicitly call out Python's performance limitations in high traffic network proxy or is this well understood in the field?

Like the first point, I think this is a reasonable thing to highlight as it could be an important factor is deciding if someone will choose our mcp gateway or not.

PR #434, opened 2025-12-09
docs: Add design document for tool prefix matching

trepel:
Documents validation approach for empty-prefix MCPServers and router changes to prioritize prefix matching over empty-prefix MCPServer fallback.

### DIFF
--- a/docs/design/tool-prefix-matching.md
+++ b/docs/design/tool-prefix-matching.md
@@ -17,15 +17,19 @@ MCPServer A: toolPrefix="" (empty)
 MCPServer B: toolPrefix="test_"
 
 Tool call: "test_tool_from_B"
-Current behavior: Routes to A (wrong)
-Expected behavior: Routes to B (correct)
+Current behavior: Routes to A if A is checked before B (wrong)
+Expected behavior: Always routes to B (correct)
 ```
 
 ### Solution
 
 Enforce single empty-prefix server and prioritize prefix matching:
 
-1. **Validation**: Only one MCPServer with an empty toolPrefix is allowed per namespace. Additional MCPServers with empty prefix fail validation with clear error message.
+1. **Validation**: Only one MCPServer with an empty toolPrefix is allowed. Additional MCPServers with empty prefix fail validation with clear error message.
+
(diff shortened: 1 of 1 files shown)


docs/design/tool-prefix-matching.md
```diff
@@ -0,0 +1,53 @@
...
+Tool call: "test_tool_from_B"
+Current behavior: Routes to A (wrong)
+Expected behavior: Routes to B (correct)
+```
+
+### Solution
+
+Enforce single empty-prefix server and prioritize prefix matching:
+
+1. **Validation**: Only one MCPServer with an empty toolPrefix is allowed per namespace. Additional MCPServers with empty prefix fail validation with clear error message.
```
david-martin: what is 'namespace' in this context?
A kubernetes namespace?

esnible: Why is only one empty toolPrefix allowed per namespace?

I almost **never** use a prefix, because I don't want to rewrite my agents to use different names.

david-martin: The current implementation uses the tool prefix to determine which server to route a tool call to.
If there's more than 1 mcp server without a tool prefix, there's no way to know which server a tool call should be routed to.

> it is possible to have more MCP Gateways deployed on cluster so there could be multiple MCPServer CRs having empty prefix as long as each is registered to a different MCP Gateway

Understood. That's a good point for consideration in #436 
As things stand, only 1 gateway instance is supported.

david-martin:
On hold pending https://github.com/Kuadrant/mcp-gateway/pull/451

maleck13:
I think this has changed quite a bit now. 

1) We have conflict detection so it is a first come first served situation
2) We want to route based on the full tool name without prefix. 
3) As an option, the prefix can be added for conflict resolution 

I think it makes sense to close this and update the https://github.com/Kuadrant/mcp-gateway/blob/main/docs/design/routing.md doc to go into more details about tool call routing

trepel:
@shaneutt @maleck13 I agree, I am closing this one. As @david-martin said this is on hold anyway due to ongoing effort in https://github.com/Kuadrant/mcp-gateway/pull/451 At this point it is clear that routing is about to be reworked significantly so updating routing.md makes more sense.

Issue #426, opened 2025-12-07
MCPserver status not correctly set


huang195:
After installing the github external mcp server CR, this shows up in its status

```
  status:
    conditions:
    - lastTransitionTime: "2025-12-05T18:24:23Z"
      message: MCPServer successfully reconciled and validated 0 servers with 0 tools
      reason: Ready
      status: "True"
      type: Ready
```

However, in the broker pod's log, we see that it successfully discovered 40 tools

```
time=2025-12-07T13:51:03.435Z level=INFO msg="OnNotification Re-Discovered tools" mcpURL=https://api.githubcopilot.com:443/mcp #tools=40
```

huang195:
It doesn't seem to happen consistently. Sometimes it works:

```
status:
  conditions:
  - lastTransitionTime: "2025-12-07T14:47:03Z"
    message: MCPServer successfully reconciled and validated 1 servers with 40 tools
    reason: Ready
    status: "True"
    type: Ready
  discoveredTools: 40
```

maleck13:
Fixed by #448

PR #447, opened 2025-12-15
Add more validation logging

esnible:
For #426

This does not fix the problem.  It merely adds more debugging logging, of the kind I found useful when I was troubleshooting the problem.

I propose we merge this PR, and then go back to fixing the underlying problem.


esnible:
@david-martin I am thinking of closing this unless you think it should be merged quickly.

PR #451, opened 2025-12-17
Route using complete toolname

### LINKED ISSUE #449: Including an MCPServer with no prefix causes incorrect routing and undeterministic behaviour
When an MCPServer has no tool prefix, there is undeterministic behaviour.
This is due to how the server matching for a tool name is looked up when a tool call happens.
There are 2 places where this happens, in https://github.com/kagenti/mcp-gateway/blob/8665eaaa9ba666a961e0d9424c8549d1356ecf30/internal/mcp-router/request_handlers.go#L192 and https://github.com/kagenti/mcp-gateway/blob/8665eaaa9ba666a961e0d9424c8549d1356ecf30/internal/mcp-router/request_handlers.go#L226.
As the servers are iterated through, if a match is found on the tool prefix, that server is used for routing.
However, if there is no prefix defined for a server (i.e. an empty string), it will match every tool name.
So if that server is the first in the iteration, it will always match.
If the non prefix server is last in the iteration, then the behaviour is as expected.
If the non prefix server is somewhere in the middle of the iteration, some tools may get routed correctly, and some may not.

A simple fix for this would be to treat a non prefix server differently, and always have it be the last server to check, if no other prefix match was found.

NOTE: This is different from (but related to) the known problem where currently you can't have more than 1 MCPServer without a tool prefix due to matching on the first server.
See related proposal and discussion in #434 for additional context and a proposal for how this feature should behave.

esnible:
Resolves #449 

This PR causes the router to take advantage of the complete name of the tool when looking up the upstream.

This PR is based on @maleck13 's PR, rather than _main_.

cc @david-martin

### DIFF
--- a/internal/mcp-router/request_handlers.go
+++ b/internal/mcp-router/request_handlers.go
@@ -194,8 +194,23 @@ func (s *ExtProcServer) HandleToolCall(ctx context.Context, mcpReq *MCPRequest) 
 	headers := NewHeaders()
 	serverInfo, err := s.Broker.GetServerInfo(toolName)
 	if err != nil {
+		// For unknown tool, the spec says to return a JSON RPC error response,
+		// and the error is not an HTTP error, so we return a 200 status code.
+		// See https://modelcontextprotocol.io/specification/2025-06-18/server/tools#error-handling
 		s.Logger.Debug("no server for tool", "toolName", toolName)
-		calculatedResponse.WithImmediateResponse(400, "unknown tool")
+		calculatedResponse.WithImmediateJSONRPCResponse(200,
+			[]*corev3.HeaderValueOption{
+				{
+					Header: &corev3.HeaderValue{
+						Key:   "mcp-session-id",
+						Value: mcpReq.GetSessionID(),
+					},
+				},
+			},
+			`
+event: message
+data: {"result":{"content":[{"type":"text","text":"MCP error -32602: Tool not found"}],"isError":true},"jsonrpc":"2.0"}`)
+		return calculatedResponse.Build()
 	}
 	if annotations, hasAnnotations := s.Broker.ToolAnnotations(serverInfo.ID(), toolName); hasAnnotations {
 		// build header value (e.g. readOnly=true,destructive=false,openWorld=true)
--- a/tests/e2e/commons.go
+++ b/tests/e2e/commons.go
@@ -346,13 +346,13 @@ type MCPServerRegistrationBuilder struct {
 	mcpServer     *mcpv1alpha1.MCPServer
 }
 
-// NewMCPServerRegistration creates a new registration builder with defaults
-func NewMCPServerRegistration(testName string, k8sClient client.Client) *MCPServerRegistrationBuilder {
-	return NewMCPServerRegistrationEx(testName, "e2e-server2.mcp.local", "mcp-test-server2", 9090, k8sClient)
+// NewMCPServerRegistrationWithDefaults creates a new registration builder with defaults
+func NewMCPServerRegistrationWithDefaults(testName string, k8sClient client.Client) *MCPServerRegistrationBuilder {
+	return NewMCPServerRegistration(testName, "e2e-server2.mcp.local", "mcp-test-server2", 9090, k8sClient)
 }
 
-// NewMCPServerRegistration creates a new registration builder with defaults
-func NewMCPServerRegistrationEx(testName, hostName, serviceName string, port int32, k8sClient client.Client) *MCPServerRegistrationBuilder {
+// NewMCPServerRegistration creates a new registration builder
+func NewMCPServerRegistration(testName, hostName, serviceName string, port int32, k8sClient client.Client) *MCPServerRegistrationBuilder {
 	httpRoute := BuildTestHTTPRoute("e2e-server2-route-"+testName, TestNamespace,
 		hostName, serviceName, port)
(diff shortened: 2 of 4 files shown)


internal/broker/broker.go
```diff
@@ -222,6 +226,25 @@ func (m *mcpBrokerImpl) ToolAnnotations(serverID config.UpstreamMCPID, tool stri
...
+			slog.Info("[EXT-PROC] Found matching server",
+				"toolName", tool,
+				"serverPrefix", upstream.MCP.GetPrefix(),
+				"serverName", upstream.MCP.GetName())
+			retval := upstream.MCP.GetConfig()
+			return &retval
+		}
+	}
+
+	slog.Info("Tool name doesn't match any configured server prefix", "tool", tool)
```
maleck13:
```suggestion
	slog.Info("Tool name doesn't match any configured server", "tool", tool, "prefix",upstream.MCP.GetPrefix())
```

internal/mcp-router/request_handlers.go
```diff
@@ -189,36 +189,34 @@ func (s *ExtProcServer) HandleToolCall(ctx context.Context, mcpReq *MCPRequest)
 		calculatedResponse.WithImmediateResponse(404, "session no longer valid")
 		return calculatedResponse.Build()
 	}
-	serverInfo := s.RoutingConfig.GetServerInfo(toolName)
-	if serverInfo == nil {
-		s.Logger.Info("Tool name doesn't match any configured server prefix", "tool", toolName)
-		calculatedResponse.WithImmediateResponse(404, "not found")
-		return calculatedResponse.Build()
+
+	if s.Broker == nil {
```
maleck13: I think we can move this in to main under setUpRouter

internal/mcp-router/request_handlers.go
```diff
@@ -189,36 +189,34 @@ func (s *ExtProcServer) HandleToolCall(ctx context.Context, mcpReq *MCPRequest)
...
-			var parts []string
-			push := func(key string, val *bool) {
-				if val == nil {
-					parts = append(parts, fmt.Sprintf("%s=unspecified", key))
-				} else if *val {
-					parts = append(parts, fmt.Sprintf("%s=true", key))
-				} else {
-					parts = append(parts, fmt.Sprintf("%s=false", key))
-				}
+	serverInfo := s.Broker.GetServerInfo(toolName)
```
maleck13: are we missing a nil check here? Consider having GetServerInfo return an error with some useful context that we can then log?

maleck13 (commented):
some small changes and checks suggested. Also it would be good to expand the e2e coverage to cover MCPServers created with no prefix along side MCPServers created with a prefix that share tool names.

tests/e2e/commons.go
```diff
@@ -348,8 +348,13 @@ type MCPServerRegistrationBuilder struct {
 
 // NewMCPServerRegistration creates a new registration builder with defaults
 func NewMCPServerRegistration(testName string, k8sClient client.Client) *MCPServerRegistrationBuilder {
+	return NewMCPServerRegistrationEx(testName, "e2e-server2.mcp.local", "mcp-test-server2", 9090, k8sClient)
+}
+
+// NewMCPServerRegistration creates a new registration builder with defaults
+func NewMCPServerRegistrationEx(testName, hostName, serviceName string, port int32, k8sClient client.Client) *MCPServerRegistrationBuilder {
```
maleck13: lets rename this to `NewMCPServerRegistration` and the old one to `NewMCPServerRegistrationWithDefaults`

tests/e2e/happy_path_test.go
```diff
@@ -709,4 +709,65 @@ var _ = Describe("MCP Gateway Registration Happy Path", func() {
...
+		Eventually(func(g Gomega) {
+			toolsList, err := mcpGatewayClient.ListTools(ctx, mcp.ListToolsRequest{})
+			g.Expect(err).Error().NotTo(HaveOccurred())
+			g.Expect(toolsList).NotTo(BeNil())
+			g.Expect(verifyMCPServerToolPresent("echo", toolsList)).To(BeTrueBecause("%q should exist", "greet"))
+			g.Expect(verifyMCPServerToolPresent("hello_world", toolsList)).To(BeTrueBecause("%q should exist", "hello_world"))
+		}, TestTimeoutLong, TestRetryInterval).To(Succeed())
+
+		toolName := "hello_world"
+		GinkgoWriter.Println("tool", toolName)
```
maleck13: Not sure this is adding anything?

internal/mcp-router/request_handlers.go
```diff
@@ -189,36 +189,34 @@ func (s *ExtProcServer) HandleToolCall(ctx context.Context, mcpReq *MCPRequest)
...
-				if val == nil {
-					parts = append(parts, fmt.Sprintf("%s=unspecified", key))
-				} else if *val {
-					parts = append(parts, fmt.Sprintf("%s=true", key))
-				} else {
-					parts = append(parts, fmt.Sprintf("%s=false", key))
-				}
+	serverInfo, err := s.Broker.GetServerInfo(toolName)
+	if err != nil {
+		s.Logger.Debug("no server for tool", "toolName", toolName)
```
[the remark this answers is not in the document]
esnible: After reading https://modelcontextprotocol.io/specification/2025-06-18/server/tools#error-handling I have revised this code to return 200 and a JSON-RPC status code.

PR #478, opened 2026-01-12
Move references to Kuadrant org

### LINKED ISSUE #446: After moving the repo to Kuadrant, some references still point to the kagenti org
**Describe the bug**
After moving the repo to Kuadrant, some references still point to the kagenti org

**To Reproduce**
1. For example, look at at the quick install in the readme file that uses `https://github.com/kagenti` :
```
kubectl apply -k 'https://github.com/kagenti/mcp-gateway/config/install?ref=main'
```

**Expected behavior**
Should be self-contained with references pointing to the new kuadrant org

esnible:
Resolves #446

This PR does both the documentation changes and the Go module name.  The Go module name change caused a lot of tiny changes in a lot of files, but should make it easier for someone looking at the code to find the project.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -11,12 +11,6 @@ We can't wait to collaborate with you!
 Please feel free to report an issue for bugs, and any other problems you
 encounter with the MCP Gateway project.
 
-If you have something to propose (e.g. a big new feature) we recommend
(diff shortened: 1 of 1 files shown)


david-martin (approved):
Thanks for this @esnible 
Merging.

PR #501, opened 2026-01-16
add test for resolving conflict via prefix

### LINKED ISSUE #496: Add e2e test that covers clashing tools and resolving with prefix
We have a test for multiple servers with  no prefix set but we are  missing a test that exercises the following path

- add 2 servers with a common tool
- observe a conflict
- change the clashing server to have a prefix
- observe no conflict
- invoke both tools with a client to see they are routed correctly

maleck13:
closes #496 

This test also found a bug with invoking tools are using a prefix, that issue has also been fixed and the test ensures we avoid regression. 

The number of happy path e2e tests has increased quite a lot so I have bumped the timeout also.

### DIFF
--- a/tests/e2e/happy_path_test.go
+++ b/tests/e2e/happy_path_test.go
@@ -851,60 +851,10 @@ var _ = Describe("MCP Gateway Registration Happy Path", func() {
 
 		// Call server2's prefixed headers tool
 		res, err = mcpGatewayClient.CallTool(ctx, mcp.CallToolRequest{
-			Params: mcp.CallToolParams{Name: "server2_headers"},
+			Params: mcp.CallToolParams{Name: "server2_time"},
 		})
 		Expect(err).NotTo(HaveOccurred())
 		Expect(res).NotTo(BeNil())
 		Expect(len(res.Content)).To(BeNumerically(">=", 1))
-	})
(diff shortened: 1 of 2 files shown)


tests/e2e/happy_path_test.go
```diff
@@ -785,4 +784,127 @@ var _ = Describe("MCP Gateway Registration Happy Path", func() {
...
+		res, err := mcpGatewayClient.CallTool(ctx, mcp.CallToolRequest{
+			Params: mcp.CallToolParams{Name: "time"},
+		})
+		Expect(err).NotTo(HaveOccurred())
+		Expect(res).NotTo(BeNil())
+		Expect(len(res.Content)).To(BeNumerically(">=", 1))
+
+		// Call server2's prefixed headers tool
+		res, err = mcpGatewayClient.CallTool(ctx, mcp.CallToolRequest{
+			Params: mcp.CallToolParams{Name: "server2_headers"},
```
david-martin: would it better to call the tool that had the conflict on both servers to test that it works on both?

maleck13: Yes. Good spot. When debugging the issue I modified this to see was there any difference but didn't switch it back

PR #509, opened 2026-01-19
Release updates for new image (mcp-controller) and change to kuadrant org

david-martin:
Needed before a new release can be done

### DIFF
--- a/.github/workflows/images.yaml
+++ b/.github/workflows/images.yaml
@@ -16,15 +16,21 @@ concurrency:
 
 env:
   REGISTRY: ghcr.io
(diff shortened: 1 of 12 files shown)


.github/workflows/images.yaml
```diff
@@ -42,7 +48,7 @@ jobs:
         id: meta
         uses: docker/metadata-action@v5
         with:
-          images: ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}
+          images: ${{ env.REGISTRY }}/${{ github.repository_owner }}/${{ matrix.image }}
```
david-martin: this change should build and push an image with the same tag as the repo tag

PR #516, opened 2026-01-21
Add releasing instructions

### LINKED ISSUE #332: Fix and document how to release
Let's capture the right steps to do a release of the mcp-gateway.
If there are manual steps, let's capture how we could move towards automating them as well.

Right now, here's what I've done to release:

* Create a new release from main in https://github.com/kagenti/mcp-gateway/releases
* Run the helm release action, using that tag, at https://github.com/kagenti/mcp-gateway/actions/workflows/helm-release.yaml

We missed a change in the openshift deploy script here #331 
Can we automate this?

Any other considerations for release?

What about doing a patch release?

david-martin:
Closes #332

### DIFF
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -1,0 +1,50 @@
+# Releasing MCP Gateway
+
+Creating a GitHub release triggers automated workflows that:
+1. Build and push container images (`mcp-gateway`, `mcp-controller`) to `ghcr.io/kuadrant/`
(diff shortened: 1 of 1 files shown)


trepel (approved):
Based on the `git help commit` my understanding was that both `-s` and `-S` need to be used for `git commit` but looking at my own alias I am using just `-s` myself so this looks good. LGTM.

PR #544, opened 2026-01-29
create the gateway-system ns

maleck13:


config/openshift/deploy_openshift.sh
```diff
@@ -49,6 +49,7 @@ helm upgrade -i mcp-gateway -n $MCP_GATEWAY_NAMESPACE --create-namespace oci://g
 
 # Configure MCP Gateway Ingress using Helm
 echo "Configuring MCP Gateway Ingress using Helm..."
+kubectl create ns $GATEWAY_NAMESPACE
```
trepel: I think that this need to go above line 47

PR #564, opened 2026-02-04
set workload selector. Remove duplicate templates

maleck13:
@trepel this should resolve the issue you were seeing on OpenShift


trepel:
@maleck13 what do you think about updating the default value to match gateway name here: https://github.com/Kuadrant/mcp-gateway/blob/main/charts/mcp-gateway/values.yaml#L31-L42
It makes sense to me if these two must match

PR #577, opened 2026-02-05
Remove spaces causing helm install failure

david-martin:
Fix for this error when running the `charts/sample_local_helm_setup.sh` script.

```
Error: INSTALLATION FAILED: expected at most two arguments, unexpected arguments
```


esnible (approved):
Thanks!

PR #606, opened 2026-02-23
Helm test

maleck13:
add a simple helm test to catch issues with the helm install. 

Uses a new kind cluster
Executes the helm install command
Verifies the resources reach the correct state
Will be triggered on each PR


charts/mcp-gateway/templates/_helpers.tpl
```diff
@@ -57,7 +57,7 @@ app.kubernetes.io/instance: {{ .Release.Name }}
 Create the name of the broker service account to use
 */}}
 {{- define "mcp-gateway.brokerServiceAccountName" -}}
-{{ include "mcp-gateway.fullname" . }}-broker-router
+{{ include "mcp-gateway.fullname" . }}-mcp-gateway
```
david-martin: will this result in a service account name of `mcp-gateway-mcp-gateway`?

PR #619, opened 2026-02-26
Operator-based deployment with automatic HTTPRoute creation

maleck13:
- Controller auto-creates an HTTPRoute (mcp-gateway-route) for gateway access; opt out with
  kuadrant.io/alpha-disable-httproute: "true" annotation for custom routes
  - Removes the httpRoute.create Helm flag since the controller now owns HTTPRoute lifecycle

Note the annotations are going to be replaced with a spec in near future. Want to propose a spec first before implementing

## Summary by CodeRabbit

* **New Features**
  * MCPGatewayExtension now auto-creates an HTTPRoute named mcp-gateway-route when ready.
  * New annotation to opt out of automatic HTTPRoute creation.

* **Improvements**
  * Helm install flows no longer require explicit HTTPRoute flags — controller manages route creation.
  * RBAC updated so the controller can manage HTTPRoutes and read namespaces.

* **Documentation**
  * Migration guide and install/config docs updated to explain auto-route behavior and customization.

* **Tests**
  * New unit and e2e tests covering HTTPRoute creation, disablement, and multi-extension scenarios.

### DIFF
--- a/internal/controller/mcpgatewayextension_controller.go
+++ b/internal/controller/mcpgatewayextension_controller.go
@@ -262,9 +262,9 @@ func (r *MCPGatewayExtensionReconciler) validateGatewayTarget(ctx context.Contex
 				mcpExt.Spec.TargetRef.SectionName, targetGateway.Namespace, targetGateway.Name, mcpv1alpha1.AnnotationPublicHost))
 	}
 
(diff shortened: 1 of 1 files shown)


maleck13:
looking into the CI failures

PR #620, opened 2026-02-27
Fix mcp-gateway deployment name

david-martin:
Deployment renamed from mcp-broker-router to mcp-gateway in previous PR.
This change fixes some cmd references to that deployment


maleck13 (approved):
I think these changes might also be in my pr but ok to merge

PR #640, opened 2026-03-06
Ext proc improvements

### LINKED ISSUE #317: [Tech Debt] Handle Errors correctly in the ext proc server
**what**

In the POC error handling was not a priority. This has led to several places where error handling is not ideal. We should have a clear pattern for handling errors in the ext_proc process method and understand what returning an error here will mean to the client and envoy

https://github.com/kagenti/mcp-gateway/blob/main/internal/mcp-router/server.go#L73 as an example of this

maleck13:
Closes #317 

Improve the error handling in the ext_proc server. Ensuring to avoid issues with the body and potential panics 

1. Empty request body — guarded against, although an unlikely occurrence 
  2. Unmarshal failure — was sending a do-nothing response then falling through to validation (double response, protocol
  violation). Now returns ImmediateResponse(400) and continues.
  3. Nil request headers — was returning error (generic Envoy 500). Now sends ImmediateResponse(500) before returning.
  4. Body before headers — replaced dead code block (redundant nil checks, misleading error messages) with a clean
  ImmediateResponse(500) and return.
  6. EndOfStream — captures EndOfStream flag from request headers. If the body phase arrives despite EndOfStream=true, sends  do-nothing response and continues instead of trying to parse. If “true“, then there is no message body associated with this request or response.
  7. Removed streaming variable — was always false, replaced with endOfStream. Streaming responses are handled by the backend mcp server

## Summary by CodeRabbit

* **Bug Fixes**
  * Consistent immediate 500 responses for missing or malformed request headers/body.
  * Avoids processing body after end-of-stream and sends do-nothing responses when appropriate.
  * Improved handling of early/empty body arrivals to prevent protocol errors and ensure correct response flows.

* **Tests**
  * Split end-to-end tests into focused cases covering invalid bodies, header errors, end-of-stream, send/receive failures, and span lifecycle.
  * Added deterministic failure simulation and strict response-consumption verification.

* **Chores**
  * Minor update to project word list.

### DIFF
--- a/internal/mcp-router/server_test.go
+++ b/internal/mcp-router/server_test.go
@@ -123,7 +123,7 @@ func TestProcess_HappyPath(t *testing.T) {
 				},
 			},
 		},
-		responseHeadersStep("200"),
+		responseHeadersStep(),
 	})
 
 	err := srv.Process(mock)
@@ -154,7 +154,7 @@ func TestProcess_EmptyBody(t *testing.T) {
 				},
 			},
 		},
-		responseHeadersStep("200"),
+		responseHeadersStep(),
 	})
 
 	err := srv.Process(mock)
@@ -179,7 +179,7 @@ func TestProcess_UnmarshalError(t *testing.T) {
 				immediateResponse(400),
 			},
 		},
-		responseHeadersStep("200"),
+		responseHeadersStep(),
 	})
 
 	err := srv.Process(mock)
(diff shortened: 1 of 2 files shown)


david-martin:
@bentito I didn't realise @maleck13 was working on this issue as well (#317) which was assigned to you.
If you've made a start on it, perhaps you can help with the review here and any other suggestions you may have based on how you were approaching it?

Issue #349, opened 2025-11-14
Elicitation Support


david-martin:
Jira: https://issues.redhat.com/browse/CONNLINK-807

Implement support for Elicitations, as per the design doc in https://github.com/kagenti/mcp-gateway/pull/315

Summary of changes needed:

* update the client initialize logic in the mcp-broker to send the 'elicitations' capabilities flag
* intercept elicitation/create messages sent back over the event-stream response in the mcp-router, and map the 'request id' field to include a unique server prefix. Note this will require switching the response body mode to streamed. Let's be careful doing this that we don't break tool progress updates (see #328 )
* as elicitation responses are sent by the client in a subsequent POST request, intercept and map back to the original request id, and route to the correct mcp server

david-martin:
Updates to the design doc to reflect the current level of support around notifications #612

PR #643, opened 2026-03-06
feat(elicitation): handle elicitation request and response flow

Cali0707:
Part of #349 

This adds support for streaming elicitation requests to the client and routing elicitation responses to the backend MCP servers

I will add e2e tests in a follow up PR

## Summary by CodeRabbit

* **New Features**
  * Elicitation flows: clients can opt in; router streams and rewrites SSE elicitations and routes elicitation responses back to the originating backend.
  * External processor config now allows mode overrides for flexible processing.

* **Tests**
  * Expanded test suite covering SSE rewriting, elicitation parsing, mapping, routing, streaming, and session lifecycle.

* **Chores**
  * Added pluggable ID map backends (in-memory and Redis) and session-level elicitation flag storage/cleanup.

### DIFF
--- a/tests/e2e/elicitation_test.go
+++ b/tests/e2e/elicitation_test.go
@@ -1,0 +1,201 @@
+//go:build e2e
+
+package e2e
+
+import (
+	"context"
+	"fmt"
+	"strings"
+
+	mcpclient "github.com/mark3labs/mcp-go/client"
+	"github.com/mark3labs/mcp-go/mcp"
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// acceptHandler returns an elicitation handler that accepts with provided content
+type acceptHandler struct {
+	content map[string]any
+}
+
+func (h *acceptHandler) Elicit(_ context.Context, req mcp.ElicitationRequest) (*mcp.ElicitationResult, error) {
+	GinkgoWriter.Printf("elicitation request received: %s\n", req.Params.Message)
+	return &mcp.ElicitationResult{
+		ElicitationResponse: mcp.ElicitationResponse{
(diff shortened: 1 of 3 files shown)


internal/mcp-router/server.go
```diff
@@ -178,6 +183,21 @@ func (s *ExtProcServer) Process(stream extProcV3.ExternalProcessor_ProcessServer
 			statusCode := getSingleValueHeader(r.ResponseHeaders.Headers, ":status")
 			span.SetAttributes(attribute.String("http.status_code", statusCode))
 
+			if mcpRequest != nil && mcpRequest.isToolCall() {
+				clientElicitation, elErr := s.SessionCache.GetClientElicitation(ctx, mcpRequest.GetSessionID())
```
maleck13: consider / add follow on for perhaps adding a property to the MCPRequest for "elicitationSupported" ?

At the moment we might hit the cache twice here rather than once

maleck13 (commented):
I think we are pretty close now. I think the double redis connection is the main thing and ideally we wouldn't hit the cache twice for checking if the elicitation is supported

maleck13:
@Cali0707 So this is looking good now. Thanks for all the work so far. There is one thing more which is a e2e. We have a slash command for helping with that here: https://github.com/Kuadrant/mcp-gateway/blob/main/.claude/commands/e2e-tests.md . We are currently defining our e2e test cases in in a file here https://github.com/Kuadrant/mcp-gateway/blob/main/tests/e2e/test_cases.md . Let me know if you need any help.

maleck13:
@Cali0707 those look great thanks. I tested this locally with MCP inspector and it worked as expected. Tried also with Claude code but it didn't work. I am not putting that down to the implementation though. The tests also looks good. Have you tried this with claude. It says it supports elicitation but it doesn't seem to work from the terminal.

PR #652, opened 2026-03-10
Update bundle and catalog to be built for amd64 & arm64

Patryk-Stefanski:
Adds multiplatform build for linux/amd64,linux/arm64

## Summary by CodeRabbit

* **Chores**
  * Enhanced CI/CD pipeline with improved Docker build and deployment process.
  * Added multi-architecture support (amd64 and ARM64) for container images enabling broader platform compatibility.
  * Streamlined and simplified the workflow for building and pushing bundle and catalog container images.
  * Improved build reliability and efficiency through updated build infrastructure tooling.

### DIFF
--- a/.github/workflows/images.yaml
+++ b/.github/workflows/images.yaml
@@ -165,16 +165,19 @@ jobs:
             ORG=${{ steps.owner.outputs.name }} \
             IMAGE_TAG=${{ github.ref_name }}
 
+      - name: Set up Docker Buildx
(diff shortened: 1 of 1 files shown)


trepel:
Changes look good to me, just not sure how to actually try because it is only triggered on push.

Also, I think that OLM bundle image does not have to be multiarch manifest, it does not contain anything platform specific. I guess it does not cause any harm either though

PR #701, opened 2026-03-30
409 fix inspect server targets

### LINKED ISSUE #409: Fix inspect-server-template targets
There are several `inspect-server` targets in https://github.com/kagenti/mcp-gateway/blob/main/build/inspect.mk
I think they are meant to open the mcp inspector with the external route to those servers, but they seem broke.

~Let's remove them as I don't see any issue about them being broker, so they likely aren't being used.~

Currently gives this output:

```
make inspect-server1
make: Nothing to be done for `inspect-server1'.
```

Let's fix these and extend them to create an external httproute for accessing them directly (note that external routes for test servers are to be removed in #410)

We should also have some cleanup target to 'uninspect' (remove the httproute and any other resources).

### LINKED ISSUE #410: Remove external httproutes for all test mcp servers
Let's remove all the external mcp server HTTPRoutes like https://github.com/kagenti/mcp-gateway/blob/main/config/test-servers/server1-httproute-ext.yaml

I don't think they are being used.
I believe they could cause confusion as they aren't actually required for an MCP Server to function behind the mcp-gateway.
The point of confusion could arise as there *is* an HTTPRoute required for internal routing with an arbitrary hostname like mysever.mcp.local.

jasonmadigan:
Re: https://github.com/Kuadrant/mcp-gateway/issues/409

- Fixes up the individual server inspect targets; adds an httproute temporarily for each, removes on ctrl+c/exit
- Small cleanup of `make help` (we sure do have a lot of commands!)

## Summary by CodeRabbit

* **New Features**
  * Standardized inspector commands for simpler, consistent usage.

* **Chores**
  * Reorganized Makefile help sections for clearer grouping.
  * Added a descriptive label for the unit test command.
  * Updated maintained word list with a new entry ("authpolicies").

### DIFF
--- a/build/inspect.mk
+++ b/build/inspect.mk
@@ -2,6 +2,29 @@
 
 open := $(shell { which xdg-open || which open; } 2>/dev/null)
 
+INSPECT_LOCAL_PORT ?= 9999
+
+# generic template for inspecting MCP servers via port-forward
+# args: $(1) = server name, $(2) = service name, $(3) = namespace, $(4) = service port, $(5) = mcp path, $(6) = extra notes
+define inspect-server-template
+	@set -eu; \
+		echo "Port-forwarding to $(1) ($(2).$(3):$(4))..."; \
+		kubectl port-forward -n $(3) svc/$(2) $(INSPECT_LOCAL_PORT):$(4) & \
+		PF_PID=$$!; \
+		trap "echo 'Cleaning up...'; kill $$PF_PID 2>/dev/null || true; exit" EXIT INT TERM; \
+		sleep 2; \
+		$(if $(6),echo "$(6)";) \
+		INSPECTOR_URL="http://localhost:6274/?transport=streamable-http&serverUrl=http://localhost:$(INSPECT_LOCAL_PORT)$(5)"; \
+		echo "MCP server: http://localhost:$(INSPECT_LOCAL_PORT)$(5)"; \
+		echo "Inspector:  $$INSPECTOR_URL"; \
+		echo ""; \
+		MCP_AUTO_OPEN_ENABLED=false DANGEROUSLY_OMIT_AUTH=true npx @modelcontextprotocol/inspector@latest & \
+		sleep 2; \
+		$(open) "$$INSPECTOR_URL" 2>/dev/null || echo "(Could not open browser automatically)"; \
+		echo "Press Ctrl+C to stop and cleanup"; \
+		wait
+endef
+
 # URLs for services
 urls-impl:
 	@echo "=== MCP Gateway URLs ==="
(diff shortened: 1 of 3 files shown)


build/inspect.mk
```diff
@@ -2,67 +2,82 @@
...
 open := $(shell { which xdg-open || which open; } 2>/dev/null)
 
+# generic template for inspecting MCP servers via temporary external httproute
+# args: $(1) = server name, $(2) = hostname slug, $(3) = service name, $(4) = service port, $(5) = mcp path, $(6) = tools description, $(7) = extra notes
+define inspect-server-template
+	@ROUTE_NAME=$(2)-inspect-route; \
+		echo "Creating temporary HTTPRoute for $(1)..."; \
+		echo '{"apiVersion":"gateway.networking.k8s.io/v1","kind":"HTTPRoute","metadata":{"name":"'$$ROUTE_NAME'","labels":{"mcp-inspect":"true"}},"spec":{"parentRefs":[{"name":"mcp-gateway","namespace":"gateway-system"}],"hostnames":["$(2).127-0-0-1.sslip.io"],"rules":[{"matches":[{"path":{"type":"PathPrefix","value":"/"}}],"backendRefs":[{"name":"$(3)","port":$(4)}]}]}}' | kubectl apply -n mcp-test -f -; \
+		echo "Waiting for HTTPRoute to be accepted..."; \
+		kubectl wait -n mcp-test httproute/$$ROUTE_NAME --for=jsonpath='{.status.parents[0].conditions[?(@.type=="Accepted")].status}'=True --timeout=30s; \
```
trepel: This fails with `error: jsonpath wait format must be --for=jsonpath='{.status.readyReplicas}'=3 or --for=jsonpath='{.status.readyReplicas}'` for me

build/inspect.mk
```diff
@@ -2,67 +2,82 @@
...
 	@echo "=== MCP Gateway URLs ==="
 	@echo "Gateway: http://mcp.127-0-0-1.sslip.io:$(KIND_HOST_PORT_MCP_GATEWAY)"
 	@echo "Keycloak: https://keycloak.127-0-0-1.sslip.io:$(KIND_HOST_PORT_KEYCLOAK)"
 
+##@ Inspection
+
 .PHONY: inspect-server1
 inspect-server1: ## Open MCP Inspector for test server 1
-	$(call inspect-server-template,test server 1,mcp-test-server1,9090,hi time slow headers)
+	$(call inspect-server-template,test server 1,server1,mcp-test-server1,9090,/mcp,hi time slow headers)
```
trepel: This tool list (`hi time slow headers`) is no longer accurate. `greet time slow headers` are the current tools.
These lists might be obsolete for other servers, it seems to me that it would be easy to forget about updating these, do we really need these lists?

build/inspect.mk
```diff
@@ -2,67 +2,82 @@
...
+		echo "Waiting for HTTPRoute to be accepted..."; \
+		kubectl wait -n mcp-test httproute/$$ROUTE_NAME --for=jsonpath='{.status.parents[0].conditions[?(@.type=="Accepted")].status}'=True --timeout=30s; \
+		trap "echo 'Cleaning up HTTPRoute...'; kubectl delete httproute -n mcp-test $$ROUTE_NAME 2>/dev/null || true; exit" INT TERM; \
+		echo "Opening MCP Inspector for $(1) at http://$(2).127-0-0-1.sslip.io:$(KIND_HOST_PORT_MCP_GATEWAY)$(5)"; \
+		echo "Available tools: $(6)"; \
+		$(if $(7),echo "$(7)";) \
+		echo ""; \
+		MCP_AUTO_OPEN_ENABLED=false DANGEROUSLY_OMIT_AUTH=true npx @modelcontextprotocol/inspector@latest & \
+		sleep 2; \
+		$(open) "http://localhost:6274/?transport=streamable-http&serverUrl=http://$(2).127-0-0-1.sslip.io:$(KIND_HOST_PORT_MCP_GATEWAY)$(5)"; \
```
trepel: When I call this for let's say test-server3 and call tool list, I can see tools for all the other test MCP servers, is that expected?

build/inspect.mk
```diff
@@ -2,67 +2,82 @@
...
-		echo "Opening MCP Inspector for custom path server at http://localhost:9094/v1/special/mcp"; \
-		echo "NOTE: This server uses a custom path /v1/special/mcp instead of /mcp"; \
-		echo ""; \
-		MCP_AUTO_OPEN_ENABLED=false DANGEROUSLY_OMIT_AUTH=true npx @modelcontextprotocol/inspector@latest & \
-		sleep 2; \
-		$(open) "http://localhost:6274/?transport=streamable-http&serverUrl=http://localhost:9094/v1/special/mcp"; \
-		echo "Press Ctrl+C to stop and cleanup"; \
-		wait; \
-		kill $$PF_PID 2>/dev/null || true
+	$(call inspect-server-template,custom path server,custom-path,mcp-custom-path-server,8080,/v1/special/mcp,echo_custom path_info timestamp,NOTE: This server uses a custom path /v1/special/mcp instead of /mcp)
```
trepel: this does not work, MCP Inspector can't connect via the inspect-route. Probably because there is no listener configured in Gateway for custom-path MCP server

trepel (changes requested):

build/inspect.mk
jasonmadigan: not expected if it's hitting just the route (point of this change generally is to be able to poke at the individual MCP servers for debugging, without the gateway). I wasn't able to reproduce this one. Wondering: I have seen MCP inspector "cache" old MCP server URLs sometimes, we sure that isn't at play?

build/inspect.mk
jasonmadigan: Agree!

build/inspect.mk
jasonmadigan: Hmm, not sure how this worked previously. Fixed now tho

build/inspect.mk
trepel: Hmm, I keep seeing all the other tools from other servers for some reason. It does not seem to be MCP Inspector related, I can see the tools if using curl too:
```
curl -s -D /tmp/mcp_headers -X POST http://server2.127-0-0-1.sslip.io:8001/mcp \
  -H "Content-Type: application/json" \
  -d '{"jsonrpc": "2.0", "id": 1, "method": "initialize", "params": {"protocolVersion": "2025-11-25", "capabilities": {}, "clientInfo": {"name": "test-client", "version": "1.0.0"}}}'

SESSION_ID=$(grep -i "mcp-session-id:" /tmp/mcp_headers | cut -d' ' -f2 | tr -d '\r')

curl -s http://server2.127-0-0-1.sslip.io:8001/mcp \
  -H "Content-Type: application/json" \
  -H "mcp-session-id: $SESSION_ID" \
  -d '{
    "jsonrpc": "2.0",
    "id": 1,
    "method": "tools/list"
  }' | jq
```
Since it works for you, any idea what I am doing wrong?

build/inspect.mk
```diff
@@ -2,67 +2,77 @@
...
 
+# generic template for inspecting MCP servers via temporary external httproute
+# args: $(1) = server name, $(2) = hostname slug, $(3) = service name, $(4) = service port, $(5) = mcp path, $(6) = extra notes
+define inspect-server-template
+	@set -eu; \
+		ROUTE_NAME=$(2)-inspect-route; \
+		echo "Creating temporary HTTPRoute for $(1)..."; \
+		echo '{"apiVersion":"gateway.networking.k8s.io/v1","kind":"HTTPRoute","metadata":{"name":"'$$ROUTE_NAME'","labels":{"mcp-inspect":"true"}},"spec":{"parentRefs":[{"name":"mcp-gateway","namespace":"gateway-system"}],"hostnames":["$(2).127-0-0-1.sslip.io"],"rules":[{"matches":[{"path":{"type":"PathPrefix","value":"/"}}],"backendRefs":[{"name":"$(3)","port":$(4)}]}]}}' | kubectl apply -n mcp-test -f -; \
+		echo "Waiting for HTTPRoute to be accepted..."; \
+		kubectl wait -n mcp-test httproute/$$ROUTE_NAME --for=condition=Accepted --timeout=30s; \
```
trepel: This does not work for me. The reason seems to be that the condition is not top level (`.status.conditions`) but is nested (`.status.parents.conditions`)

Claude suggested
```
  kubectl wait -n mcp-test httproute/$ROUTE_NAME \                                                                                                                                                 
    --for=jsonpath='{.status.parents[0].conditions[0].type}'=Accepted \
    --timeout=30s
```
But I am not sure whether we can rely on the fact that the condition will always be on 0th index so maybe smt like
```
timeout 30 bash -c 'until kubectl get httproute -n mcp-test "$ROUTE_NAME" -o jsonpath="{.status.parents[*].conditions[*].message}" | grep -q "Route was valid"; do sleep 1; done'
```
might be more stable

trepel (changes requested):

build/inspect.mk
trepel: I used `make local-env-setup` + `make deploy-test-servers` + `make deploy-example` from `main` to prepare my environment. Not sure if that makes any difference

jasonmadigan: I think I may have had a funny setup after testing over the past couple of days with different gateway than the usual one setup. With a clean setup, I see the same issue as you - it's because these routes use the same gateway.

Given the spirit of the issue is to make it easy to get mcp inspector to hit the mcp servers as though the gateway isn't there, I think some targets with port forwarding is way easier/cleaner. I'll amend this to rewrite this. Sorry for the review churn while we get an approach!

trepel (approved):
LGTM

Issue #714, opened 2026-04-02
Controller should generate and manage JWT_SESSION_SIGNING_KEY via Secret


maleck13:
## Problem

The `JWT_SESSION_SIGNING_KEY` used for signing gateway session JWTs currently defaults to a hardcoded string in `main.go` unless overridden via the flag / envvar directly.

## Proposal

The MCPGatewayExtension controller should generate a random value for `JWT_SESSION_SIGNING_KEY`, store it in a Kubernetes Secret, and mount it into the broker-router deployment as an environment variable. This follows the same pattern already used by `trustedHeadersKey` where the controller generates key material and wires it into the deployment.

### Expected behavior

1. When reconciling an MCPGatewayExtension, the controller generates a cryptographically random signing key (if the Secret does not already exist)
2. The key is stored in a Secret in the MCPGatewayExtension namespace with an owner reference
3. The Secret is mounted into the broker-router Deployment to set the `JWT_SESSION_SIGNING_KEY` env var
4. The hardcoded default in `main.go` is removed — if no key is provided, the broker should fail to start

### Notes

- HS256 is symmetric so the key is just a random byte sequence — no key pair needed
- Existing `trustedHeadersKey` pattern in the controller can be referenced for implementation
- The Secret should not be regenerated on every reconcile — only created if absent

PR #715, opened 2026-04-02
generate JWT session signing key via controller-managed secret, instead of hardcoded default

jasonmadigan:
Re: https://github.com/Kuadrant/mcp-gateway/issues/714

## Summary by CodeRabbit

* **New Features**
  * Automatic JWT session signing key management: controller ensures a secret exists, generates a secure key when needed, and associates metadata.

* **Bug Fixes**
  * Startup now requires an explicit signing key in standalone mode; clearer validation and failure behavior.
  * Deployment manifests now include the session signing key as a default environment variable.

* **Tests**
  * Added tests covering secret creation, regeneration, repair, and reconciliation behavior.

* **Chores**
  * Expanded controller permissions for secret management.

### DIFF
--- a/internal/controller/session_signing_key.go
+++ b/internal/controller/session_signing_key.go
@@ -1,0 +1,122 @@
+package controller
+
+import (
+	"context"
+	"crypto/rand"
+	"encoding/hex"
+	"fmt"
+
+	mcpv1alpha1 "github.com/Kuadrant/mcp-gateway/api/v1alpha1"
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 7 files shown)


internal/controller/session_signing_key.go
```diff
@@ -0,0 +1,87 @@
...
+	if _, err := rand.Read(keyBytes); err != nil {
+		return fmt.Errorf("failed to generate session signing key: %w", err)
+	}
+
+	secret := &corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      sessionSigningKeySecretName,
+			Namespace: mcpExt.Namespace,
+			Labels: map[string]string{
+				labelManagedBy: labelManagedByValue,
```
maleck13: I see it is here 👍

Issue #499, opened 2026-01-15
API to discover/list MCP servers behind a gateway


onmete:
# Feature Request: API to discover/list MCP servers behind a gateway

## Use Case

OpenShift Lightspeed (OLS) is an AI-powered assistant that will integrate with MCP Gateway. As an MCP client, OLS needs to:

1. **Discover available MCP servers** behind the gateway
2. **Expose this list** to administrators/users so they can select which MCPs should be available to the assistant
3. **Connect to a subset** of MCPs relevant to the OLS use case (not all MCPs behind the gateway)

## Problem

Currently, discovering MCP servers behind a gateway requires access to `MCPServerRegistration` resources and scanning/following them manually. This is not ideal for clients that need programmatic discovery.

## Proposed Solution

Provide an API or mechanism for MCP clients to:
- List all MCP servers registered with a gateway
- Retrieve metadata about each MCP server (name, available tools, description)
- Optionally filter by virtual server or other criteria

## Context

In a multi-tenant environment where multiple teams register their MCP servers with a shared gateway, clients like OLS need to:
- Know what's available
- Allow users to configure which MCPs are relevant to their use case
- Avoid receiving tools from irrelevant MCPs

This ties into the virtual MCP server concept and the `X-Mcp-Virtualserver` header mechanism.

david-martin:
Dropping some initial thoughts here.

Related to #433 

I was thinking would tool annotations help here, like an annotation with the original MCP Server name.
However, there's no built in way to the MCP protocol to search or filter based on annotation. It's just a [tools list](https://modelcontextprotocol.io/specification/2025-11-25/server/tools#listing-tools), with optional pagination/cursor. So that feels very limiting and would require client side filtering.

A built in mechanism to the protocol could help (discussed somewhat in #433).

Another approach is a new API, as suggested in the description above.

Question @onmete Is there a need for a more dynamic 'virtual' tool list than the existing MCPVirtualServer functionality?
As in, one that works via the HTTP API (like more headers or different endpoints) rather than via CRDs?
Or am I misunderstanding the ask?

maleck13:
@david-martin one thing to consider here is the status endpoints. I was already considering adding a CLI to query those they might be the solution @onmete is looking for here?

esnible:
Is there a rationale for creating a REST interface to ask the MCP Gateway for details, rather than using `kubectl get mcpserverregistrations --all-namespaces` or the equivalent K8s API query?

[...]

PR #716, opened 2026-04-02
POC tool discovery proposal (dont merge)

maleck13:
Related to #499 

This is a purely a POC **DO NOT MERGE** 

It shows what we may be able to do with categories hints and some gateway specific tools

You can clone this branch and follow the guide here to see it in action

https://github.com/maleck13/mcp-gateway/blob/c29bfe548fe6b0e8246f1f25e1755189767815ba/docs/guides/tool-discovery.md


docs/design/tool-discovery.md
```diff
@@ -0,0 +1,219 @@
...
+| **Extra turns** | 1 (browse + select) | 1 (select only) | 1 (search + select) |
+| **Server complexity** | Low — metadata indexing only | None | Medium — search index, optional embeddings |
+| **Protocol compliance** | Pure MCP | Pure MCP | Pure MCP |
+
+### Why LLM-driven over server-side search
+
+Server-side search (keyword, BM25, semantic) works well for exact matches but struggles with intent. In the restaurant booking example, a keyword search for "restaurant booking" would find restaurant tools but likely miss calendar tools. The LLM understands that booking implies checking availability — it captures the user's intent, not just their words.
+
+The progressive discovery approach lets the LLM make these connections while keeping the token cost low by presenting metadata rather than full schemas.
+
```
blublinsky, line 212: Alternative is RAG driven https://github.com/blublinsky/tools_rag

maleck13: Interesting. This proposed approach is quite simplistic and lightweight. Do you see that they could both be options?

blublinsky: What I am suggesting is by far cheaper than LLM usage

maleck13: Sure. Not questioning that. What I was alluding to was that the rag approach increases the deployment complexity and overhead. If I understand correctly, you would need to deploy additional components and manage those? The proposal is for the gateway to offer an out of the box improvement. This could then be complimented / replaced depending on the users requirements by something more involved such as the linked repo? Or do you see no value in the gateway only approach?

blublinsky: Both can be used, sure. I would only suggest to try LLM one with a large amount of tools > 500 to see if you will kill it. The other thing - relevant to both approaches is a good tools definition semantics - garbage in/garbage out. 
Are you guys planning on any kind of tools definition validation? With automatic tools addition, description collision might become a real issue

maleck13: Yes that makes sense. I will add a section to validate the approach with larger tool sets 👍 

You are correct on the definitions. I discovered this immediately when trying this locally with our test servers. At the moment we don't do any validation of descriptions etc currently.  But I will add that as a consideration to the doc.

@jasonmadigan I am going to remove the POC code from this PR and just keep it narrowed down to the proposal

maleck13:
closing in favour of just the proposals in #902

PR #733, opened 2026-04-08
fixes for the register mcp servers guide

jasonmadigan:
## Summary

Fixes issues found during an end-to-end walkthrough of the register-mcp-servers guide.

- Add note that only one MCPGatewayExtension per gateway listener is allowed, with skip-to-step-2 guidance for users who already have one
- Move ReferenceGrant creation before `kubectl wait` so users don't hit a 60s timeout
- Replace log-tailing verification with `kubectl get mcpsr -A` (users likely won't have access to broker logs)
- Update protocol version from `2025-06-18` to `2025-11-25`
- Fix heading level inconsistency

Fixes #731

## Summary by CodeRabbit

* **Documentation**
  * Restructured server registration guide into explicit numbered steps for setup, verification, and tool discovery
  * Added constraint: one extension per Gateway listener and clarified section names must match listener names
  * Clarified cross-namespace grant ordering and changed readiness checks to use explicit wait + get/describe examples
  * Simplified troubleshooting to inspect the registration resource only
  * Updated example protocol version to the latest specification

### DIFF
--- a/docs/guides/register-mcp-servers.md
+++ b/docs/guides/register-mcp-servers.md
@@ -10,18 +10,13 @@ You must register your MCP servers to be discovered and routed by the MCP Gatewa
 
 ## Procedure
 
-To connect an MCP server to MCP Gateway, you need:
-1. An MCPGatewayExtension resource that targets your Gateway
-2. A ReferenceGrant if the MCPGatewayExtension is in a different namespace than the Gateway
-3. An HTTPRoute that routes to your MCP server
-4. An MCPServerRegistration resource that references the HTTPRoute
+## Step 1: Ensure MCPGatewayExtension exists
 
-The MCPGatewayExtension tells the controller which Gateway this MCP Gateway instance serves. Without it, MCPServerRegistration resources will remain in NotReady status.
+An MCPGatewayExtension tells the controller which Gateway this MCP Gateway instance serves. Without it, MCPServerRegistration resources will remain in NotReady status.
 
-## Step 1: Create MCPGatewayExtension
+> **Note:** Only one MCPGatewayExtension is allowed per namespace. If you followed the [quick start](./quick-start.md) or [install guide](./how-to-install-and-configure.md), an MCPGatewayExtension already exists. Check with `kubectl get mcpgatewayextension -A`. If one is already present and Ready, skip to Step 2.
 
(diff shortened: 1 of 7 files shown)


trepel (dismissed):
LGTM!

If you will be force pushing this for whatever reason, you might consider fix the stale `grep "Discovered tools"` in docs/guides/troubleshooting.md even though that's separate concern.

trepel (dismissed):
LGTM

docs/guides/register-mcp-servers.md
```diff
@@ -10,18 +10,13 @@ You must register your MCP servers to be discovered and routed by the MCP Gatewa
...
-4. An MCPServerRegistration resource that references the HTTPRoute
+## Step 1: Ensure MCPGatewayExtension exists
 
-The MCPGatewayExtension tells the controller which Gateway this MCP Gateway instance serves. Without it, MCPServerRegistration resources will remain in NotReady status.
+An MCPGatewayExtension tells the controller which Gateway this MCP Gateway instance serves. Without it, MCPServerRegistration resources will remain in NotReady status.
 
-## Step 1: Create MCPGatewayExtension
-
-First, create an MCPGatewayExtension in the same namespace as your MCP Gateway broker/router deployment. It should target a unique Gateway resource. 
+> **Note:** Only one MCPGatewayExtension is allowed per namespace. If you followed the [quick start](./quick-start.md) or [install guide](./how-to-install-and-configure.md), an MCPGatewayExtension already exists. Check with `kubectl get mcpgatewayextension -A`. If one is already present and Ready, skip to Step 2.
```
trepel: `MCPGatewayExtension is allowed per namespace` - I am a bit confused here. I can see that `MCPGatewayExtension` contains `.targetRef.sectionName` which makes me think that there can be more `MCPGatewayExtensions` per namespace as long as they target different Gateway Listener?

jasonmadigan: right to be confused, re-reading this.

sectionName would selects which listener on the Gateway to read config from (port, hostname), but the controller enforces one MCPGatewayExtension per namespace regardless. Multiple can target the same Gateway from different namespaces on different ports, but not same namespace. Updated the note here and fixed the misleading comment on the type itself, which I imagine caused the wrongness

jasonmadigan: fixed

trepel (approved):
LGTM

PR #734, opened 2026-04-08
Authentication review

Patryk-Stefanski:
**Quick-start port alignment (8001)**
  - Changed Kind hostPort from 7001 to 8001 in scripts/quick-start.sh, docs/guides/quick-start.md, and
  charts/sample_local_helm_setup.sh
  - Aligns with config/kind/cluster.yaml and all other guides (authentication, authorization, virtual servers, etc.)

  **Authentication guide fixes**
  - Added GATEWAY_IP fallback to pod IP for Kind clusters where gateway has no LoadBalancer address
  - Fixed scopes_supported example response to include all 4 scopes (basic, groups, roles, profile) matching the env var set in Step
   2
  - Added note to Step 5 about port-forwarding Keycloak (port 8002) when using the quick-start setup

## Summary by CodeRabbit

* **Documentation**
  * Updated Quick Start and authentication guides to use new local ports (8001/8002)
  * Added Keycloak port-forward note for local testing
  * Expanded OAuth metadata examples with extra supported scopes
  * Improved Gateway IP resolution in setup instructions

* **Chores**
  * Updated local setup scripts and sample setup to reflect revised port mappings (8001/8002)

### DIFF
--- a/charts/sample_local_helm_setup.sh
+++ b/charts/sample_local_helm_setup.sh
@@ -34,10 +34,10 @@ nodes:
         node-labels: "ingress-ready=true"
   extraPortMappings:
   - containerPort: 30080
-    hostPort: 7001
+    hostPort: 8001
     protocol: TCP
   - containerPort: 30089
-    hostPort: 7002
+    hostPort: 8002
     protocol: TCP
 EOF
 fi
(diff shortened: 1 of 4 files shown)


trepel (approved):
LGTM!

Do you know whether it is ok for the Gateway not being programmed? It seems to be working fine but it is a bit confusing.

trepel:
> > LGTM!
> > Do you know whether it is ok for the Gateway not being programmed? It seems to be working fine but it is a bit confusing.
> 
> I Think were okay. It `Programmed` status is false because the Service has no external IP assigned.
> 
> ```
>   address pending for hostname "mcp-gateway-istio.gateway-system.svc.cluster.local"                                                                                                                      
> ```
> 
> The Istio-created Service (mcp-gateway-istio) is type LoadBalancer, but in a Kind cluster there's no cloud load balancer provider, so status.loadBalancer stays empty {}. Istio's gateway controller sees the address is still pending and sets Programmed: False with reason AddressNotAssigned.

Ok, I got it, thanks! It is `Programmed` if using `make local-env-setup` because it uses metallb as a load balancer.

PR #736, opened 2026-04-08
Fix broken verification and resource names in external MCP server guide

jasonmadigan:
Resolves https://github.com/Kuadrant/mcp-gateway/issues/735

Fixes found by walking through the guide end-to-end on a local kind cluster (`make local-env-setup`) against `release-0.6.0`.

## Changes

- **Step 7**: Replace log-grep wait loop (`grep "Discovered.*tools.*github"`) with `kubectl get mcpsr -n mcp-test` -- the grep pattern never matches actual broker logs, causing the loop to hang forever. Users also likely won't have access to broker logs.
- **Verification**: Merged into Step 7 (was redundant and also used the broken grep)
- **Step 5 text**: `MCPServer` -> `MCPServerRegistration`
- **Cleanup**: `kubectl delete mcpserver` -> `kubectl delete mcpserverregistration`
- **Test Integration**: Removed standalone `curl tools/list` that returns `Invalid session ID` without prior session initialisation

## Summary by CodeRabbit

* **Documentation**
  * Updated external MCP server setup guide with streamlined verification process.
  * Replaced log-polling configuration checks with direct readiness verification commands.
  * Enhanced testing approach using interactive inspector tool instead of direct API calls.
  * Updated expected output formats and diagnostic indicators.

### DIFF
--- a/docs/guides/external-mcp-server.md
+++ b/docs/guides/external-mcp-server.md
@@ -159,7 +159,7 @@ The `mcp.kuadrant.io/secret=true` label is required. Without it the MCPServerReg
 
 ## Step 5: Create the MCPServerRegistration Resource
 
-Create the `MCPServer` resource that registers the GitHub MCP server with the gateway:
+Create the `MCPServerRegistration` resource that registers the GitHub MCP server with the gateway:
 
 ```bash
 kubectl apply -f - <<EOF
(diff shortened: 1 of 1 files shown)


trepel (approved):
LGTM!

PR #770, opened 2026-04-15
rc2 isolated gateway deployment guide review

Patryk-Stefanski:
-  Remove stale Helm flags (envoyFilter.create, envoyFilter.name, broker.create) from all Helm commands in the isolated gateway deployment guide. These values were removed from the chart when EnvoyFilter and broker-router creation moved to the controller (8e3279a). 
-  Fix claim that the controller creates "ServiceAccount and RBAC" — it only creates a ServiceAccount, no Role/RoleBinding resources.


trepel (approved):
LGTM

PR #774, opened 2026-04-16
Fix CI image mismatch and remove stale docs

jasonmadigan:
## Summary
- Fix CI image mismatch on version bump PRs: build-image now also tags as latest, CI kustomize overlay references latest, conformance workflow no longer re-deploys via mcp-system overlay
- Remove stale x-mcp-api-key header references (never implemented, documented in error)
- Fix dead file path in AGENTS.md (pkg/controller -> internal/controller)


jasonmadigan:
CC: @maleck13 removed those dead header refs.

maleck13 (approved):

PR #780, opened 2026-04-17
add additional handler for client requesting /.well-known/oauth-prote…

### LINKED ISSUE #776: Controller-managed HTTPRoute missing /.well-known/oauth-protected-resource path
## Bug

When `httpRouteManagement` is `Enabled` (the default) on an MCPGatewayExtension, the controller's `buildGatewayHTTPRoute()` only creates a single rule for `/mcp`. The `/.well-known/oauth-protected-resource` rule is missing.

The kustomize overlay used by `make local-env-setup` works correctly because it defines the HTTPRoute statically with both rules — but users relying on auto-managed HTTPRoutes lose the well-known OAuth discovery endpoint.

### No workaround with httpRouteManagement enabled

The controller's reconcile loop overwrites `spec.rules` on every update (`broker_router.go:388`), so manually adding the well-known rule to the controller-managed HTTPRoute will be reverted on the next reconcile.

The only workaround is `httpRouteManagement: Disabled` with a manually managed HTTPRoute.

## Fix

1. **Overlay**: Add `httpRouteManagement: Disabled` to the MCPGatewayExtension in `config/mcp-gateway/` so the static HTTPRoute with both rules isn't overwritten
2. **Controller**: Update `buildGatewayHTTPRoute()` to include a second rule for `/.well-known/oauth-protected-resource`

## Affects

0.6

maleck13:
…cted-resource/mcp and not falling back to /.well-known/oauth-protected-resource

Allows for clients requesting `/.well-known/oauth-protected-resource/mcp` and those that might fallback to /.well-known/oauth-protected-resource

related to investigation around #776 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved URL handling for the OAuth protected resource endpoint to accept requests with and without trailing slashes.

### DIFF
--- a/cmd/mcp-broker-router/main.go
+++ b/cmd/mcp-broker-router/main.go
@@ -304,6 +304,7 @@ func setUpHTTPServer(address string, mcpBroker broker.MCPBroker, sessionManager 
 
 	oauthHandler := broker.ProtectedResourceHandler{Logger: logger}
 	mux.HandleFunc("/.well-known/oauth-protected-resource", oauthHandler.Handle)
+	mux.HandleFunc("/.well-known/oauth-protected-resource/", oauthHandler.Handle)
 
 	// WriteTimeout of 0 (disabled) is important for SSE connections (GET /mcp).
 	// SSE streams notifications indefinitely - any write timeout would kill the connection.


jasonmadigan:
Change looks good, the gevals error looks like some dep changed:

`Error: unknown command "verify" for "mcpchecker"`

david-martin:
> gevals error looks like some dep changed

Fixed in #784

jasonmadigan (approved):

PR #782, opened 2026-04-17
Claude oauth access

maleck13:
@guicassolato would appreciate a quick look

This allows CLI based tools like claude code to use the example OAuth guide

## Release Notes

* **New Features**
  * Added `offline_access` OAuth scope support for enhanced authentication flexibility.
  * Implemented per-user authentication with GitHub token replacement—requests now use the `x-github-pat` header for secure credential handling.

* **Documentation**
  * Updated authentication guides to reflect new OAuth scope support and GitHub token authentication workflow.
  * Restructured external MCP server setup guide with clearer basic and advanced authentication sections.

### DIFF
--- a/config/keycloak/realm-import.yaml
+++ b/config/keycloak/realm-import.yaml
@@ -642,11 +642,11 @@ data:
         "basic",
         "email",
         "profile",
-        "roles",
-        "offline_access"
+        "roles"
       ],
       "defaultOptionalClientScopes": [
-        "groups"
+        "groups",
+        "offline_access"
       ],
       "components": {
         "org.keycloak.services.clientregistration.policy.ClientRegistrationPolicy": [
--- a/config/keycloak/realm-import.yaml
+++ b/config/keycloak/realm-import.yaml
@@ -257,6 +257,10 @@ data:
               "temporary": false
             }
           ],
+          "realmRoles": [
+            "default-roles-mcp",
+            "offline_access"
+          ],
           "groups": [
             "mcp-users",
             "accounting",
@@ -638,7 +642,8 @@ data:
         "basic",
         "email",
         "profile",
-        "roles"
+        "roles",
+        "offline_access"
       ],
       "defaultOptionalClientScopes": [
         "groups"
--- a/docs/guides/external-mcp-server.md
+++ b/docs/guides/external-mcp-server.md
@@ -1,65 +1,49 @@
 # Connecting to External MCP Servers
 
-This guide demonstrates how to connect MCP Gateway to external MCP servers using Gateway API and Istio. We'll use the public GitHub MCP server as an example.
+This guide walks through connecting MCP Gateway to an external MCP server. We use the GitHub MCP server as an example, but the pattern applies to any external service.
 
-Clients call your Gateway's hostname, and the Gateway rewrites and routes traffic to the external service.
+The guide has two parts:
+
+1. **Basic setup** — connect to the external server so the broker can discover and list its tools
+2. **Per-user authentication** — add OIDC and let each user supply their own upstream credential for tool calls via a custom header
 
 ## Prerequisites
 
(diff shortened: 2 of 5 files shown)


config/keycloak/realm-import.yaml
```diff
@@ -257,6 +257,10 @@ data:
               "temporary": false
             }
           ],
+          "realmRoles": [
+            "default-roles-mcp",
```
david-martin, line 261: what is 'default-roles-mcp'?

maleck13:
This work compliments work I am doing to add an example of using a user defined token to connect to the GH mcp server via the gateway. I am using claude as the client to prove it out

config/keycloak/realm-import.yaml
maleck13: it is a set of default roles for the mcp realm

image

docs/guides/external-mcp-server.md
```diff
@@ -1,65 +1,49 @@
...
-- Validate your GITHUB_PAT environment variable and token format
-- Create ServiceEntry, DestinationRule, HTTPRoute, Secret, and MCPServerRegistration
-- Apply the AuthPolicy for OAuth + API key handling
+`auth-example-setup` creates its own AuthPolicies (`mcp-auth-policy` and `mcps-auth-policy` in `gateway-system`). If you ran that command, remove the existing AuthPolicies before following this guide:
 
-All the sample YAML files are available in `config/samples/remote-github/` for reference or customization. For a detailed explanation of each component, continue reading the manual setup steps below.
+```bash
+kubectl delete authpolicy mcp-auth-policy -n gateway-system --ignore-not-found
+kubectl delete authpolicy mcps-auth-policy -n gateway-system --ignore-not-found
+```
```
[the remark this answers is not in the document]
david-martin: +1 on not including any local dev stuff in this guide.
We could still mention if you just want to try out something locally, point to the repo and mention the targets, but let's not actually instruct people to do that.
A guide is meant to be more generally applicable to any environment (maybe you were thinking of a tutorial that's hand holding start to finish).

So, no need to go overboard. We shouldn't give instructions on how to install keycloak, kuadrant and an external MCP server. They can be pre-reqs, or link to relevant install docs.

david-martin (commented):
Checking, is the doc meant to be a how-to guide or a tutorial? (https://diataxis.fr/)
We can make things a lot easier on ourselves if we stick to one or the other.
To date we haven't really had any tutorials, but this doc seems to bounce between the 2.

maleck13:
I guess it is a tutorial I have not seen that distinction before

david-martin:
It's used in https://docs.kuadrant.io, with 4 top level docs types aligning to that framework (Concepts, Reference, Tutorials & Guides)

All tutorials start at 0 and 'handhold' at every step so everyone gets to the exact same point. They're not intended to solve production use cases. Moreso for learning. (like a lesson)
All guides show how to do a specific thing regardless of the environment. They can assume prior knowledge and expect the user to apply what's in the guide to their own unique setup.

There's a lot in this doc, enough that it feels like theres content for both (e.g. a guide for how connect your MCP gateway to an external MCP server with auth (using the github MCP as an inline example), and a tutorial that's github MCP server specific where everyone that follows has that example working at the end).
However, we can merge and have a follow up to split/rework if you prefer?

maleck13:
Sure lets go with a follow on issue describing what we want from each. Should be simple to change at that point

david-martin (approved):
Approving. CI is green, flow looks good.

I'll open a follow-up PR to fix the issuerUrl scheme inconsistency — `auth.mk` and `authentication.md` use `http://` while this PR's files use `https://`. They need to match for JWT validation. Separate files so easier to do as a standalone PR.

Will also create an issue for restructuring this doc into a guide + tutorial per our Diataxis conversation.

PR #793, opened 2026-04-21
fix(oidc-server): handle JWT aud claim as string or array

### LINKED ISSUE #792: [Bug] OIDC server fails to parse JWT tokens with array-formatted aud claim
## Description

The OIDC test server fails to authenticate requests when the JWT token contains an `aud` (audience) claim formatted as an array instead of a string. This causes authentication failures with the error:

```
Failed to parse claims: json: cannot unmarshal array into Go struct field .aud of type string
```

## Root Cause

The JWT specification (RFC 7519) allows the `aud` claim to be either:
- A single string: `"aud": "client-id"`
- An array of strings: `"aud": ["client-id", "other-audience"]`

The OIDC server was only handling the string format, causing it to reject valid tokens with array-formatted audiences.

## When This Occurs

This issue appears in certain Keycloak configurations and OAuth flows:
- When using the `oidc-audience-resolve-mapper` protocol mapper
- During token exchange flows
- When a user has roles in multiple clients
- Differences between Keycloak versions (older versions may return string, newer versions may return array)

## Impact

Users unable to call tools on the OIDC server despite having valid OAuth tokens, particularly when using:
- Token exchange flows
- Multi-audience token scenarios

## Solution

The fix updates the claims parsing logic to handle both formats using a type switch:
- Change `Aud` field from `string` to `interface{}`
- Add validation logic for both string and array types
- Ensure audience validation works for both formats

## Additional Context

- Affects: `tests/servers/oidc-server/main.go`
- Test servers may need rebuilding after merge
- The fix should maintain backward compatibility with string-formatted aud claims

trepel:
### Summary

resolves https://github.com/Kuadrant/mcp-gateway/issues/792

This fix updates the OIDC test server to accept both formats by:
  - Changing Aud field type from string to interface{}
  - Adding type switch to validate against both string and []interface{}
  - Preventing authentication failures for valid tokens with array audiences

  Fixes authentication errors like `cannot unmarshal array into Go struct
  field .aud of type string` when using tokens from certain OIDC flows.

### Verification Steps

  1. Build and deploy the updated OIDC server:
  ```
cd tests/servers/oidc-server && docker build -t test-oidc-server:latest .
kubectl set image deployment/mcp-oidc-server -n mcp-test oidc-server=test-oidc-server:latest
```
  2. Prepare the env
```
make local-env-setup
make auth-example-setup
```
  3. Call the `hello_world` tool via MCP Inspector - should succeed
 
Note: if using KIND 'aud' seems to be string. On OpenShift it is array.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved OpenID Connect token validation to handle audience claims in both string and array formats, enabling broader compatibility with various OIDC providers that represent audience values differently.

### DIFF
--- a/tests/servers/oidc-server/main.go
+++ b/tests/servers/oidc-server/main.go
@@ -87,8 +87,8 @@ func (a *authMiddleware) ServeHTTP(w http.ResponseWriter, req *http.Request) {
 		}
 
 		var claims struct {
-			Aud string `json:"aud"`
-			Sub string `json:"sub"`
+			Aud interface{} `json:"aud"` // can be string or []string
+			Sub string      `json:"sub"`
 		}
 		if err := accessToken.Claims(&claims); err != nil {
 			log.Printf("Failed to parse claims: %v", err)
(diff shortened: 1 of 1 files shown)


maleck13 (approved):
This look ok @trepel Yes the aud can be a string or array so good to catch this. https://github.com/golang-jwt/jwt/blob/main/types.go#L102

Issue #629, opened 2026-03-03
Tech Debt - Improve how checking the status of the MCPServers with the gateway is handled


maleck13:
**What**

Currently the controller sends a request to the broker to get the current status of that MCP Server and report it in the status of the MCPServerRegistration. 

There is legacy code here from the POC that looks up the endpoint slice. Really it should be possible to just use the service name 

**Option**

We could also remove the status checks from the controller. THe controller can consider the MCPServerRegistration complete once it has written the config for that MCP into the correct locations.

rishi-jat:
Hey @maleck13 , I would like to work on this can you please assign it to me ? thanks

malladinagarjuna2:
hey @maleck13  i have studied this issue, i know this will be helpful in solving 255, as you have self assigned it , but i want  to once try this issue

malladinagarjuna2:
@maleck13 @david-martin i have a question should the  condition type for a disabled server bem,accepted= false, reason= Disabled,  I mean "we accepted the config but it's intentionally off" or keep a separate ready= false, reason= disabled condition alongside the new accepted condition, the rest of the tests use VerifyMCPServerRegistrationNotReadyWithReason(..., "server is disabled") which checks the ready condition type. If we rename everything to accepted, that verifier needs updating too.?

PR #796, opened 2026-04-21
prompt federation design doc

Patryk-Stefanski:
## Summary by CodeRabbit

* **Documentation**
  * Added comprehensive design docs for federating prompts through the gateway, covering routing/aggregation rules, prompt scoping, discovery/prefixing and conflict detection, metadata filtering/stripping before responses, and proposed API field/header updates.
  * Details component/controller extension points and a testing strategy across unit, integration, and end-to-end scenarios.

### DIFF
--- a/docs/design/prompts-federation.md
+++ b/docs/design/prompts-federation.md
@@ -1,0 +1,190 @@
+# Feature: MCP Prompts Federation
+
+## Summary
+
+Add support for federating MCP Prompts through the gateway, following the same pattern used for tools. The broker discovers prompts from upstream MCP servers, applies prefixing to avoid collisions, and exposes them to clients. The router handles `prompts/get` requests by stripping prefixes and routing to the correct upstream server. Ref: [#787](https://github.com/Kuadrant/mcp-gateway/issues/787), split from [#208](https://github.com/Kuadrant/mcp-gateway/issues/208).
+
+## Goals
+
+- Federate prompts from multiple upstream MCP servers through a single gateway endpoint
+- Rename `toolPrefix` to `prefix` on MCPServerRegistration CRD (breaking change) and use it for both tool and prompt prefixing
+- Support `prompts/list` and `prompts/get` MCP methods
+- Handle `notifications/prompts/list_changed` from upstream servers
+- Apply VirtualServer and authorization filtering to prompts
+- Report discovered prompt counts in MCPServerRegistration status
+
+## Non-Goals
+
+- Resource federation — tracked separately in [#788](https://github.com/Kuadrant/mcp-gateway/issues/788)
+- Prompt-specific authorization claims in JWT — initial implementation applies server-level filtering only;
+- Prompt validation (prompts have no JSON schemas like tools, so `invalidToolPolicy` does not apply)
+
+## Design
+
+### Backwards Compatibility
+
+**Breaking change**: The `toolPrefix` field on MCPServerRegistration is renamed to `prefix`. This field has always been a server-level namespace, not tool-specific, and the rename aligns the API with its actual semantics now that it applies to both tools and prompts.
+
+**Migration**: Users must replace `toolPrefix` with `prefix` in their MCPServerRegistration manifests. The CEL immutability validation carries over to the renamed field.
+
+**Scope of rename** (57 files affected):
+- CRD types: `api/v1alpha1/types.go` — rename field and JSON tag
+- Config types: `internal/config/types.go` — rename `ToolPrefix` to `Prefix`
+- Manager/broker/router: update all references to `GetPrefix()`, `ToolPrefix`, etc.
+- CRD manifests, Helm charts, samples, docs, tests
+- Run `make generate-all` to regenerate CRDs and sync Helm
+
+All other changes (prompt federation, new CRD fields) are additive and non-breaking.
+
+### Architecture Changes
+
+No new components. The existing broker, manager, and router are extended.
+
+```
+prompts/list flow:
+
+  Client ──► Envoy ──► ext_proc (router) ──► HandleNoneToolCall()
+                                                    │
+                                              sets headers:
+                                              mcp-server-name=mcpBroker
+                                                    │
+                                              Envoy routes to broker
+                                                    │
+                                              Broker's mcp-go server
+                                              handles prompts/list
+                                                    │
+                                              AddAfterListPrompts hook
+                                              applies filtering
+                                                    │
+                                              returns federated prompts
+                                              to client
+
+
+prompts/get flow:
+
+  Client ──► Envoy ──► ext_proc (router) ──► HandlePromptGet()
+                                                    │
+                                              1. Extract prompt name
+                                              2. GetServerInfoByPrompt()
+                                              3. Strip prefix
+                                              4. Set routing headers
+                                              5. Init backend session
+                                                    │
+                                              Envoy routes to upstream
+                                              MCP server
+                                                    │
+                                              returns prompt messages
+                                              to client
+```
+
+`prompts/list` follows the same path as `tools/list` — it passes through the router to the broker's listening MCP server, which aggregates prompts from all managers and applies filtering via hooks.
+
+`prompts/get` follows the same path as `tools/call` — the router identifies the upstream server by the prefixed prompt name, strips the prefix, sets routing headers, and forwards to the correct upstream.
+
+### API Changes
+
+#### MCPServerRegistration Status
+
+Add `discoveredPrompts` to report prompt discovery alongside tools:
+
+```yaml
+status:
+  discoveredTools: 4
+  discoveredPrompts: 2    # new
+  conditions:
+    - type: Ready
+      status: "True"
+```
+
+#### MCPVirtualServer Spec
+
+Add optional `prompts` field:
+
+```yaml
+apiVersion: mcp.kuadrant.io/v1alpha1
+kind: MCPVirtualServer
+metadata:
+  name: my-virtual-server
+spec:
+  description: "Scoped MCP server"
+  tools:
+    - weather_forecast
+    - weather_alerts
+  prompts:                    # new, optional
+    - weather_report
+    - weather_summary
+```
+
+When `prompts` is omitted, all prompts are exposed (same behavior as tools today).
+
+### Component Changes
+
+The implementation follows the same pattern as tools throughout. Each component that handles tools gets a parallel set of prompt logic.
+
+#### Upstream Client (`internal/broker/upstream/mcp.go`)
+
+Add `ListPrompts()` and `SupportsPromptsListChanged()` to the `MCP` interface. These wrap the mcp-go client methods and check the initialize response capabilities.
+
+#### MCPManager (`internal/broker/upstream/manager.go`)
+
+Add a `PromptsAdderDeleter` interface mirroring `ToolsAdderDeleter`. The mcp-go `server.MCPServer` already implements `AddPrompts()`/`DeletePrompts()`, so the broker's listening server satisfies this interface.
+
+> **Note**: mcp-go does **not** expose a public `ListPrompts()` on the server (unlike `ListTools()`). The manager maintains its own prompt maps for lookups.
+
+**Cross-server conflict detection**: For tools, `findToolConflicts()` calls `gatewayServer.ListTools()` to see all tools from all managers and detect name collisions. Since there is no equivalent `ListPrompts()`, the broker instead aggregates prompt maps from all managers and passes them into the conflict checker. This preserves the same safety guarantee — two servers registering the same prefixed prompt name will be detected and rejected, same as tools.
+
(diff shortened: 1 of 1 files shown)


docs/design/prompts-federation.md
```diff
@@ -0,0 +1,190 @@
...
+
+- Resource federation — tracked separately in [#788](https://github.com/Kuadrant/mcp-gateway/issues/788)
+- Prompt-specific authorization claims in JWT — initial implementation applies server-level filtering only;
+- Prompt validation (prompts have no JSON schemas like tools, so `invalidToolPolicy` does not apply)
+
+## Design
+
+### Backwards Compatibility
+
+**Breaking change**: The `toolPrefix` field on MCPServerRegistration is renamed to `prefix`. This field has always been a server-level namespace, not tool-specific, and the rename aligns the API with its actual semantics now that it applies to both tools and prompts.
```
Patryk-Stefanski, line 26: In two minds about this, not sure its necessary we could either:
1. Duplicate this for prefixes for prompts and likely for resources federation as well
2. Leave it as is and just note that toolPrefix is just a prefix at the end of the day

docs/design/prompts-federation.md
```diff
@@ -0,0 +1,190 @@
...
+
+Enable `server.WithPromptCapabilities(true)` on the listening MCP server. Register an `AddAfterListPrompts` hook that calls `FilterPrompts()`. Add `GetServerInfoByPrompt()` to the `MCPBroker` interface — same pattern as `GetServerInfo()` for tools, searching managers by prefixed prompt name.
+
+The manager constructor receives the listening server as both `ToolsAdderDeleter` and `PromptsAdderDeleter`.
+
+#### Prompt Filtering (`internal/broker/filtered_prompts_handler.go`)
+
+New file mirroring `filtered_tools_handler.go`. Applies VirtualServer filtering and strips `kuadrant/id` gateway metadata from prompts before returning to clients.
+
+There is no `allowed-prompts` JWT claim in this implementation. If a user has access to any tools on a server via the `allowed-tools` JWT, they implicitly have access to all prompts on that server too. Per-prompt JWT filtering can be added later without breaking changes.
```
Patryk-Stefanski: Should we have JWT claims for prompts like we have for tools or is the approach in the doc okay for the initial implementation?

docs/design/prompts-federation.md
```diff
@@ -0,0 +1,190 @@
...
+
+Add `ListPrompts()` and `SupportsPromptsListChanged()` to the `MCP` interface. These wrap the mcp-go client methods and check the initialize response capabilities.
+
+#### MCPManager (`internal/broker/upstream/manager.go`)
+
+Add a `PromptsAdderDeleter` interface mirroring `ToolsAdderDeleter`. The mcp-go `server.MCPServer` already implements `AddPrompts()`/`DeletePrompts()`, so the broker's listening server satisfies this interface.
+
+> **Note**: mcp-go does **not** expose a public `ListPrompts()` on the server (unlike `ListTools()`). The manager maintains its own prompt maps for lookups.
+
+**Cross-server conflict detection**: For tools, `findToolConflicts()` calls `gatewayServer.ListTools()` to see all tools from all managers and detect name collisions. Since there is no equivalent `ListPrompts()`, the broker instead aggregates prompt maps from all managers and passes them into the conflict checker. This preserves the same safety guarantee — two servers registering the same prefixed prompt name will be detected and rejected, same as tools.
```
Patryk-Stefanski, line 123: I'm not comfortable relying on the broker to aggregate the prompts, but it's the best I could think of at the moment. Any other ideas are welcome

docs/design/prompts-federation.md
```diff
@@ -0,0 +1,190 @@
...
+Add support for federating MCP Prompts through the gateway, following the same pattern used for tools. The broker discovers prompts from upstream MCP servers, applies prefixing to avoid collisions, and exposes them to clients. The router handles `prompts/get` requests by stripping prefixes and routing to the correct upstream server. Ref: [#787](https://github.com/Kuadrant/mcp-gateway/issues/787), split from [#208](https://github.com/Kuadrant/mcp-gateway/issues/208).
+
+## Goals
+
+- Federate prompts from multiple upstream MCP servers through a single gateway endpoint
+- Rename `toolPrefix` to `prefix` on MCPServerRegistration CRD (breaking change) and use it for both tool and prompt prefixing
+- Support `prompts/list` and `prompts/get` MCP methods
+- Handle `notifications/prompts/list_changed` from upstream servers
+- Apply VirtualServer and authorization filtering to prompts
+- Report discovered prompt counts in MCPServerRegistration status
```
maleck13: I think we can leave this last one out as we are looking to remove the status checking from the controller in the next release #629

docs/design/prompts-federation.md
```diff
@@ -0,0 +1,190 @@
...
+
+## Non-Goals
+
+- Resource federation — tracked separately in [#788](https://github.com/Kuadrant/mcp-gateway/issues/788)
```
maleck13, line 19: are the two of these things complimentary? I wonder are the patterns going to be repeated?

docs/design/prompts-federation.md
```diff
@@ -0,0 +1,190 @@
...
+- Prompt-specific authorization claims in JWT — initial implementation applies server-level filtering only;
```
maleck13: is this referring to the x-authorized-tools header?

maleck13: may be worth thikning about generalizing that also in a new proposal `x-mcp-authorized` or something that has map[tools]map[server][]tools and map[resources].. map[prompts]....

docs/design/prompts-federation.md
maleck13: I think changing it makes sense We are not v1 yet so it is a good time to make this type of change

docs/design/prompts-federation.md
```diff
@@ -0,0 +1,190 @@
...
+
+`prompts/list` follows the same path as `tools/list` — it passes through the router to the broker's listening MCP server, which aggregates prompts from all managers and applies filtering via hooks.
+
+`prompts/get` follows the same path as `tools/call` — the router identifies the upstream server by the prefixed prompt name, strips the prefix, sets routing headers, and forwards to the correct upstream.
+
+### API Changes
+
+#### MCPServerRegistration Status
+
+Add `discoveredPrompts` to report prompt discovery alongside tools:
```
maleck13: Yeah lets not go down this route. It is a misleading status and can fall out of sync with the truth in the gateway.

docs/design/prompts-federation.md
maleck13: Yeah. So the gatewayserver is the MCP server that returns tools etc to the client from the broker. If it has no ListPrompts we might be able to wrap that server and extend it with a list prompts. It would still be the broker keeping track though. I am wondering how the mcp library handles listing prompts for the client

docs/design/prompts-federation.md
maleck13: I think we can layer this on afterwards. But I do think prompts and tools are separate things. But we maybe able to improve the x-authorized-tools header or add a new one for prompts. Lets call it out as a security consideration.

docs/design/prompts-federation.md
```diff
@@ -0,0 +1,190 @@
...
+`prompts/get` gets a new `HandlePromptGet()` handler following the same pattern as `HandleToolCall()`: extract prompt name, look up upstream server by prefix, strip prefix, manage backend session, set routing headers, forward via Envoy. A `PromptName()` method is added to `MCPRequest` mirroring `ToolName()`.
+
+#### Config and CRD Types
+
+- `internal/config/types.go`: Rename `ToolPrefix` to `Prefix` on `MCPServer`. Add `Prompts []string` to `VirtualServer`.
+- `api/v1alpha1/types.go`: Rename `toolPrefix` to `prefix` on MCPServerRegistration spec. Add `discoveredPrompts` to status. Add `prompts` to MCPVirtualServer spec.
+
+#### Controller (`internal/controller/mcpserverregistration_controller.go`)
+
+Extend `ServerValidationStatus` with `TotalPrompts`. Update `updateStatus()` to populate `DiscoveredPrompts` from the broker status response.
```
maleck13: can remove

docs/design/prompts-federation.md
```diff
@@ -0,0 +1,190 @@
...
+
+#### Notifications
+
+The existing notification design (`docs/design/notifications.md`) already accounts for `notifications/prompts/list_changed` as a state change event. The MCPManager re-fetches prompts on notification receipt, same as tools. The notifications doc status table should be updated from "Not applicable" to "Implemented" once complete.
+
+### Security Considerations
+
+- Prompt filtering reuses the existing VirtualServer mechanism. Prompts not listed in a VirtualServer's `prompts` field are not exposed.
+- The `kuadrant/id` metadata added to prompts during federation is stripped before returning to clients, same as tools.
+- Backend credentials for `prompts/get` routing use the same `credentialRef` and `x-mcp-api-key` header mechanism as `tools/call`.
```
maleck13:
```
- Backend credentials for `prompts/get` routing use the same `credentialRef` and `x-mcp-api-key` header mechanism as `tools/call`.
```

This is a bit of a hallucination I think. CredentialRef is only used for the broker to list the tools / prompts. x-mcp-api-key isn't used it just appeared in an AuthPolicy. 

The Auth mechanism is the same for tools calls though. User credential is provided / accessed in some way via an AuthPolicy and sent to the gateway. The Gateway does the initialize and prompt get using the Authorization header sent by the client

maleck13 (commented):
I think the core of this is correct. Few comments and one possible hallucination. But looking good

docs/design/prompts-federation.md
Patryk-Stefanski: I think they're complementary to a point.

Pattern repeat for resources
  -  prefix field on MCPServerRegistration applies to resource names/URIs, same as tools and prompts
  - VirtualServer filtering — a resources field alongside tools and prompts
  - Authorization header generalization (x-mcp-authorized) covers resources too
  - Manager pattern: discovery, diffing, conflict detection, add/delete on the gateway server
  - Router pattern: extract identifier, look up upstream server, strip prefix, set routing headers
  - Notification handling: notifications/resources/list_changed follows the same callback pattern
  
  
Additional considerations
- URI-based addressing: Resources are identified by URIs (e.g., file:///project/src/main.rs), not simple names. Prefixing a URI is a bit more complicated and needs a bit of thought
- Handling resource templates
- Resources can return binary data. How do we deal with it
- Handling pagination when aggregating a resource/list call
- There are also resource subscriptions (need to read up on ) that need to be considered.

docs/design/prompts-federation.md
Patryk-Stefanski: Turns out mcp-go have a list function for [prompts](https://github.com/mark3labs/mcp-go/blob/main/server/server.go#L1371) and [resources](https://github.com/mark3labs/mcp-go/blob/main/server/server.go#L1111).

The private handleListPrompts and handleListResources already exist and handle client requests. All that's missing is a public Go method on MCPServer, I'll make a PR to add them in so we can use it here.

docs/design/prompts-federation.md
```diff
@@ -0,0 +1,286 @@
...
+- Resource federation — tracked separately in [#788](https://github.com/Kuadrant/mcp-gateway/issues/788)
+- Prompt validation (prompts have no JSON schemas like tools, so `invalidToolPolicy` does not apply)
+
+## Design
+
+### Backwards Compatibility
+
+**Breaking change**: The `toolPrefix` field on MCPServerRegistration is renamed to `prefix`. This field has always been a server-level namespace, not tool-specific, and the rename aligns the API with its actual semantics now that it applies to both tools and prompts.
+
+**Migration**: Users must replace `toolPrefix` with `prefix` in their MCPServerRegistration manifests. Since the field has CEL immutability validation, existing resources must be deleted and recreated (not patched in-place). For bulk updates, a `sed` one-liner or `yq` edit on manifest files before `kubectl apply` is sufficient.
```
maleck13, line 28: This will need to be highlighted in release docs for 0.7

docs/design/prompts-federation.md
```diff
@@ -0,0 +1,286 @@
...
+  "resources": {
+    "github": ["repo://org/repo"]
+  }
+}
+```
+
+**Go type change**: The JWT parsing function signature stays the same per-capability — each filter handler (`filterToolsByServerMap`, `filterPromptsByServerMap`) still receives `map[string][]string`. The change is in the parsing layer, which unmarshals `map[string]map[string][]string` and hands each capability key to the relevant filter.
+
+**Enforcement semantics**: A missing capability key (e.g. no `"prompts"` key in the JWT) means the JWT makes no assertion about that capability — behavior depends on the enforcement flag, same as today. An empty map (`"prompts": {}`) explicitly denies all prompts.
+
```
maleck13, line 204: yeah this should also go in release notes

maleck13 (approved):

PR #802, opened 2026-04-24
move to streamed mode for the ext_proc body

maleck13:
Switches the Envoy ext_proc request body mode from BUFFERED to STREAMED. Instead of Envoy buffering the entire
  request body (constrained by Envoy's internal buffer limits), the ext_proc now receives and reassembles body chunks
  itself. Small requests — which covers most MCP calls — arrive as a single chunk, so there's no impact on existing
  performance. For larger payloads, this removes the dependency on Envoy's buffer size and gives the gateway direct
  control. Adds a configurable max request body size (--max-request-body-size, default 5MB) — requests exceeding this
  limit are rejected with 413.

## Summary by CodeRabbit

* **New Features**
  * Added configurable CLI parameter to set maximum request body size limit
  * Migrated from buffered to streamed request body processing for improved handling of large payloads
  * Implemented request body size validation with 413 error response for oversized requests

### DIFF
--- a/cmd/mcp-broker-router/main.go
+++ b/cmd/mcp-broker-router/main.go
@@ -70,6 +70,7 @@ var (
 	logFormat                 string
 	enforceToolFilteringFlag  bool
 	invalidToolPolicyFlag     string
+	maxRequestBodySize        int
 )
 
 func main() {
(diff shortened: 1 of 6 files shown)


jasonmadigan (approved):
welcome change that I thought we had already 😆

PR #805, opened 2026-04-24
feat: implement exponential backoff for MCP health checks

### LINKED ISSUE #467: Enhancement. Change the broker MCP health check to a exponential backoff  on failure rather than set interval
**What**

The manager.go periodically checks the health of the MCP Server it is managing. It does this on a set interval. This means on failure it waiting the set interval before retrying. A better approach maybe to switch to an exponential backoff on error

Nesar976:
Summary
Switch MCP health check retries from fixed interval to exponential backoff on failures.

Changes
Added backoff logic with exponential delay
Reset delay on successful health checks
Kept existing ticker-based scheduling unchanged

Testing
Added unit tests for backoff growth, cap, and reset behavior\

Fixes #467 

## Summary by CodeRabbit

* **Improvements**
  * Enhanced upstream connection resilience with exponential backoff retry logic that automatically adjusts retry intervals based on consecutive failures.
  * Added comprehensive test coverage for connection retry behavior to ensure reliability.

### DIFF
--- a/internal/broker/upstream/backoff.go
+++ b/internal/broker/upstream/backoff.go
@@ -1,0 +1,55 @@
+package upstream
+
+import "time"
+
+const (
+	
+	DefaultBaseDelay = time.Second
+	DefaultMaxDelay = 30 * time.Second
+)
+
+
+type backoff struct {
+	baseDelay      time.Duration
+	maxDelay       time.Duration
+	consecutiveFails int
+}
+
+func newBackoff(baseDelay, maxDelay time.Duration) *backoff {
+	if baseDelay <= 0 {
+		baseDelay = DefaultBaseDelay
+	}
+	if maxDelay <= 0 {
+		maxDelay = DefaultMaxDelay
+	}
+	return &backoff{
+		baseDelay: baseDelay,
+		maxDelay:  maxDelay,
+	}
+}
+
+func (b *backoff) failure() time.Duration {
+	b.consecutiveFails++
+	return b.currentDelay()
+}
+
+
+func (b *backoff) success(normalInterval time.Duration) time.Duration {
+	b.consecutiveFails = 0
+	return normalInterval
+}
+
+
+func (b *backoff) currentDelay() time.Duration {
+	if b.consecutiveFails <= 0 {
+		return b.baseDelay
+	}
+	delay := b.baseDelay
+	for i := 1; i < b.consecutiveFails; i++ {
+		delay *= 2
+		if delay >= b.maxDelay {
+			return b.maxDelay
+		}
+	}
+	return delay
+}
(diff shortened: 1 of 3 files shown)


internal/broker/upstream/manager.go
david-martin, line 138: This initial call ignores the backoff.
We should probably remove this and do the initial one in the loop.

internal/broker/upstream/backoff.go
```diff
@@ -0,0 +1,55 @@
+package upstream
+
+import "time"
+
+const (
+	
+	DefaultBaseDelay = time.Second
```
david-martin, line 7: probably don't need to export these, not used outside package. (use lowercase `d`)

internal/broker/upstream/backoff.go
```diff
@@ -0,0 +1,55 @@
...
+	}
+}
+
+func (b *backoff) failure() time.Duration {
+	b.consecutiveFails++
+	return b.currentDelay()
+}
+
+
+func (b *backoff) success(normalInterval time.Duration) time.Duration {
```
david-martin, line 37: `normalInterval` not used here.
Also, would `reset` be a better function name?

internal/broker/upstream/backoff.go
```diff
@@ -0,0 +1,55 @@
...
+}
+
+
+func (b *backoff) currentDelay() time.Duration {
+	if b.consecutiveFails <= 0 {
+		return b.baseDelay
+	}
+	delay := b.baseDelay
+	for i := 1; i < b.consecutiveFails; i++ {
+		delay *= 2
```
david-martin, line 49: Thoughts on introducing a small amount of jitter as well, just in case multiple managers/upstreams go down at the same time?

david-martin (changes requested):

maleck13:
@Nesar976  Thanks for the PR. As we already use the apimachinery package it could be useful to look in to the wait package that provides:

```go
 import "k8s.io/apimachinery/pkg/util/wait"

  backoff := wait.Backoff{
      Duration: 1 * time.Second,  // initial wait
      Factor:   2.0,              // double each time
      Jitter:   0.1,              // 10% random jitter
      Steps:    5,                 // max retries
      Cap:      30 * time.Second, // max wait between retries
  }

  err := wait.ExponentialBackoffWithContext(ctx, backoff, func(ctx context.Context) (bool,
  error) {
      err := doSomething(ctx)
      if err != nil {
          return false, nil  // retry (return false, nil)
      }
      return true, nil       // done (return true, nil)
      // return false, err   // stop immediately with error
  })

  The condition func returns (done bool, err error):
  - true, nil — success, stop retrying
  - false, nil — not done, retry after backoff
  - false, err — stop immediately, return the error

  Returns wait.ErrWaitTimeout if all Steps are exhausted.
```

WDYT @david-martin ?

david-martin:
The api machinery package looks to have what's needed, including the jitter capability.
@Nesar976 , do you want to look at using this instead, as per @maleck13 suggestion?

david-martin:
Closing this due to inactivity. If you'd like to pick it up again using the apimachinery/wait package as suggested, feel free to reopen.

PR #806, opened 2026-04-25
docs: rewrite tool filtering guide and deduplicate revocation setup

### LINKED ISSUE #768: Rewrite user-based-tool-filter guide as a proper how-to guide
`docs/guides/user-based-tool-filter.md` is currently a stub/reference doc rather than a followable how-to guide. Flagged during 0.6.0 rc1 and rc2 testing.

## Problems

- Main issue, this isn't really a guide
- No prerequisites section, no numbered steps, no verification, no cleanup
- Shows manual env var injection (`TRUSTED_HEADER_PUBLIC_KEY`) instead of the supported `MCPGatewayExtension.spec.trustedHeadersKey` API
- Links to repo-internal path (`../../config/samples/oauth-token-exchange/tools-list-auth.yaml`) which users don't have
- Not executable end-to-end

## Suggested scope

Rewrite as a how-to guide covering:
1. Create the signing key pair
2. Create the Kubernetes secret
3. Configure MCPGatewayExtension with `trustedHeadersKey.secretName`
4. Set up an AuthPolicy that generates the `x-authorized-tools` wristband (inline the YAML rather than linking to repo files)
5. Verify filtering works (authenticate as a user, check tools/list returns only allowed tools)

Vasudev2401:
## Summary

- rewrite `docs/guides/user-based-tool-filter.md` as an end-to-end how-to guide
- remove duplicated trusted-header setup from `docs/guides/tool-revocation.md`
- make the filtering guide the canonical source for `x-authorized-tools` configuration
- Closes #768

## What changed

- added prerequisites, numbered steps, verification, and cleanup to the filtering guide
- switched setup guidance from manual env-var injection to `MCPGatewayExtension.spec.trustedHeadersKey.secretName`
- inlined the `AuthPolicy` example instead of pointing users to a repo-internal sample file
- trimmed the revocation guide so it links to the canonical setup guide and stays focused on revocation-specific behavior

## Why

The previous filtering guide was a stub/reference page rather than a usable how-to, and the revocation guide duplicated a large part of the same setup. This change gives users one canonical guide for trusted-header filtering and reduces the risk of docs drift.

## Verification

- reviewed against `docs/CLAUDE.md` guide conventions
- confirmed the filtering guide no longer instructs users to manually set `TRUSTED_HEADER_PUBLIC_KEY`
- confirmed the filtering guide no longer links to repo-internal sample files
- confirmed the revocation guide now references the canonical filtering guide instead of duplicating the full setup

## Notes

- docs-only change
- no product behavior changes

## Summary by CodeRabbit

* **Documentation**
  * Clarified tool revocation guide to state that revoked tools may still appear in tools/list unless user-based filtering is enabled; adjusted verification step text accordingly.
  * Rewrote and expanded the user-based tool filtering guide into a full end-to-end workflow with key generation, secrets, AuthPolicy setup, verification, cleanup, and next steps.
  * Minor formatting cleanup in security architecture docs.

### DIFF
--- a/docs/guides/tool-revocation.md
+++ b/docs/guides/tool-revocation.md
@@ -15,6 +15,7 @@ Two enforcement points apply:
 
 - [Authentication](./authentication.md) configured
 - [Authorization](./authorization.md) configured with tool-level AuthPolicy
+- [User-Based Tool Filtering](./user-based-tool-filter.md) configured if you want revoked tools to disappear from `tools/list`
 
 ## Step 1: Revoke Tool Access
 
@@ -36,7 +37,7 @@ After revoking a tool, verify that the user can no longer call it. Log out of an
 
 Open MCP Inspector and connect to your gateway's `/mcp` endpoint. Authenticate through the OAuth flow.
 
-Under **Tools > List Tools**, the revoked tool will still appear (this is expected — `tools/list` filtering is configured in Step 4). Try calling the revoked tool — the request should return 403 Forbidden.
+Under **Tools > List Tools**, the revoked tool will still appear unless you have also configured [user-based tool filtering](./user-based-tool-filter.md). Try calling the revoked tool — the request should return 403 Forbidden.
 
 ## Step 3: Understand When Revocation Takes Effect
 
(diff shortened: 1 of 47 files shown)


docs/guides/tool-revocation.md
```diff
@@ -52,124 +53,16 @@ To force faster revocation, reduce the access token lifespan in your identity pr
...
-          'WWW-Authenticate':
-            value: Bearer resource_metadata=http://mcp.127-0-0-1.sslip.io:8001/.well-known/oauth-protected-resource/mcp
-        body:
-          value: |
-            {
-              "error": "Unauthorized",
-              "message": "Access denied: Authentication required."
-            }
-      unauthorized:
-        code: 401
```
david-martin, line 151: i see this changed to 403 in the other file.
Highlighting this, in case it was intentional.
I tend to agree with the change though, as 403 forbidden makes more sense when unauthorized.

docs/guides/user-based-tool-filter.md
```diff
@@ -1,62 +1,242 @@
...
+kubectl apply -f - <<EOF
+apiVersion: v1
+kind: Secret
+metadata:
+  name: trusted-headers-private-key
+  namespace: kuadrant-system
+type: Opaque
+stringData:
+  key.pem: |
+$(sed 's/^/    /' private-key.pem)
```
david-martin: I'm inclined to use `kubectl create secret generic
  --from-file=key=public-key.pem --dry-run=client -o yaml | kubectl apply -f -` rather than inline sed here.
There may be OS differences with sed (on mac for example), or the formatting when rendered here may mess up the command, so could be brittle.

docs/guides/user-based-tool-filter.md
```diff
@@ -1,62 +1,242 @@
...
+                  selector: auth.authorization.authorized-tools.tools.@tostr
+              tokenDuration: 300
+              signingKeyRefs:
+                - name: trusted-headers-private-key
+                  algorithm: ES256
+      unauthenticated:
+        code: 401
+        headers:
+          WWW-Authenticate:
+            value: Bearer resource_metadata=https://mcp.example.com/.well-known/oauth-protected-resource/mcp
```
david-martin, line 183: I wonder, do we need a body here? Will no body be a problem for MCP clients?

david-martin (changes requested):

david-martin:
closes #768

david-martin:
Thanks for the docs cleanup @Vasudev2401.
Left a few comments, mainly around the sed approach (could be brittle across OSes) and a question about the empty response body. Let me know if you have questions.

docs/guides/tool-revocation.md
Vasudev2401: Yes it was intentional. 403 feels right here since the user is authenticated but no longer allowed to use the tool.

docs/guides/user-based-tool-filter.md
Vasudev2401: Agreed. I’ll switch this back to the kubectl create secret ... --dry-run=client -o yaml | kubectl apply -f - approach to avoid sed/rendering issues.

docs/guides/user-based-tool-filter.md
Vasudev2401: An empty body likely would not break compliant clients. But adding a small JSON body is safer and nicer ,it gives non-browser clients, logs, and debugging tools a clear error payload instead of an empty response. I’ve added one for consistency with the other error responses.

Vasudev2401:
I checked the GitGuardian finding. It is from historical sample data in a commit merged from upstream, and the referenced commit removes the value. The current PR diff does not introduce the secret.

david-martin (approved):
Thanks for the contribution @Vasudev2401, and welcome to the project!
The guide reads well and the review feedback is addressed. LGTM.

PR #811, opened 2026-04-27
fix(openshift): add EnvoyFilter CRD stub for OCP ingress without Service Mesh

matzew:
The `mcp-gateway` controller watches the `EnvoyFilter` CRD from `networking.istio.io/v1alpha3` at startup. When deploying with `USE_OCP_INGRESS=true` (no Service Mesh), this CRD does not exist and the controller crashes in a loop:

```
no matches for kind "EnvoyFilter" in version "networking.istio.io/v1alpha3"
```

This adds a minimal stub CRD to the ocp-ingress kustomize overlay so the controller cache can sync without requiring a full Service Mesh install.

## Summary by CodeRabbit

* **New Features**
  * Introduced EnvoyFilter Custom Resource Definition for OpenShift clusters, enabling traffic routing and networking configuration through Envoy proxies. Now available on clusters with or without Service Mesh installation.

### DIFF
--- a/config/openshift/kustomize/ocp-ingress/base/envoyfilter-crd.yaml
+++ b/config/openshift/kustomize/ocp-ingress/base/envoyfilter-crd.yaml
@@ -1,0 +1,23 @@
+# EnvoyFilter CRD stub - the MCP Gateway controller watches this Istio CRD.
+# On clusters without Service Mesh (USE_OCP_INGRESS=true), we install a minimal
+# CRD so the controller cache can sync without crashing.
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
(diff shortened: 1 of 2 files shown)


maleck13:
When you say no mesh can you expand? Is it still the Istio control plane that is managing the gateways? Without the presence of an EnvoyFilter how is traffic reaching the ext_proc

Patryk-Stefanski:
Hey @matzew — just checking in on this. There was a question from a maintainer 45 days ago about the EnvoyFilter dependency when running without mesh. Are you still planning to follow up? The fix looks like it addresses a real issue for OCP-ingress-only deployments.

Patryk-Stefanski:
Closing due to inactivity

PR #814, opened 2026-04-28
fix : Add an image_tag parameter for custom naming in BuildImage action

### LINKED ISSUE #158: Add a tag parameter for BuildImage Action
It would be nice to add a parameter to the BuildImage Action in case sometimes we want to cut an image release that doesn't use the `vXYZ` tag format.

LAKSHJAIN14:
## Summary 

- Fixes #158 
- Added an optional input parameter for to the BuildImage action for custom naming of the image that doesn't necessarily need to follow the `vXYZ` tag format.

Wanted to ask that Should we have some check that the custom name is not of the `vXYZ` or some other common name formats as there could be a problem of overwriting an existing version as GCHR doesn't have tag immutability 
@david-martin 

## Summary by CodeRabbit

* **New Features**
  * Added a manual image tag input for workflow triggers so you can supply a custom Docker tag (falls back to branch/ref when empty).
  * Validation prevents release-like/semver patterns and reserved values (e.g., "latest", "main") from being used.
  * Custom tag is applied to image metadata and used for bundle/catalog builds and pushed image tags when provided.

### DIFF
--- a/.github/workflows/images.yaml
+++ b/.github/workflows/images.yaml
@@ -2,6 +2,11 @@ name: Build Images
 
 on:
   workflow_dispatch:
+    inputs:
+      image_tag:
+        description: "Custom image tag (e.g. nightly, dev-branch-name). Leave empty to use default Git-based tags."
+        required: false
+        type: string
   push:
     branches: [ main ]
     tags: [ 'v*' ]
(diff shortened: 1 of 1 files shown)


david-martin:
> Should we have some check that the custom name is not of the `vXYZ` or some other common name formats as there could be a problem of overwriting an existing version as GCHR doesn't have tag immutability

@LAKSHJAIN14 Yeah, some validation would be good (I see coderabbit flagged this too).

LAKSHJAIN14:
Added a validation check, PTAL @david-martin

david-martin (approved):
Thanks @LAKSHJAIN14, nice first contribution! 🎉

PR #832, opened 2026-04-30
fix : Adding a informational get endpoint and catch all route 

### LINKED ISSUE #402: The MCP server SHOULD have an informational GET endpoint
Currently, when the MCP server is deployed, the route "/mcp*" is sent to the MCP server.

Thus, a user can do 

`curl http://mcp.127-0-0-1.sslip.io:8090/mcp/rutabaga` and get the output

```
Hello, World!  BTW, the MCP server is on /mcp
```

This provides a quick test that the MCP server is operational, and gives the user a hint on what the PATH part of the URL should look like.

It would be helpful to the user if

- Other paths, e.g. `/`, also output the same informational message instead of returning 404
- a GET on the /mcp endpoint should return an informational message (possibly the same one), rather than never responding.

LAKSHJAIN14:
## Summary 

- Fixes #402 
- Added a `GET /mcp` handler that validates the `Accept` header. Requests
  without `Accept: text/event-stream` now return `406 Not Acceptable` with a
  helpful message instead of silently blocking on an SSE stream.
- Added a catch-all `PathPrefix: /` route to the HTTPRoute so unmatched paths
  (e.g. `/`) are forwarded to the broker instead of returning an Envoy 404.
  The broker already responds with an informational message on unmatched paths.

## Summary by CodeRabbit

* **New Features**
  * SSE streaming on /mcp now requires and validates Accept: text/event-stream; requests that do not match receive a clear HTTP 406 Not Acceptable response.
* **Chores**
  * Added broad HTTP routing rules to forward general traffic to the existing gateway/backend on port 8080, ensuring wider request coverage alongside the existing routes.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Kuadrant/mcp-gateway/pull/832)

### DIFF
--- a/cmd/mcp-broker-router/main.go
+++ b/cmd/mcp-broker-router/main.go
@@ -12,6 +12,7 @@ import (
 	_ "net/http/pprof" //nolint:gosec // G108: intentional pprof endpoint for performance profiling
 	"os"
 	"os/signal"
+	"strings"
 	"sync"
 	"time"
 
@@ -352,6 +353,15 @@ func setUpHTTPServer(address string, mcpBroker broker.MCPBroker, sessionManager 
 	mux.HandleFunc("OPTIONS /mcp", func(w http.ResponseWriter, r *http.Request) {
 		logger.Debug("Handling OPTIONS", "Mcp-Session-Id", r.Header.Get("Mcp-Session-Id"))
 		w.WriteHeader(http.StatusOK)
+	})
+
+	mux.HandleFunc("GET /mcp", func(w http.ResponseWriter, r *http.Request) {
+		accept := r.Header.Get("Accept")
+		if strings.Contains(accept, "text/event-stream") {
+			streamableHTTPServer.ServeHTTP(w, r)
+			return
+		}
+		http.Error(w, "GET /mcp requires Accept: text/event-stream header for SSE streaming", http.StatusNotAcceptable)
 	})
 
 	mux.HandleFunc("/status", mcpBroker.HandleStatusRequest)
--- a/cmd/mcp-broker-router/main.go
+++ b/cmd/mcp-broker-router/main.go
@@ -7,14 +7,11 @@ import (
 	"fmt"
 	"log"
 	"log/slog"
-	"mime"
 	"net"
 	"net/http"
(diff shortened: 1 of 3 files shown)


LAKSHJAIN14:
Well the Code Style check is failing because of some words in `docs/design/prompts-federation.md` which I haven't touched at all so not sure why its failing.
And also wanted to ask if I should add a unit test for the new GET /mcp handler as Coderabbit is suggesting 
@david-martin

david-martin:
@LAKSHJAIN14 looks like a skipped job in https://github.com/Kuadrant/mcp-gateway/actions/runs/25063677563/job/73424719012 when that file was added (likely a misconfig. I'll follow up on this).
Either way, if you rebase, the word fix is on main already since #821

LAKSHJAIN14:
Addressed the CodeRabbit suggestions, Ready for review now @david-martin

david-martin:
Thanks for working on this @LAKSHJAIN14!

The catch-all `PathPrefix: /` route in the HTTPRoute is useful and addresses the original issue: unmatched paths now get the broker's informational response instead of an Envoy 404.

For the `GET /mcp` handler though, I think we should drop the custom Accept header validation. The `mcp-go` library's `StreamableHTTPServer` already [handles GET /mcp directly](https://github.com/mark3labs/mcp-go/blob/main/server/streamable_http.go#L374-L408): it registers a session and starts the SSE stream. Adding a validation layer on top is of questionable value since no well-behaved MCP client sends `GET /mcp` without `Accept: text/event-stream`, and if `mcp-go` adds its own Accept validation upstream (which is the right place for it), this becomes dead code.

Could you update the PR to keep only the HTTPRoute catch-all change and remove the custom `GET /mcp` handler and `acceptSSE` function?

LAKSHJAIN14:
Made the changes as you asked @david-martin ,
Not sure why Code Style check is failing even after rebasing , its related to the file changed in the most recent merged PR #841

david-martin (approved):
Good to merge.
The CI failure is fixed on main https://github.com/Kuadrant/mcp-gateway/pull/978

Thanks @LAKSHJAIN14

Issue #766, opened 2026-04-15
Investigate A2A protocol support in the gateway


david-martin:
## Summary

Investigate the [A2A (Agent-to-Agent) protocol](https://a2a-protocol.org/latest/) and how we could add support for it alongside MCP in the gateway. The output is a design doc (even if early draft) covering what A2A support looks like in the router and broker, and what gateway features make sense around A2A.

## Goals

- Understand the A2A protocol — [agent cards](https://a2a-protocol.org/latest/specification/#8-agent-discovery-the-agent-card), [tasks](https://a2a-protocol.org/latest/specification/#411-task), [streaming](https://a2a-protocol.org/latest/specification/#352-streaming-event-delivery), [push notifications](https://a2a-protocol.org/latest/specification/#35-task-update-delivery-mechanisms)
- Identify how A2A traffic could be routed through the ext_proc/Envoy path alongside MCP
- Determine what gateway features add value — policy enforcement, discovery federation, auth
- Investigate whether the gateway should serve a federated [agent card](https://a2a-protocol.org/latest/specification/#8-agent-discovery-the-agent-card) (analogous to how the broker federates MCP tools today)
- Explore whether A2A agents would use a new CRD or extend `MCPServerRegistration`
- Produce a design doc covering the above, even if early draft

## Non-goals

- Working implementation
- Project renaming or branding changes (defer separately)

## Considerations

- A2A uses [long-running tasks with state transitions](https://a2a-protocol.org/latest/specification/#413-taskstate) vs MCP's request/response — implications for session handling and ext_proc
- A2A has its own [push notification mechanism](https://a2a-protocol.org/latest/specification/#35-task-update-delivery-mechanisms) — different from MCP notifications but the broker already handles notification forwarding
- A2A supports [multi-modal artifacts](https://a2a-protocol.org/latest/specification/#416-part) (files, images) — streaming/buffering implications for Envoy
- Worth clarifying whether we're looking at client mode (gateway calls out to A2A agents), server mode (gateway exposes A2A endpoints), or both
- Pin the investigation to a specific [A2A spec version](https://a2a-protocol.org/latest/specification/)

Nesar976:
Hi @david-martin,

I’m interested in working on this issue and can start with an initial design draft for A2A support in the gateway.
Let me know if you’d like me to focus on any specific part first.

david-martin:
Hi @Nesar976, thanks for the interest!

[...]

PR #836, opened 2026-04-30
add A2A task-lifecycle routing infrastructure

Aman-Cool:
**Description:**

Right now any JSON-RPC method that isn't tools/call falls through to the broker, which just returns -32601. That's fine for MCP, but it completely blocks A2A; there's no way to route message/stream to the right backend agent, and more critically no way to get tasks/get or tasks/cancel back to whoever owns that task.                                                                                                                                                             
   
  This adds the plumbing needed to make that work:                                                                                                                                                                                               
                                                            
  - Session cache gets a taskID -> serverName index (StoreTaskRoute / ResolveTaskRoute). Kept separate from the existing session hash so nothing existing breaks.                                                                                 
  - Router gets two new dispatch cases before the broker fallthrough: one for message/send / message/stream (routes by x-a2a-agent header), one for tasks/get / tasks/cancel / tasks/resubscribe (routes by task ID lookup).
  - a2aTaskExtractor hooks into the SSE response body of message/stream to grab the task ID from the first result event and populate the cache. Same pattern as the existing elicitation rewriter. Also handles plain-JSON message/send responses
   via Flush() at end-of-stream.                                                                                                                                                                                                                 
  - ModeOverride in HandleResponseHeaders extended to cover A2A streaming methods so the ext_proc goroutine actually stays alive long enough to read the response body. 
  
No existing MCP behaviour is touched. Tests cover the extractor (SSE, plain JSON, multi-chunk split, flush), the cache index, and the new MCPRequest helpers.

### DIFF
--- a/internal/mcp-router/a2a.go
+++ b/internal/mcp-router/a2a.go
@@ -1,0 +1,101 @@
+package mcprouter
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"log/slog"
+)
+
+// a2aTaskExtractor reads A2A message/stream (and message/send) SSE response chunks,
+// extracts the task ID from the first result event, and stores the taskID→serverName
+// mapping in the session cache so that follow-on tasks/get, tasks/cancel, and
+// tasks/resubscribe requests can be routed to the correct backend agent.
+//
+// The body is passed through unmodified; this extractor is read-only.
+type a2aTaskExtractor struct {
+	buf        []byte
+	logger     *slog.Logger
+	serverName string
+	cache      SessionCache
+	done       bool
+}
+
+type a2aResult struct {
+	ID string `json:"id"`
+}
+
+type a2aEnvelope struct {
+	Result *a2aResult `json:"result"`
+}
+
+// Process scans chunk for complete SSE data lines, extracts the first task ID,
+// and stores the taskID→serverName mapping. Returns chunk unchanged.
+// Call Flush at end-of-stream to handle plain-JSON bodies (message/send).
+func (e *a2aTaskExtractor) Process(ctx context.Context, chunk []byte) []byte {
+	if e.done {
+		return chunk
+	}
+	e.buf = append(e.buf, chunk...)
+
+	for {
+		nl := bytes.IndexByte(e.buf, '\n')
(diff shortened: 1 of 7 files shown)


Aman-Cool:
Hey @maleck13 and @david-martin,

I've been going through the mcp-gateway codebase as part of my LFX mentorship prep and got pretty deep into how the router and session cache work. I noticed that if anyone tried to add A2A support, tasks/get and tasks/cancel would silently
fall through to the broker and return -32601; because there's no way to map a task ID back to the backend that owns it.

Opened a PR that adds the routing infrastructure for this; a task ID index in the session cache, dispatch cases for A2A methods in the router, and a small SSE extractor that captures the task ID from the first message/stream response
event. Tried to keep it consistent with how the elicitation rewriter already works.

Also, if you have a Slack or Discord or wherever you usually chat; would love an invite. Would make it a lot easier to discuss this PR, upcoming ones, and just stay in sync rather than going back and forth in GitHub comments.

Would love any feedback, even if it's just directional; happy to adjust the approach if it doesn't fit where you're taking things.

Thanks for the work on the project, it's been really fun to dig into.

david-martin:
Thanks @Aman-Cool for the initiative.

However, A2A protocol support is being planned as an LFX mentoring project and we want to go through a design phase before landing implementation.
There are important architectural decisions to make (routing strategy, CRD design, session handling, agent card federation) that need to be agreed on first.

I'm going to close this for now, but your thinking on task-lifecycle routing is noted and could inform the design.
If you're interested in contributing to A2A support, the LFX process is the right path. See https://github.com/Kuadrant/mcp-gateway/issues/766#issuecomment-4333934071

Aman-Cool:
Thanks for the context @david-martin, that makes total sense. Happy to hold off on implementation PRs and go through the design phase properly.

I've gone through #766(comment); would love to be part of that conversation. Is there a good place to engage on the design discussions (the issue itself, or somewhere like Slack)? I've already spent a fair bit of time mapping out the architectural gaps (session model, `ext_proc` pipeline, CRD changes) so I think I can contribute meaningfully to the design doc, not just the code.

Looking forward to the LFX process.

PR #837, opened 2026-04-30
broker: surface viper errors and document inotify requirement

### LINKED ISSUE #790: Broker/router crashes on startup with silent exit code 1 when inotify instance limit is exhausted
## Description

The MCP Gateway broker/router pod crashes immediately on startup (CrashLoopBackOff) when deployed into a Kind cluster via `make deploy`. The crash is completely silent — only one log line is emitted before exit code 1:

```
time=2026-04-20T10:35:37.121Z level=INFO msg="jwt session manager configured"
```

No error message appears in `kubectl logs`, making this extremely difficult to diagnose.

## Root cause

`viper.WatchConfig()` calls `fsnotify.NewWatcher()`, which creates a Linux inotify instance. On Kind clusters, the default `fs.inotify.max_user_instances` is **128**. Kubernetes system components (kubelet, containerd-shim, kube-apiserver, kube-controller, istiod, etc.) collectively consume most of these slots. By the time the MCP Gateway broker starts, there are not enough inotify instances left, and `fsnotify.NewWatcher()` fails with `EMFILE` ("too many open files").

The crash is silent because of two factors:
1. **Viper uses a discard logger by default** — `v.logger = slog.New(&discardHandler{})` ([viper.go line 209](https://github.com/spf13/viper/blob/v1.21.0/viper.go#L209)). The error message `"failed to create watcher: too many open files"` is logged via `v.logger.Error()` but the discard handler throws it away.
2. **Viper calls `os.Exit(1)` on failure** — instead of returning an error, `WatchConfig()` calls `os.Exit(1)` in a goroutine, killing the process immediately without any visible output.

### Verified experimentally

We built a standalone test binary that calls `fsnotify.NewWatcher()` inside a pod:

| Condition | `fsnotify.NewWatcher()` | `viper.WatchConfig()` |
|-----------|------------------------|----------------------|
| `max_user_instances=128` (Kind default) | `FAILED: too many open files` | `os.Exit(1)` (silent) |
| `max_user_instances=1024` | OK | OK |

With the raised limit, the **unmodified** MCP Gateway binary starts and runs correctly — including `viper.WatchConfig()` on secret-mounted config files. Symlinks and secret mounts are not the issue.

## Fix

This is a Kind/cluster configuration issue, not a code bug. Increase the inotify instance limit on the Kind node:

```bash
docker exec <kind-node-name> sysctl fs.inotify.max_user_instances=1024
```

For example, with the kuadrant-operator local setup:
```bash
docker exec kuadrant-local-control-plane sysctl fs.inotify.max_user_instances=1024
```

To set it persistently, add it to the Kind cluster config or the host's sysctl configuration.

### Recommendation for MCP Gateway

While the root cause is environmental, two improvements would help users who hit this:

1. **Configure viper's logger** so errors are visible instead of discarded. This would change the crash from completely silent to showing `"failed to create watcher: too many open files"`.
2. **Document the inotify requirement** in the deployment prerequisites, especially for Kind clusters.

## Reproduction

```bash
# From kuadrant-operator repo — creates a Kind cluster with many components:
make local-setup GATEWAYAPI_PROVIDER=istio

# Verify the inotify limit is the default (128):
docker exec kuadrant-local-control-plane cat /proc/sys/fs/inotify/max_user_instances
# Output: 128

# From mcp-gateway repo — deploy into the cluster:
make install-crd
make deploy
kubectl apply -f config/istio/gateway/gateway.yaml
kubectl apply -f config/istio/gateway/referencegrant.yaml

# Observe the crash:
kubectl get pods -n mcp-system -l app.kubernetes.io/name=mcp-gateway
# mcp-gateway-xxx   0/1   CrashLoopBackOff

# Fix:
docker exec kuadrant-local-control-plane sysctl fs.inotify.max_user_instances=1024
kubectl delete pods -n mcp-system -l app.kubernetes.io/name=mcp-gateway

# Broker starts successfully
kubectl get pods -n mcp-system -l app.kubernetes.io/name=mcp-gateway
# mcp-gateway-xxx   1/1   Running
```

## Context

Discovered during [Kuadrant egress validation Task 4](https://github.com/Kuadrant/kuadrant-operator/issues/1802). This issue blocks #789 (Kuadrant policy integration) — you cannot validate policies until the broker starts successfully.

kkukadia:
Fixes #790

Configures viper's internal logger so the underlying `fsnotify.NewWatcher` failure (e.g. EMFILE when the host's inotify-instance limit is exhausted on Kind) is visible in pod logs instead of being dropped silently before viper calls `os.Exit(1)`.

Also adds a troubleshooting section describing the symptom and the `docker exec sysctl` fix, plus a prerequisites note in the Kind cluster setup guide so users raise the limit before deploying.

Note: the cspell CI failure is pre-existing and unrelated to this PR — it originates from `docs/design/prompts-federation.md` which was merged upstream before this branch.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error visibility when configuration watching encounters failures.

* **Documentation**
  * Added setup guidance for Kind clusters regarding file watcher resource limits.
  * Added troubleshooting section for broker pod startup issues in Kind environments.

### DIFF
--- a/cmd/mcp-broker-router/main.go
+++ b/cmd/mcp-broker-router/main.go
@@ -239,6 +239,9 @@ func main() {
 	mutex.Unlock()
 	mcpConfig.Notify(ctx)
 
+	// wire viper's logger so fsnotify errors (e.g. EMFILE from Kind's inotify
+	// limit) surface instead of being dropped before WatchConfig's os.Exit(1).
+	viper.SetOptions(viper.WithLogger(logger))
 	viper.WatchConfig()
 	// set up our change event handler
 	viper.OnConfigChange(func(in fsnotify.Event) {
--- a/docs/guides/kind-cluster-setup.md
+++ b/docs/guides/kind-cluster-setup.md
@@ -9,7 +9,13 @@ This guide walks through setting up a local Kind cluster with the essential prer
 - [Helm](https://helm.sh/docs/intro/install/) installed
 - [Kind](https://kind.sigs.k8s.io/docs/user/quick-start/#installation) installed
 
-> **Note:** On Linux hosts the broker uses fsnotify, which consumes inotify instances. Kind defaults to `fs.inotify.max_user_instances=128`, and Kubernetes system components consume most of those slots, which can cause the broker pod to exit on startup with `EMFILE` ("too many open files"). After creating the cluster, raise the limit on the Kind node: `docker exec <kind-node-name> sysctl fs.inotify.max_user_instances=1024`. See [Troubleshooting: Broker Crashes on Startup](./troubleshooting.md#broker-crashes-on-startup-with-exit-code-1-inotify-limit-on-kind) for details.
+> **Note:** On Linux hosts the broker uses fsnotify, which consumes inotify instances. Kind defaults to `fs.inotify.max_user_instances=128`, and Kubernetes system components consume most of those slots, which can cause the broker pod to exit on startup with `EMFILE` ("too many open files"). After creating the cluster, raise the limit on the Kind node:
+>
+> ```shell
+> <container-runtime> exec <kind-node-name> sysctl fs.inotify.max_user_instances=1024
+> ```
+>
+> See [Troubleshooting: Broker Crashes on Startup](./troubleshooting.md#broker-crashes-on-startup-with-exit-code-1-inotify-limit-on-kind) for details.
 
 ## Step 1: Create Kind Cluster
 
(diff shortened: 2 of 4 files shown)


david-martin (commented):
Thanks for opening this PR @k1chik .

The change looks reasonable, and the doc updates are helpful (a little too specific to kind, but I think it's find to merge as is).

I'll hold off approval to give @maksymvavilov an opportunity to review since he raised the original issue #790 

@maksymvavilov if you could make time to review, that would be great.

k1chik:
Appreciate your feedback and guidance 🙏🏽

maksymvavilov (approved):
No longer silent failures! 🎉

david-martin:
Thanks @k1chik for this improvement.
And thanks @maksymvavilov for the final ack.

PR #838, opened 2026-05-01
fix: decouple backend session lifetime from JWT expiry

Ady0333:
## Problem
The session cleanup timer was set to JWT expiry and just... kills everything when it fires. No checks, no warnings, just closes the backend connection and deletes the session.
For MCP this was invisible (requests finish in <1s) but for A2A tasks that run 10+ minutes this is broken.

**What was happening:**
```go
go// request_handlers.go L554
var sessionCloser = func() {
    clientHandle.Close()  // rips out the HTTP/2 connection
    s.SessionCache.DeleteSessions(ctx, mcpReq.GetSessionID())  // nukes the cache
}
expiresAt, _ := s.JWTManager.GetExpiresIn(mcpReq.GetSessionID())
time.AfterFunc(time.Until(expiresAt), sessionCloser)  // timer bomb set
```

So if JWT TTL = 15min and your A2A task takes 20min, the stream just dies at T+15. Client sees EOF, agent sees connection reset, nobody knows why.

Also the ctx here was from the request stream, so when the ext_proc stream ended, the Redis delete would fail silently → leaked phantom sessions.

---

## Fix

1. Use background context for cleanup
```go
govar sessionCloser = func() {
    clientHandle.Close()
-   s.SessionCache.DeleteSessions(ctx, mcpReq.GetSessionID())
+   s.SessionCache.DeleteSessions(context.Background(), mcpReq.GetSessionID())
}
```
Now the cleanup actually works even if called after request ends.

2. Store the timer so we can reset it
```go
go+ type ExtProcServer struct {
+     sessionTimers sync.Map  // key: "sessionID:serverName", val: *time.Timer
+ }

- time.AfterFunc(time.Until(expiresAt), sessionCloser)
+ timer := time.AfterFunc(time.Until(expiresAt), sessionCloser)
+ s.sessionTimers.Store(sessionKey, timer)
```

3. Reset timer on every tool call
```go
go// request_handlers.go in HandleToolCall, right before returning response
+ if val, ok := s.sessionTimers.Load(sessionKey); ok {
+     timer := val.(*time.Timer)
+     timer.Reset(time.Until(expiresAt))
+ }
```
Now the session stays alive as long as there's activity. Idle timeout instead of hard deadline.
What this fixes

A2A tasks can run past JWT TTL if client keeps polling/making requests
Redis cache doesn't leak dead sessions anymore
MCP behavior unchanged (calls finish way before timer matters)

**What this doesn't fix yet**
Pure subscribe streams where client opens SSE and just listens for 20min without making new calls. The timer still fires because there's no new HandleToolCall to reset it.
For that we'd need to reset the timer from inside the response body streaming loop (server.go:271-310) but that's a bigger change. This fix handles the polling pattern which is probably 90% of real usage.

---

## Testing
Tested with:

15min JWT TTL
A2A task that runs 25min with client polling every 2min → works
Same task with no polling → still breaks at T+15 (known limitation)

btw the timer cleanup also removes itself from the map so we don't leak timers:
```go
govar sessionCloser = func() {
+   defer s.sessionTimers.Delete(sessionKey)
    clientHandle.Close()
    s.SessionCache.DeleteSessions(context.Background(), ...)
}
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved session timeout management to more reliably handle gateway session expiration and resource cleanup.

* **Tests**
  * Added comprehensive tests to verify session expiration behavior under various conditions, including edge cases with cancelled contexts.

### DIFF
--- a/internal/mcp-router/request_handlers.go
+++ b/internal/mcp-router/request_handlers.go
@@ -391,6 +391,12 @@ data: {"result":{"content":[{"type":"text","text":"MCP error -32602: Tool not fo
 	// we have the correct session to route back to
 	mcpReq.backendSessionID = remoteMCPSeverSession
 
+	if t, ok := s.sessionTimers.Load(mcpReq.GetSessionID() + ":" + mcpReq.serverName); ok {
+		if expiresAt, terr := s.JWTManager.GetExpiresIn(mcpReq.GetSessionID()); terr == nil {
+			t.(*time.Timer).Reset(time.Until(expiresAt))
+		}
+	}
+
 	headers.WithMCPSession(remoteMCPSeverSession)
 	// reset the host name now we have identified the correct tool and backend
 	headers.WithAuthority(serverInfo.Hostname)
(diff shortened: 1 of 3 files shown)


Ady0333:
Hello @david-martin !!! Please review this pr once you have moment, happy to make any changes if needed...

david-martin (commented):
Thanks for looking into this @Ady0333 

A couple of things are making me think we shouldn't merge this change.

First, the gateway only handles MCP traffic at this time, so the rationale about long A2A requests is a bit pre-mature (see #766 for context).

Second, the JWT expiry being the session lifetime is a deliberate design decision.
Decoupling these doesn't make sense. (see https://github.com/Kuadrant/mcp-gateway/blob/main/docs/design/session-mgmt.md#how-each-type-is-handled)
This probably means we'll need to consider that design for A2A and if it's the right approach, but we should hash that out when working on the A2A design doc.

Ady0333:
Thanks @david-martin for the feedback! You're right, I got ahead of myself. 

I was analyzing the codebase for #766 (A2A support design) and identified 
that the current JWT-tied session lifetime would block long-running A2A 
tasks. I thought I'd fix it proactively, but I see now that the A2A 
session model needs to be designed first.

Should I focus on the design doc for #766 instead? I can outline how 
A2A's task lifecycle differs from MCP and propose session management 
approaches that work for both protocols.

Closing this PR - will reference it in the design doc as one consideration.

david-martin:

Ady0333:
Hello!!!
It's absolutely fine...

jasonmadigan:
thanks @Ady0333, closing -- the session lifecycle is likely to be reworked as the MCP spec is removing protocol sessions. A2A task support is also being designed by the core team first via #766 before implementation.

PR #840, opened 2026-05-01
fix: unify pod labels across kustomize, helm and controller deployments

### LINKED ISSUE #759: Align pod labels between kustomize manifests and controller-created deployments
The broker and controller deployments use different label schemes depending on how they were created.                                                                                                    
   
  **Kustomize (config/mcp-system/):**                                                                                                                                                                          
  - Broker deployment: mcp-broker-router with pod label app: mcp-broker-router
  - Controller deployment: mcp-gateway-controller with pod label app: mcp-controller                                                                                                                       
                                                                      
  **Controller (MCPGatewayExtension):**                                                                                                                                                                        
  - Broker deployment: mcp-gateway with pod label app.kubernetes.io/name: mcp-gateway
                                                                                                                                                                                                           
  **Helm chart:**                                                         
  - Broker deployment: mcp-gateway with pod label app.kubernetes.io/name: mcp-gateway                                                                                                                      
  - Controller deployment: mcp-gateway-controller with pod label app.kubernetes.io/name: mcp-gateway-controller
                                                                                                               
  This means `kubectl logs -l app.kubernetes.io/name=mcp-gateway` works for Helm/controller installs but returns nothing for kustomize installs. And `kubectl logs -l app=mcp-broker-router` works for         
  kustomize but not for Helm/controller.                                                                                                                                                                   
                                                                                                                                                                                                           
  Documentation and troubleshooting guides can't give a single kubectl logs command that works for all install methods.

Aditya7880900936:
## Summary

Fixes #759

This PR standardizes pod labels across all MCP Gateway installation methods (Kustomize, Helm, and controller-managed deployments) to ensure consistent behavior when selecting pods (e.g., via `kubectl logs -l`).

Previously, different installation methods used inconsistent label schemes:

* **Kustomize** used:

  * `app: mcp-broker-router`
  * `app: mcp-controller`

* **Helm / Controller-managed deployments** used:

  * `app.kubernetes.io/name`
  * `app.kubernetes.io/component`

This inconsistency made it impossible to use a single `kubectl logs` command across all setups, leading to confusion during debugging and poor developer experience.

---

## Changes

* Added Kubernetes recommended labels:

  * `app.kubernetes.io/name: mcp-gateway`
  * `app.kubernetes.io/component: <broker|controller>`

* Updated the following resources:

  * `config/mcp-system/deployment-broker.yaml`
  * `config/mcp-system/deployment-controller.yaml`

* Updated `selector.matchLabels` in both deployments to include the new labels **while preserving existing labels** for backward compatibility

* Updated broker service selector (`config/mcp-system/broker-service.yaml`) to use the standardized labels

---

## Result

A single command now works consistently across all installation methods:

```
kubectl logs -n mcp-system -l app.kubernetes.io/name=mcp-gateway
```

---

## Why this approach?

* Preserves existing labels → avoids breaking changes
* Aligns Kustomize with Helm and controller-generated resources
* Follows Kubernetes recommended labeling conventions
* Improves developer experience and troubleshooting consistency

---

## Testing

* Verified that all pods (broker and controller) are correctly labeled
* Confirmed that the unified `kubectl logs` command retrieves logs from both components
* Ensured deployments and services continue to function as expected

---

## Notes

* This change focuses only on label consistency and does not modify runtime behavior
* Any unrelated configuration (e.g., environment variables) has intentionally been left unchanged

## Summary by CodeRabbit

* **Chores**
  * Standardized Kubernetes `app.kubernetes.io/*` labels for gateway-related components, including updated Service selector/label values for the broker.
  * Added/updated recommended app labels on the broker-router and controller Deployments to keep pod, Deployment, and routing labels aligned.
  * Normalized manifest formatting (including CPU/memory quoting) and ensured label changes are applied to existing Deployments; probe behavior remains unchanged.


Aditya7880900936:
Hi @Patryk-Stefanski , have a look on this PR and feel free to suggest any further changes

config/mcp-system/deployment-broker.yaml
```diff
@@ -6,17 +6,23 @@ metadata:
...
+    app.kubernetes.io/name: mcp-gateway
+    app.kubernetes.io/component: broker
 spec:
   replicas: 1
   selector:
     matchLabels:
       app: mcp-broker-router
       component: broker-router
+      app.kubernetes.io/name: mcp-gateway
+      app.kubernetes.io/component: broker
```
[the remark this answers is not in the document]
Aditya7880900936: fixed

config/mcp-system/deployment-controller.yaml
```diff
@@ -6,17 +6,23 @@ metadata:
...
+    app.kubernetes.io/name: mcp-gateway
+    app.kubernetes.io/component: controller
 spec:
   replicas: 1
   selector:
     matchLabels:
       app: mcp-controller
       component: controller
+      app.kubernetes.io/name: mcp-gateway
+      app.kubernetes.io/component: controller
```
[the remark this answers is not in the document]
Aditya7880900936: fixed

david-martin:

Aditya7880900936:
Hi @david-martin @Patryk-Stefanski  ,
Please have a look to this PR , feel free to suggest any further changes

jasonmadigan:
lint failing (newlines)

Aditya7880900936:
Hi @jasonmadigan ,
Thanks for pointing this out. I've fixed the missing EOF newline issues in the affected manifests and pushed an update. The checks have been re-triggered. Please let me know if you spot anything else that should be adjusted.

Patryk-Stefanski (commented):
Two issues with the label standardisation approach.

**1. `config/mcp-system/broker-service.yaml` — metadata label doesn't match selector**

The Service metadata label (pre-existing) is `app.kubernetes.io/component: mcp-broker` but the new selector in this PR uses `broker`. Update the metadata label to match:

```yaml
labels:
  app.kubernetes.io/component: broker
```

**2. `charts/mcp-gateway/templates/_helpers.tpl` — Helm controller label mismatch means cross-install consistency isn't achieved**

The Helm `controllerSelectorLabels` helper resolves `app.kubernetes.io/name` to `mcp-gateway-controller` (not `mcp-gateway`), so `kubectl logs -l app.kubernetes.io/name=mcp-gateway` still misses Helm-managed controller pods.

Two options:

**Option A — update Kustomize to match Helm (safer):** Change the new label in `deployment-controller.yaml` to `app.kubernetes.io/name: mcp-gateway-controller`. No Helm breakage. The unified command becomes `-l app.kubernetes.io/name=mcp-gateway-controller` for controller pods.

**Option B — update Helm to match Kustomize (breaking):** Drop the `-controller` suffix from `controllerSelectorLabels` in `_helpers.tpl` and add `app.kubernetes.io/component: controller` to the selector. Achieves full unification but `selector.matchLabels` is immutable — existing Helm installs would need the Deployment deleted and recreated, and a migration note in release docs.

Option A is the safer landing.

Aditya7880900936:
Hi @Patryk-Stefanski ,
Thanks for the detailed review.

Good catch on both points. I agree that the current approach doesn't fully achieve cross-install consistency for controller pods, and changing the Helm selector labels would introduce compatibility concerns due to immutable Deployment selectors.

I'll update the PR following Option A:

* Align the broker Service metadata label with the selector.
* Update the Kustomize controller label to match the existing Helm convention (`app.kubernetes.io/name: mcp-gateway-controller`).

That should preserve compatibility while keeping the labeling scheme consistent across installation methods.

Aditya7880900936:
Thanks for the detailed review @Patryk-Stefanski .

I followed Option A to preserve compatibility with existing Helm installations.

* Updated the broker Service metadata label to match the selector.
* Updated the Kustomize controller labels to align with the existing Helm convention (`app.kubernetes.io/name: mcp-gateway-controller`).

I've pushed the changes and would appreciate another look.

Patryk-Stefanski (commented):
Thanks for addressing the previous feedback — the metadata label and controller name fixes look correct.

One remaining issue that means the unification goal isn't fully achieved for production installs.

**`internal/controller/broker_router.go` — `brokerRouterLabels()` does not emit `app.kubernetes.io/component: broker`**

The Go controller's `brokerRouterLabels()` function only sets two labels on broker pods:
- `app.kubernetes.io/name: mcp-gateway`
- `app.kubernetes.io/managed-by: mcp-gateway-controller`

No `app.kubernetes.io/component` key. So when broker pods are deployed via `MCPGatewayExtension` (the operator path, which is the primary production deployment method), they will not have the `component: broker` label that this PR adds to the static Kustomize pod template.

This means:
- `kubectl logs -l app.kubernetes.io/component=broker` works after a Kustomize install, but returns **zero pods** on a controller-managed install
- On reconcile, `serviceNeedsUpdate` will overwrite the broker Service selector back to the `brokerRouterLabels()` set, **silently reverting the Service selector change** for any cluster running the operator

The fix needs to be in Go, not just YAML: add `app.kubernetes.io/component: broker` to `brokerRouterLabels()` in `internal/controller/broker_router.go`, and ensure `buildBrokerRouterService` selects on it.

Aditya7880900936:
Thanks for catching this @Patryk-Stefanski .

You're absolutely right—the previous changes only covered the static manifests, while controller-managed deployments were still missing the `app.kubernetes.io/component: broker` label.

I've updated `brokerRouterLabels()` to include the component label. Since `buildBrokerRouterService()` already uses `brokerRouterLabels()` for its selector, the Service selector is now includes the same label as well.

I've pushed the changes. Thanks again for the thorough review!

internal/controller/broker_router.go
```diff
@@ -58,11 +58,11 @@ var managedEnvVarNames = []string{
...
-		labelAppName:   brokerRouterName,
-		labelManagedBy: labelManagedByValue,
+		labelAppName:                  brokerRouterName,
+		labelManagedBy:                labelManagedByValue,
+		"app.kubernetes.io/component": "broker",
 	}
 }
-
 func (r *MCPGatewayExtensionReconciler) buildBrokerRouterDeployment(mcpExt *mcpv1alpha1.MCPGatewayExtension, publicHost, internalHost string) *appsv1.Deployment {
 	labels := brokerRouterLabels()
```
Patryk-Stefanski, line 93: `brokerRouterLabels()` now returns a third label (`app.kubernetes.io/component: broker`), but `deploymentNeedsUpdate` does not check pod template labels — so on an existing operator-managed cluster this reconcile loop runs without triggering a Deployment rollout:

1. `deploymentNeedsUpdate` returns false (only checks image, command, ports, mounts, env — not `Spec.Template.Labels`) → existing pods are not rolled, they keep their old two-label set with no `component: broker`
2. `serviceNeedsUpdate` returns true (selector changed) → Service selector is updated immediately to require `component: broker`
3. Existing pods no longer match the selector → kube-proxy removes them from Endpoints → **broker traffic drops to zero** until pods restart for another reason

Note: `deploymentNeedsUpdate` is a pre-existing function, not introduced by this PR. This PR is the first change to `brokerRouterLabels()` that only affects labels, which is what exposes the gap.

The fix is to add pod template label detection to `deploymentNeedsUpdate`:

```go
if !equality.Semantic.DeepEqual(desired.Spec.Template.Labels, existing.Spec.Template.Labels) {
    return true, fmt.Sprintf("pod template labels changed: %v -> %v",
        existing.Spec.Template.Labels, desired.Spec.Template.Labels)
}
```

And in the update block, apply the new labels before calling `r.Update`:

```go
existingDeployment.Spec.Template.Labels = deployment.Spec.Template.Labels
```

This ensures the Deployment rolls first (and the reconcile returns early at line 412), so the Service selector is only updated on the next reconcile once pods already carry the new label.

Aditya7880900936: Thanks for catching this edge case @Patryk-Stefanski .

I've updated the reconciliation logic so that pod template label changes are detected and applied during Deployment updates, ensuring a rollout occurs before the Service selector changes take effect. I also added a regression test covering pod template label changes.

I've pushed the changes—I'd appreciate another look. Thanks again for the thorough review!

Patryk-Stefanski:
Closing due to inactivity

Aditya7880900936:
Hi @Patryk-Stefanski, thanks for all the review feedback throughout this PR. I noticed it was closed due to inactivity, but I'd really appreciate if we could keep this contribution alive.

I addressed all the review comments, pushed the requested follow-up changes, updated the implementation based on your feedback, and rebased it with the latest main. This wasn't an abandoned PR—I simply missed following up after pushing the fixes.

I spent a considerable amount of time investigating the issue, ensuring the label changes were consistently applied across Kustomize, Helm, and the controller, as well as addressing the rollout/reconciliation concerns raised during review. I'd really like to see this work merged if the change is still relevant.

If anything else needs to be updated or rebased, I'm more than happy to do it immediately. If reopening isn't possible, I'm also happy to open a fresh PR with the latest changes. I'd sincerely appreciate another review rather than having this work discarded due to inactivity. Thank you!

Patryk-Stefanski:
@Aditya7880900936

Patryk-Stefanski:
@Aditya7880900936 I've reopened the pr please ensure all tests are passing. We will not be reviewing PRs that have conflicts and failing tests for 2weeks.

Aditya7880900936:
Hi @Patryk-Stefanski, thanks for reopening the PR. The GitHub Action appears to be closing it automatically because issue #759 is not yet assigned to me. Could you please assign the issue to me? I'll also make sure the PR is rebased on the latest main and all CI checks are passing before requesting another review.

Aditya7880900936:
Hi @Patryk-Stefanski , can you please reopen the PR so that i can give it a fix

internal/controller/broker_router.go
```diff
@@ -418,7 +418,12 @@ func (r *MCPGatewayExtensionReconciler) reconcileBrokerRouter(ctx context.Contex
 		existingContainer.VolumeMounts = mergeVolumeMounts(desiredContainer.VolumeMounts, existingContainer.VolumeMounts)
 		existingContainer.ReadinessProbe = desiredContainer.ReadinessProbe
 		existingDeployment.Spec.Template.Spec.Containers[0] = existingContainer
-		existingDeployment.Spec.Template.Spec.Volumes = mergeVolumes(deployment.Spec.Template.Spec.Volumes, existingDeployment.Spec.Template.Spec.Volumes)
+		existingDeployment.Spec.Template.Spec.Volumes = mergeVolumes(
+			deployment.Spec.Template.Spec.Volumes,
+			existingDeployment.Spec.Template.Spec.Volumes,
```
Patryk-Stefanski, line 424: Duplicate assignment — both lines set the same field to the same value. Remove one:

```go
existingDeployment.Spec.Template.Labels = deployment.Spec.Template.Labels
existingDeployment.Spec.Template.Labels = deployment.Spec.Template.Labels  // remove this line
```

internal/controller/broker_router.go
```diff
@@ -83,11 +83,11 @@ var logLevelFlagValues = map[mcpv1.LogLevel]string{
 
 func brokerRouterLabels() map[string]string {
 	return map[string]string{
-		labelAppName:   brokerRouterName,
-		labelManagedBy: labelManagedByValue,
+		labelAppName:                  brokerRouterName,
+		labelManagedBy:                labelManagedByValue,
+		"app.kubernetes.io/component": "broker",
 	}
```
Patryk-Stefanski, line 89: `brokerRouterLabels()` gains `app.kubernetes.io/component: broker` but no `app.kubernetes.io/name`. Operator-managed broker pods will be missing the `app.kubernetes.io/name` label that the Kustomize manifests set, so the unified `kubectl logs -l app.kubernetes.io/name=mcp-gateway` command will not select them.

Add the matching key:

```go
func brokerRouterLabels() map[string]string {
	return map[string]string{
		labelAppName:                  brokerRouterName,
		labelManagedBy:                labelManagedByValue,
		"app.kubernetes.io/name":      "mcp-gateway",
		"app.kubernetes.io/component": "broker",
	}
}
```

config/mcp-system/deployment-controller.yaml
```diff
@@ -6,6 +6,8 @@ metadata:
   labels:
     app: mcp-controller
     component: controller
+    app.kubernetes.io/name: mcp-gateway-controller
```
Patryk-Stefanski: The PR sets three different `app.kubernetes.io/name` values across installation paths:

| Path | Value |
|---|---|
| `deployment-broker.yaml` (Kustomize) | `mcp-gateway` |
| `deployment-controller.yaml` (Kustomize) | `mcp-gateway-controller` |
| `brokerRouterLabels()` (operator-managed broker) | _(missing — see other comment)_ |

The Kustomize controller manifest uses `mcp-gateway-controller`, which means `kubectl logs -l app.kubernetes.io/name=mcp-gateway` will still miss the controller pods. Either use a single shared name for all components and distinguish them via `app.kubernetes.io/component`, or document clearly that the label selector targets only the broker. The current state does not achieve the stated goal of a single command working across all components.

internal/controller/broker_router.go
Aditya7880900936: Good catch, thanks! I removed the duplicate assignment so the pod template labels are updated only once.

internal/controller/broker_router.go
Aditya7880900936: Thanks! I updated brokerRouterLabels() so the operator-managed broker resources use app.kubernetes.io/name: mcp-gateway, aligning them with the Kustomize-managed broker resources while continuing to distinguish components via app.kubernetes.io/component.

config/mcp-system/deployment-controller.yaml
Aditya7880900936: Thanks for pointing this out. I aligned the controller manifest to use app.kubernetes.io/name: mcp-gateway as well, while keeping app.kubernetes.io/component: controller for component-specific identification. This makes the app.kubernetes.io/name=mcp-gateway label selector consistent across the managed components.

david-martin:
Thanks for the work on this. I'm going to close the PR as it's been around a while without activity. Please reopen if you'd like to continue it.

PR #845, opened 2026-05-01
fix: run spell check on docs-only PRs

david-martin:
Split cspell into its own workflow so it triggers on all PRs, including docs-only changes.
The code-style workflow keeps its existing path filters and skips docs-only PRs as before.

This fixes an issue where docs only prs didn't have spelling run


Patryk-Stefanski (approved):

david-martin:
Thanks @Patryk-Stefanski

Issue #677, opened 2026-03-23
Env vars set via Broker/Route Deployment get reconciled back by MCP Controller


trepel:
**Summary**
Environment variables set via Broker/Route Deployment get reconciled back by MCP Controller. Some of the guides (e.g. https://github.com/Kuadrant/mcp-gateway/blob/main/docs/guides/authentication.md#step-2-configure-mcp-gateway-oauth-environment) expect such changes to be permanent.

**To Reproduce**
1. Have MCP Gateway (Broker/Router and Controller) up & running in a cluster
2. Patch Broker/Router Deployment CR to change environment variables:
```
kubectl set env deployment/mcp-gateway \
  OAUTH_RESOURCE_NAME="MCP Server" \
  OAUTH_RESOURCE="http://mcp.127-0-0-1.sslip.io:8001/mcp" \
  OAUTH_AUTHORIZATION_SERVERS="http://keycloak.127-0-0-1.sslip.io:8002/realms/mcp" \
  OAUTH_BEARER_METHODS_SUPPORTED="header" \
  OAUTH_SCOPES_SUPPORTED="basic,groups,roles,profile" \
  -n mcp-system
```
3. Observe that it gets reverted

Note: there's no revert if you scale the MCP Controller down before applying the patch

**Expected behavior**
Environment variable changes are permanent.

**Additional context**
I am not aware of any workarount.

PR #848, opened 2026-05-02
feat(broker): add /readyz endpoint and wire readinessProbe

### LINKED ISSUE #760: Add proper health and readiness endpoints to the broke
The broker (mcp-broker-router) does not implement /healthz or /readyz HTTP handlers on port 8080. Requests to these paths fall through to the / catch-all handler, which always returns 200 "ok" regardless of actual broker state.
                                                                                                                                                                                                           
  **Registered broker routes (port 8080):**                                                                                                                                                                    
  - / — catch-all, always returns "ok"
  - /mcp — Streamable HTTP MCP server                                                                                                                                                                      
  - /status — broker status (servers, tools, health counts)           
  - /status/ — broker status for a specific server                                                                                                                                                         
  - /.well-known/oauth-protected-resource — OAuth handler                                                                                                                                                  
                                                                                                                                                                                                           
  No /healthz or /readyz handlers exist.                                                                                                                                                                   
                                                                                                                                                                                                           
  **This creates several problems:**                                                                                                                                                                           
                                                                                                                                                                                                           
  1. Static deployment probes are no-ops. config/mcp-system/deployment-broker.yaml defines liveness (/healthz) and readiness (/readyz) probes on port 8080. These always pass due to the catch-all, so     
  Kubernetes never detects an unhealthy broker.                       
                                                                                                                                                                                                           
  2. Controller-created deployments have no probes. buildBrokerRouterDeployment() in internal/controller/broker_router.go does not set LivenessProbe or ReadinessProbe on the broker container. The        
  deployed pod has no health checking at all.
                                                                                                                                                                                                           
  3. Helm chart test references non-existent endpoint. charts/mcp-gateway/templates/tests/test-connection.yaml uses curl -f http://mcp-gateway:8080/healthz, which only passes due to the catch-all.       
   
  4. CI readiness check is fragile. .github/workflows/gevals.yaml polls http://localhost:8001/health through the gateway. Envoy returns 404 (no route for /health), but curl -s without -f treats any HTTP 
  response as success, so it passes by coincidence.

kkukadia:
Fixes #760

Adds an explicit `/readyz` handler in the broker and wires `ReadinessProbe`
on the controller-built Deployment, mirroring the managed-field pattern from
PR #677 (managed env vars).

## What this PR does

- New `/readyz` handler in `cmd/mcp-broker-router/main.go` returning 200.
- `ReadinessProbe` set on the broker Container in `buildBrokerRouterDeployment`
  (HTTPGet on named port `http`).
- Probe equality added to `deploymentNeedsUpdate` and assignment added to the
  merge block so probe drift is auto-reconciled.
- Tests for the Container shape and drift detection.

## Out of scope (per maintainer guidance in the issue thread)

- No `livenessProbe` — relying on the process running being sufficient.
- No catch-all narrowing — deferred to a separate issue.

## Summary by CodeRabbit

* **New Features**
  * Added readiness health check endpoint for broker-router service monitoring
  * Configured Kubernetes readiness probe to automatically detect service availability
  * Improved deployment reliability by preventing traffic routing to unhealthy broker-router instances

### DIFF
--- a/cmd/mcp-broker-router/main.go
+++ b/cmd/mcp-broker-router/main.go
@@ -321,6 +321,10 @@ func setUpHTTPServer(address string, mcpBroker broker.MCPBroker, sessionManager 
 		_, _ = fmt.Fprint(w, "Hello, World!  BTW, the MCP server is on /mcp")
 	})
 
+	mux.HandleFunc("/readyz", func(w http.ResponseWriter, _ *http.Request) {
+		w.WriteHeader(http.StatusOK)
+	})
+
 	oauthHandler := broker.ProtectedResourceHandler{Logger: logger}
 	mux.HandleFunc("/.well-known/oauth-protected-resource", oauthHandler.Handle)
 	mux.HandleFunc("/.well-known/oauth-protected-resource/", oauthHandler.Handle)
--- a/cmd/mcp-broker-router/main.go
+++ b/cmd/mcp-broker-router/main.go
@@ -322,6 +322,10 @@ func setUpHTTPServer(address string, mcpBroker broker.MCPBroker, sessionManager 
 	})
 
 	mux.HandleFunc("/readyz", func(w http.ResponseWriter, _ *http.Request) {
+		if !mcpBroker.IsReady() {
+			w.WriteHeader(http.StatusServiceUnavailable)
+			return
+		}
 		w.WriteHeader(http.StatusOK)
 	})
 
--- a/internal/broker/broker.go
+++ b/internal/broker/broker.go
@@ -40,6 +40,12 @@ type MCPBroker interface {
 	// ValidateAllServers performs comprehensive validation of all registered servers and returns status
 	ValidateAllServers() StatusResponse
 
+	// IsReady reports whether the broker can serve traffic.
+	// Returns true when no upstream servers are configured (empty tool list is valid),
+	// or when at least one configured upstream is healthy.
+	// Returns false only when servers are configured but none have connected yet.
+	IsReady() bool
+
 	// HandleStatusRequest handles HTTP status endpoint requests
 	HandleStatusRequest(w http.ResponseWriter, r *http.Request)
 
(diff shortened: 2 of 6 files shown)


david-martin:
Heads up — there was a duplicate PR for this (#854) which I've closed in favour of yours. One thing worth considering from that PR: making /readyz return 503 during startup until at least one upstream has connected, rather than always returning 200. That way the readiness probe actually gates traffic. @Patryk-Stefanski what do you think?

k1chik:
Thanks for the heads up on #854 and flagging the startup-aware readiness logic, that makes sense. Happy to incorporate whatever direction you and @Patryk-Stefanski land on.                                                                              

One more thing, when I was tracing how `ValidateAllServers()` works, I noticed it holds a read lock and  then calls `RegisteredMCPServers()` which tries to grab the same read lock again. From what I can tell the Go [docs](https://pkg.go.dev/sync#RWMutex) say this can deadlock if a writer sneaks in between the two (recursive read-locking is prohibited). Probably been fine so far since `/status` is only hit by humans, but with the `readyz` probe firing every `10s` automatically I wondered if this becomes more of a concern during config reloads.

Could be I'm missing something, worth a look or should I open a separate issue?

kkukadia:
@Patryk-Stefanski, when you get a chance, could you please share your feedback?

Patryk-Stefanski:
+1 to returning 503 until at least one upstream is connected. 

Good catch on the deadlock. Please open up a separate PR for that issue.

kkukadia:
Thank you! I'm on it

kkukadia:
The GEVALS failure is actually a pre-existing issue tracked in #868, not caused by this PR.
The readiness wait in the workflow polls `/health` (Envoy's liveness endpoint), which goes green as soon as the proxy starts, has no idea whether the broker has connected to any upstream yet. Before this PR, the race was occasionally lost and the eval saw no tools. Now with the 503 logic, `/readyz` actively signals not-ready during that window, so the race is lost consistently and the failure shows up every time.

PR #921 passes for the same reason. It doesn't change `/readyz`, so the endpoint still returns 200 via the catch-all and the pod looks ready regardless of upstream state. The underlying fix is on the workflow side (polling `/readyz` instead of `/health`), which sounds like what's being tracked in #868.

david-martin (approved):
Thanks @k1chik 
I'm going to merge this despite the failing CI (tracked separately).

kkukadia:
@david-martin Thank you so much!

Issue #628, opened 2026-03-03
Feature Request: Horizontal Pod Autoscaler (HPA) Support


arielharush96:
**Is your feature request related to a problem? Please describe.**

The mcp-gateway currently runs as a single replica with no horizontal scaling support. As the number of concurrent MCP sessions might grows, a single pod could becomes a bottleneck and a single point of failure.

**Describe the solution you'd like**

Add HPA support for deployment so it can scale horizontally based on load.

**Describe alternatives you've considered**

- Manual scaling (manually adjusting replica count)- requires operator intervention and doesn't respond to load changes automatically.
- Vertical scaling (increasing resource limits on a single pod) -might has a hard ceiling per node and remains a single point of failure.

**Additional context**

This came up during a performance review meeting. The next research phase will focus on high concurrent load and a large number of registered MCP servers, so the need for horizontal scaling was agreed upon by the team.

david-martin:
Moving this to 0.7, with the initial scaling guide added in 0.6 in #635 and available at https://github.com/Kuadrant/mcp-gateway/blob/main/docs/guides/scaling.md

PR #854, opened 2026-05-02
add /healthz and /readyz endpoints to broker-router

Aman-Cool:
Both endpoints were referenced in `config/mcp-system/deployment-broker.yaml` and the Helm chart smoke test (`test-connection.yaml`) but never actually registered on the HTTP server. Any `GET /healthz` or `GET /readyz` just fell through to the `"/"` catch-all and returned `200` unconditionally; meaning the readiness probe was always passing regardless of whether a single upstream MCP server had connected.

This matters most during startup. The broker's first connection attempt fires on the one-minute ticker (`DefaultTickerInterval`). Without a real readiness gate, Kubernetes marks the pod `Ready` immediately and starts routing traffic before any upstream is available. Every `tools/call` during that window fails. This makes HPA scale-out (#628) unreliable; new pods advertise themselves as ready ~60 seconds before they actually are.

**What this adds:**

- `/healthz` : simple liveness, always `200` once the HTTP server is up
- `/readyz` : `503` when upstream servers are configured but none are healthy yet; `200` otherwise (including when no servers are configured at all, since the broker can still serve `initialize` and return an empty tool list)
- `LivenessProbe` and `ReadinessProbe` added to `buildBrokerRouterDeployment` in the operator — operator-deployed pods were missing probes entirely
- `deploymentNeedsUpdate` now checks probes so existing deployments get updated on next reconcile

**Files changed:**
- `cmd/mcp-broker-router/main.go` : register the handlers in `setUpHTTPServer`
- `internal/controller/broker_router.go` : add probes to deployment spec and update detection
- `cmd/mcp-broker-router/main_test.go` : table tests for `readyzHandler` (no servers / none healthy / partial / all healthy)
- `internal/controller/deployment_test.go` : assert probe spec is set correctly and that probe changes trigger a deployment update

```
make test-unit  # all pass
```

### DIFF
--- a/cmd/mcp-broker-router/main.go
+++ b/cmd/mcp-broker-router/main.go
@@ -348,11 +348,28 @@ func setUpHTTPServer(address string, mcpBroker broker.MCPBroker, sessionManager 
 		w.WriteHeader(http.StatusOK)
 	})
 
+	mux.HandleFunc("/healthz", func(w http.ResponseWriter, _ *http.Request) {
+		w.WriteHeader(http.StatusOK)
+	})
+	mux.Handle("/readyz", readyzHandler(mcpBroker))
 	mux.HandleFunc("/status", mcpBroker.HandleStatusRequest)
 	mux.HandleFunc("/status/", mcpBroker.HandleStatusRequest)
 	mux.Handle("/mcp", streamableHTTPServer)
 
 	return httpSrv, streamableHTTPServer
+}
+
+// readyzHandler returns 503 when upstream MCP servers are configured but none are healthy,
+// blocking traffic during the startup window before connections are established.
+func readyzHandler(b broker.MCPBroker) http.HandlerFunc {
+	return func(w http.ResponseWriter, _ *http.Request) {
+		s := b.ValidateAllServers()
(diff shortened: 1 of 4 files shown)


Aman-Cool:
Hey @david-martin and @maleck13, noticed that `/healthz` and `/readyz` were already wired up in the deployment manifest and the Helm smoke test, but the HTTP server never actually registered them; both paths just fell through to the catch-all and returned `200` regardless of broker state.

The readiness one was the more annoying gap: during startup the pod would go `Ready` immediately, before any upstream connection had been attempted, so traffic would hit it and fail for the first ~60 seconds. Wanted to fix that before looking at HPA support (#628) since you can't really scale out safely without a meaningful readiness gate.

Kept the logic simple ;`/readyz` only blocks when servers are configured but none have connected yet. If there are no servers configured at all it stays `200`, since the broker can still handle `initialize` requests fine.

david-martin:
Thanks for this, but k1chik already has a PR open for the same feature (#848) that came in earlier. Going to close this in favour of that one to avoid duplicate work. Appreciate the thorough implementation though — the startup-aware readiness logic is a good idea and I've flagged it on #848.

PR #858, opened 2026-05-03
Makefile: print MCP Gateway URL after local-env-setup

### LINKED ISSUE #401: "make local-env-setup" SHOULD print the URL of the MCP server to stdout
Because the port environment variables had changed, I spent an hour troubleshooting why the MCP server wasn't where I expected it.

It would simplify the user experience if `make local-env-setup` output something like

```
The MCP Gateway is available at http://mcp.127-0-0-1.sslip.io:8090/mcp
```

shiavm006:
Fixes #401

Adds the MCP Gateway URL to the trailing output of `make local-env-setup`and `make local-env-setup-olm` so users know where to point their MCP client without grepping the Makefile or running `make urls` separately.

## Summary by CodeRabbit

* **Chores**
  * Consolidated local environment setup completion message output for consistent display across setup targets.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -603,7 +603,7 @@ local-env-setup: setup-cluster-base ## Setup complete local demo environment wit
 	# Deploy everything server for local dev (use 'make deploy-test-servers' for all servers)
 	"$(MAKE)" deploy-everything-server
 	"$(MAKE)" deploy-example-minimal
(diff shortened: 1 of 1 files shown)


shiavm006:
cc @david-martin  @maleck13

maleck13 (approved):

Issue #817, opened 2026-04-29
Implement MCPServerRegistration enabled/disabled support


maleck13:
## Summary

The `Enabled` field exists on the internal `config.MCPServer` type but is never used to gate any behavior. An MCPServerRegistration currently cannot be disabled without deleting it.

## Current State

- `internal/config/types.go:66` — `MCPServer` struct has `Enabled bool` field
- `internal/controller/mcpserverregistration_controller.go:386` — always hardcodes `Enabled: true`
- `cmd/mcp-broker-router/main.go:394` — only logged for debug
- `internal/broker/upstream/mcp.go:49` — copied in `GetConfig()` but never checked
- The MCPServerRegistration CRD (`api/v1alpha1/types.go`) has **no** `enabled` field in its spec

## Proposal

1. Add an optional `enabled` field to the `MCPServerRegistrationSpec` CRD (defaulting to `true`)
2. Have the controller read this field and set it on the config accordingly
3. Have the broker skip connecting to servers where `Enabled == false`
4. Have the router skip routing to disabled servers

This would allow operators to temporarily disable an upstream MCP server without removing the registration — useful for maintenance, incident response, or staged rollouts.

## Acceptance Criteria

- [ ] `MCPServerRegistrationSpec` has an optional `enabled` field (default `true`)
- [ ] Controller sets `Enabled` from CRD spec
- [ ] Broker skips disabled servers (no connection, tools removed)
- [ ] Router rejects tool calls for disabled servers
- [ ] E2E test: disable a registration, verify tools are unavailable, re-enable, verify tools return
- [ ] API reference docs updated

david-martin:
Consider using an enum (e.g. `Enabled`, `Disabled`) rather than a boolean for the `enabled` field, to allow for future extensibility (e.g. `Draining`, `Maintenance`). See Kubernetes API conventions on primitive types: https://github.com/kubernetes/community/blob/main/contributors/devel/sig-architecture/api-conventions.md#primitive-types

malladinagarjuna2:
hey @david-martin  can i work on this issue ?

david-martin:
Hi @malladinagarjuna2, go for it!
No formal assignment needed.
Open a PR when you have something ready.

Note the comment above about using an enum rather than a boolean for the field.

malladinagarjuna2:
ok sure sir!

malladinagarjuna2:
[...]

PR #861, opened 2026-05-03
Add State-based enable/disable support for MCPServerRegistration

Aryanburnwal05:
This PR adds support for enabling and disabling MCP servers via a new `state` field in the MCPServerRegistration CRD.

Instead of a boolean, a string enum (`Enabled` / `Disabled`) is used to keep the API extensible and aligned with Kubernetes conventions.

Key changes :-

1. Added `state` field with default `Enabled`
2. Propagated state from CRD to internal config
3. Updated ConfigChanged to detect state transitions
4. Broker now skips disabled servers and removes their tools
5. Re-enables reconnect and restore tools automatically
6. Improved status: Ready=False, Reason=Disabled
7. Optimized manager loop to avoid repeated work when disabled

Tests cover :-

1. Enabled → Disabled
2. Disabled → Enabled
3. Initially Disabled

Note:- The previous `Enabled` field was unused. The new `State` field is now the single source of truth for lifecycle management.

### DIFF
--- a/api/v1alpha1/types.go
+++ b/api/v1alpha1/types.go
@@ -4,6 +4,11 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+const (
+	MCPServerStateEnabled  string = "Enabled"
+	MCPServerStateDisabled string = "Disabled"
+)
+
 // +kubebuilder:object:root=true
 // +kubebuilder:subresource:status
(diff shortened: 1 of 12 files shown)


david-martin:
This is a duplicate of work already in progress on #817. Closing in favour of that. Thanks for the interest though — check for existing PRs and assigned issues before picking something up.

PR #872, opened 2026-05-04
Fix: concurrent map race in session cache

Aman-Cool:
The in-memory session cache was storing a raw `map[string]string` in a `sync.Map` and returning the live reference from `GetSession`. Both `AddSession` and `RemoveServerSession` would then mutate that live map directly, with no lock around it.

This means if two tool calls hit the router at the same time, targeting different backend servers on the same gateway session ; both goroutines end up writing to the exact same map concurrently. Go's runtime detects this and panics with `fatal error: concurrent map writes`, taking down the entire `ext_proc` router process.

The fix is straightforward: serialize mutations with a mutex and use copy-on-write. Instead of mutating the stored map in place, `AddSession` and `RemoveServerSession` now copy the current entries into a fresh map, apply the change, and swap it in. Readers always get a stable snapshot. Also fixed a small pre-existing issue where `RemoveServerSession` on a non-existent key would accidentally write an empty map entry.

Added two concurrent tests (50 goroutines, race detector enabled) that would have panicked on the old code.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved session cache synchronization to ensure safe, reliable behavior during concurrent operations without changing public APIs.

* **Tests**
  * Added concurrency tests covering session add/remove/delete flows to validate thread-safety and prevent race conditions.

### DIFF
--- a/internal/session/cache_test.go
+++ b/internal/session/cache_test.go
@@ -6,6 +6,7 @@ import (
 	"sync"
 	"testing"
 
+	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
 
@@ -278,7 +279,7 @@ func TestInMemoryCache_AddSession_Concurrent(t *testing.T) {
 				fmt.Sprintf("server%d", i),
 				fmt.Sprintf("upstream-session-%d", i),
 			)
-			require.NoError(t, addErr)
+			assert.NoError(t, addErr)
 		}(i)
 	}
 	wg.Wait()
@@ -307,7 +308,7 @@ func TestInMemoryCache_RemoveServerSession_Concurrent(t *testing.T) {
 	for i := range n {
 		go func(i int) {
 			defer wg.Done()
-			require.NoError(t, cache.RemoveServerSession(ctx, "gateway-session-1",
+			assert.NoError(t, cache.RemoveServerSession(ctx, "gateway-session-1",
 				fmt.Sprintf("server%d", i),
 			))
 		}(i)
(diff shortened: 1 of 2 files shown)


Aman-Cool:
Hey @david-martin, found this while doing a deep dive into the codebase. 

The in-memory session cache returns the live `map[string]string` reference from `GetSession`, and both `AddSession` and `RemoveServerSession` mutate it directly with no lock. When a client fires tool calls to two different backends at the same time (which is pretty normal with a federated gateway), both goroutines end up writing to the same map concurrently and the router just panics and dies.

Fixed it with copy-on-write under a mutex; mutations always work on a fresh copy, so readers never touch anything that's being written to. Also threw in two concurrent tests with the race detector that would've caught this from the start.

Happy to adjust anything if needed...

maleck13:
@Aman-Cool thanks for this contribution. The fix here looks good.  The race is valid. We may need to add the lock on DeleteSessions also. A write could re-create without the lock

Aman-Cool:
@maleck13 Good catch😄, `DeleteSessions` wasn't holding `innerMu`, so a concurrent `AddSession` could load the old map, the delete would fire, and then `AddSession` would store its copy back undoing the deletion. 
Added the lock to serialize it with the rest of the copy-on-write mutations, and threw in a concurrent test that hammers `AddSession` and `DeleteSessions` on the same key simultaneously so the race detector would catch a regression.

internal/session/cache_test.go
```diff
@@ -257,6 +259,94 @@ func TestInMemoryCache_SetAndGetClientElicitation(t *testing.T) {
...
+	var wg sync.WaitGroup
+	wg.Add(n)
+	for i := range n {
+		go func(i int) {
+			defer wg.Done()
+			_, addErr := cache.AddSession(ctx, "gateway-session-1",
+				fmt.Sprintf("server%d", i),
+				fmt.Sprintf("upstream-session-%d", i),
+			)
+			require.NoError(t, addErr)
```
maleck13: may be better to use assert rather than require. Assert will mark the test as failed rather than aborting. 

```
Package require implements the same assertions as the `assert` package but stops test execution when a test fails.
```

maleck13 (commented):
If we can change to using assert. Happy to merge at that point 👍

Aman-Cool:
@maleck13 Good point, swapped all `require` calls inside goroutines to `assert` across the three concurrent tests. `require.FailNow` isn't goroutine-safe; it panics when called from a non-test goroutine. `assert` just marks the test failed and lets the goroutine finish cleanly.

maleck13 (approved):

PR #876, opened 2026-05-04
fix: make test server port dynamic in broker_test.go

### LINKED ISSUE #875: TODO : Make test server port dynamic in broker_test.go
## Problem

`internal/broker/broker_test.go` hardcodes port `8088` for the test MCP server:

```go
const (
    // MCPPort is the port the test server should listen on (TODO make dynamic?)
    MCPPort = "8088"
    MCPAddr = "http://localhost:8088/mcp"
    MCPAddrForgetAddr = "http://localhost:8088/admin/forget"
)
```

This can cause test failures if port `8088` is already in use on the host machine. There is already a TODO for it 

## Proposed Fix

Replace the hardcoded constants with vars set dynamically in `TestMain` using `net.Listen(":0")`:

```go
var (
    MCPAddr           string
    MCPAddrForgetAddr string
)

func TestMain(m *testing.M) {
    ln, err := net.Listen("tcp", "localhost:0")
    if err != nil {
        fmt.Fprintf(os.Stderr, "Failed to find free port: %v\n", err)
        os.Exit(1)
    }
    port := fmt.Sprintf("%d", ln.Addr().(*net.TCPAddr).Port)
    ln.Close() // release so RunServer can bind to it

    MCPAddr = fmt.Sprintf("http://localhost:%s/mcp", port)
    MCPAddrForgetAddr = fmt.Sprintf("http://localhost:%s/admin/forget", port)

    startFunc, shutdownFunc, err := server2.RunServer("http", port)
   
}
```

**Note:** There is a small TOCTOU race between `ln.Close()` and `RunServer` binding the port. The truly race-free solution would be modifying `RunServer` in `server2.go` to accept a `net.Listener` directly but that is a larger change. would prefer maintainer suggestion .

there are several todos across the code base should one? 

1. Open an issue first and wait for acknowledgment before submitting a PR, or
2. Submit a PR directly with the proposed solution in the description?

Rama542:
## What does this PR do?

The test server in `broker_test.go` was using a hardcoded static port number. This can cause tests to fail with "address already in use" errors in CI environments or when multiple test runs happen in parallel on the same machine.

## Changes made

- Replaced the hardcoded port in `broker_test.go` with `:0` so the OS assigns a random free port at runtime
- Retrieved the actual assigned port from the listener using `Addr()` after startup
- Updated all test client references to use the dynamically assigned address

## Why this matters

- Eliminates flaky test failures caused by port conflicts
- Makes the test suite safe to run in parallel and in shared CI environments
- Follows standard Go testing practices for ephemeral test servers

## Testing

Ran `go test ./...` locally all tests pass.

Closes #875

## Summary by CodeRabbit

* **Tests**
  * Improved test infrastructure to start servers on dynamically allocated ports via caller-provided listeners, avoiding bind races and hardcoded addresses.
  * Centralized server startup helpers to reliably create and wire HTTP server streams while preserving admin endpoints and startup/shutdown behavior.
  * Relaxed a span assertion in a router test to be less strict about end-time, reducing flakiness.

### DIFF
--- a/internal/broker/broker_test.go
+++ b/internal/broker/broker_test.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"log/slog"
+	"net"
 	"net/http"
 	"os"
 	"slices"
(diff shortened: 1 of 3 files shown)


aniketpandey05:
Hi @Rama542, I was already working on this issue and was waiting for maintainer input on the proposed approach before opening a PR.

Abdul-Moiz31:
Hi @Rama542 
 The **RunServerWithListener** approach correctly eliminates the TOCTOU race passing the open listener directly into **http.Server.Serve(ln)** keeps the port reserved with no gap between allocation and binding. That part looks solid.

**One thing worth checking:** 
The **RunServerWithListener** function duplicates the tool registration and server setup logic that already exists in **RunServer**. If **testTools** or the mux routes ever get updated in **RunServer**, **RunServerWithListener** would silently drift out of sync. Consider extracting the shared setup into a private **buildMCPServer()** helper that both functions call, so there's a single source of truth.

Do let me know if i can help in this 

Thankyou !!

Abdul-Moiz31:
@Rama542 i can help with this if you're facing any issue ? o i can also take this issue to sort out . 

Thee is no need to force push i guess . 

thanks do let me know

david-martin:

jasonmadigan (changes requested):
needs a rebase -- merge conflicts with main.

Patryk-Stefanski:
Closing due to inactivity, and the related issue is not triaged, please see [contributing.md](https://github.com/Kuadrant/mcp-gateway/blob/main/CONTRIBUTING.md)

PR #879, opened 2026-05-05
fix: use custom programmed condition for HTTPRoute

### LINKED ISSUE #878: Infinite Reconciliation Loop (Split-Brain) via HTTPRoute "Programmed" Condition Overwrite
The `updateHTTPRouteStatus` function in the mcp-gateway controller blindly mutates the `Programmed` condition of the Gateway API `HTTPRoute` parent status.

In Gateway API, `Programmed` is a standard condition strictly managed by the underlying Gateway controller (e.g., Istio) for the `RouteParentStatus` it owns. When the MCP controller overwrites this condition's `Reason` to `InUseByMCPServerRegistration`, the Gateway controller detects the mutation and reverts it. Kuadrant then detects this change and overwrites it again, resulting in an infinite reconciliation loop (split-brain), leading to 100% CPU usage and etcd bloat.

**Fix:**
The Kuadrant controller should use a custom condition type such as `mcp.kuadrant.io/Programmed` or `PolicyAccepted` instead of modifying standard Gateway API conditions.

pulkitvats2007-crypto:
## Title
Fix infinite reconciliation loop caused by HTTPRoute `Programmed` condition overwrite

## Summary
This PR fixes a critical split-brain issue between the Kuadrant MCP Gateway controller and Gateway API controllers (e.g., Istio), which caused an infinite reconciliation loop and high CPU usage.

## Problem
The controller was overwriting the standard `Programmed` condition on `HTTPRoute.Status.Parents`, a field owned by the Gateway controller.  
This resulted in both controllers continuously reverting each other's changes, causing:
- Infinite reconciliation loop
- High CPU usage
- Excessive API server and etcd load
- Potential control plane instability

Additionally, `Status().Update()` was called inside a loop, leading to multiple unnecessary API calls.

## Fix
- Replaced mutation of the standard `Programmed` condition with a custom condition:
  - `mcp.kuadrant.io/Programmed`
- Ensured Gateway-owned conditions remain untouched
- Moved `r.Status().Update()` outside the loop to perform a single update per reconciliation

## Tests
- Existing unit and integration tests pass without modification
- Added new test:
  - `TestUpdateHTTPRouteStatus`
  - Verifies:
    - Original `Programmed` condition is preserved
    - Custom condition is correctly added

## Impact
- Eliminates infinite reconciliation loop
- Prevents controller contention over status fields
- Reduces API server load and avoids etcd bloat
- Aligns with Gateway API contract and best practices
- Improves overall cluster stability

## Notes
This change ensures the MCP controller only manages its own status conditions and does not interfere with Gateway controller-owned fields.

Fixes #878 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved HTTP route status condition handling and consolidation for better reliability.

* **Tests**
  * Added comprehensive test coverage for HTTP route status updates.

### DIFF
--- a/internal/controller/mcpserverregistration_controller.go
+++ b/internal/controller/mcpserverregistration_controller.go
@@ -552,7 +552,7 @@ func (r *MCPReconciler) updateHTTPRouteStatus(ctx context.Context, mcpsr *mcpv1a
 	}
 
 	condition := metav1.Condition{
-		Type:               "Programmed",
+		Type:               "mcp.kuadrant.io/Programmed",
 		ObservedGeneration: httpRoute.Generation,
 		LastTransitionTime: metav1.Now(),
 	}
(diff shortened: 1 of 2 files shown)


david-martin:

Patryk-Stefanski:
Thanks @pulkitvats2007-crypto. The infinite reconciliation issue is worth investigating. However, be aware that PR #979 (fire-and-forget controller rewrite) also modifies the HTTPRoute status handling and will conflict with this PR. It may make sense to coordinate with the #979 author (@malladinagarjuna2) to avoid duplicate work.

Also, issue #878 is self-filed — per [CONTRIBUTING.md](https://github.com/Kuadrant/mcp-gateway/blob/main/CONTRIBUTING.md), it would help to have maintainer confirmation that the infinite reconcile loop is reproducible before the fix merges.

Patryk-Stefanski (approved):
The infinite reconciliation loop bug is real. The current code sets the standard Gateway API `Programmed` condition (line 623 in mcpserverregistration_controller.go), which Istio's gateway controller also manages — each controller sees the other's changes and re-reconciles.

The fix is correct:
- Changes condition type to `mcp.kuadrant.io/Programmed` (custom namespaced, no conflict with Istio)
- Moves `r.Status().Update()` outside the loop (single update vs N updates per reconciliation)
- Test verifies custom condition is set while original `Programmed` is untouched

Re: PR #979 conflict — on closer inspection, #979 only moves *where* `updateHTTPRouteStatus()` is called, not what it does internally. This fix is orthogonal and can merge independently.

Note: issue #878 is self-filed. Maintainers should confirm the loop is reproducible in their environment before merging, but the code change is clearly correct regardless.

Patryk-Stefanski:
Closing due to inactivity, and the related issue is not triaged. Please see [contributing.md](https://github.com/Kuadrant/mcp-gateway/blob/main/CONTRIBUTING.md)

PR #880, opened 2026-05-05
feat: add local A2A agent fixture for design validation

mugiwaraluffy56:
part of the A2A protocol support investigation under #766.

before any gateway routing work can happen for A2A, the design discussions need concrete request/response examples to stay grounded. this adds a minimal local A2A agent under tests/servers/a2a-agent that exercises the core protocol paths.

the fixture serves two skills (echo and reverse), handles both non-streaming tasks/send and streaming tasks/sendSubscribe with SSE events, and supports tasks/get and tasks/cancel. it also has a stub for tasks/pushNotificationConfig/set since that endpoint will be relevant for the push notification design work. no gateway code is changed.

it is built with only the standard library so there is no new dependency on the main module. the agent card is at /.well-known/agent.json with two stable skills and deterministic responses, which makes it usable as a reference for the config model, discovery, and streaming design issues.

the README has curl examples for all supported methods with expected output.

all tests pass.

## Summary by CodeRabbit

* **Tests**
  * Added A2A fixture agent implementation with comprehensive test coverage for message processing behaviors, task storage operations, and endpoint validation.

* **Documentation**
  * Added A2A fixture agent documentation detailing HTTP endpoints, supported JSON-RPC methods, and curl examples.

* **Chores**
  * Added Docker containerization and Go module configuration for A2A fixture agent.

### DIFF
--- a/tests/servers/a2a-agent/Dockerfile
+++ b/tests/servers/a2a-agent/Dockerfile
@@ -1,0 +1,13 @@
+FROM golang:1.25-alpine AS builder
+
+WORKDIR /app
+
+COPY go.mod go.sum ./
+RUN go mod download
+
+COPY *.go ./
+RUN CGO_ENABLED=0 GOOS=linux go build -o /a2a-fixture-agent
+
+FROM alpine:3.21
+COPY --from=builder /a2a-fixture-agent /a2a-fixture-agent
+CMD ["/a2a-fixture-agent"]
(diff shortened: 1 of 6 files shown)


mugiwaraluffy56:
closing saw that #766 is scoped for an LFX Term 02 mentee, so holding off on any A2A work until that's settled.

PR #884, opened 2026-05-06
add latest lib for mcp client that fixes #813

### LINKED ISSUE #813: Re-initialize mcp sessions on failure

maleck13:
This should fix the constant logging from #813 The session will be re-init once the timer ticks as the manager will do a ping and it will fail. Re-Connection is tested in the e2e test It("[Full] should gracefully handle an MCP Server becoming unavailable")

## Summary by CodeRabbit

* **Chores**
  * Updated MCP Go framework and various underlying dependencies to the latest versions for improved stability and compatibility.


maleck13:
@jasonmadigan @david-martin this resolves the constant logging you see when a backend MCP server becomes unavailable

david-martin (approved):
Tests look good.
Assuming this is the mcp-go fix? https://github.com/mark3labs/mcp-go/pull/839

PR #886, opened 2026-05-06
broker: guard gatewayServer nil dereference in MCPManager

### LINKED ISSUE #882: broker: nil pointer panic in findToolConflicts causes CrashLoopBackOff on upstream connect
When an upstream MCP server connects and the broker attempts to check for tool conflicts, the broker panics with a nil pointer dereference in `findToolConflicts`.

The crash happens at `man.gatewayServer.ListTools()` in `internal/broker/upstream/manager.go`. This takes down the entire broker process, putting it into CrashLoopBackOff.

Observed with the everything-server on a fresh local-env-setup cluster. The panic stack:

Image

The fix is to guard `findToolConflicts` with a nil check on `man.gatewayServer` before calling `ListTools()`.

mugiwaraluffy56:
`findToolConflicts`, `removeAllTools`, and the tool sync loop in `manage` all call methods on `man.gatewayServer` without checking for nil first. When `NewUpstreamMCPManager` is called with a nil gateway (which happens in test setup and can happen in pathological broker startup sequences), any upstream connect that triggers tool discovery panics and takes the broker into CrashLoopBackOff.

The fix adds nil guards at each of the four `gatewayServer` call sites. When the gateway is nil, `findToolConflicts` returns no conflict error and the add/delete calls are skipped.

Adds a regression test that calls `manage` with a nil gateway and asserts no panic.

Fixes #882

## Summary by CodeRabbit

* **Bug Fixes**
  * Manager creation now fails when a gateway/server is missing; the system logs and skips misconfigured upstreams to prevent unstable startups and clearer errors.

* **Tests**
  * Tests updated to assert the new error behavior and include a lightweight mock gateway; added coverage for the missing-gateway error case.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Kuadrant/mcp-gateway/pull/886)


internal/broker/upstream/manager.go
```diff
@@ -312,6 +312,9 @@ func (man *MCPManager) setStatus(err error, toolCount int, invalidTools []Invali
 }
 
 func (man *MCPManager) findToolConflicts(mcpTools []server.ServerTool) error {
+	if man.gatewayServer == nil {
+		return nil
```
maleck13: This is an error situation.  I think we should change the New function to return an error if any needed pointer is  nil

maleck13 (changes requested):
Lets change the new function to return an error if the gatewayserver is nil

internal/broker/upstream/manager.go
mugiwaraluffy56: done, findToolConflicts now returns an error when gatewayServer is nil instead of silently returning

mugiwaraluffy56:
> Lets change the new function to return an error if the gatewayserver is nil

updated, findToolConflicts now returns an error when gatewayServer is nil rather than silently returning nil

maleck13:
Sorry. A bit of confusion. I meant the `NewUpstreamMCPManager` could return an error then we don't need all the checks in different places

mugiwaraluffy56:
done, moved the nil check to NewUpstreamMCPManager so it returns an error on construction rather than having guards scattered through the methods

internal/broker/upstream/manager.go
```diff
@@ -102,7 +102,10 @@ const DefaultTickerInterval = time.Minute * 1
 // NewUpstreamMCPManager creates a new MCPManager for managing a single upstream MCP server.
 // The addTools and removeTools callbacks are used to update the gateway's tool registry.
 // The tickerInterval controls how often the manager checks backend health (use 0 for default).
-func NewUpstreamMCPManager(upstream MCP, gatewaySever ToolsAdderDeleter, logger *slog.Logger, tickerInterval time.Duration, policy mcpv1alpha1.InvalidToolPolicy) *MCPManager {
```
maleck13, line 105: pre-existing but I notice the param is in-correctly named. gatewaySever -> gatewayServer would be nice to fix as we are changing stuff around here anyway

maleck13 (commented):
changes look fine but code style is failing currently

mugiwaraluffy56:
fixed the param name and the style check

maleck13:
needs a rebase

mugiwaraluffy56:
rebased, the cspell failures are pre existing in docs/design/gateway-url-token-elicitation/ and not related to this pr

maleck13 (approved):

david-martin:
Merging these changes.
Failing CI is cspell (low priority, can be fixed after).

Thanks @mugiwaraluffy56 for the change, and @maleck13 for the review.

PR #887, opened 2026-05-06
feat: add support for productized MCP Gateway installation

trepel:
### Summary
Add support for installing MCP Gateway from the official Red Hat catalog instead of custom CatalogSource.

### Changes
- Add INSTALL_PRODUCTIZED_VERSION flag to control installation source
-  Add CHANNEL variable to configure OLM subscription channel (default: preview)
-  Skip CatalogSource creation when using productized version
-  Dynamically patch subscription channel and source based on configuration
-  Change INSTALL_RHCL default to false (RHCL is an OLM dependency)
- - but create Kuadrant CR if it does not exist already  
-  Quote shell variables throughout script for safety

### Verification
You can validate this by running against OpenShift using various values for env vars, example:
```
MCP_GATEWAY_VERSION=0.6.0 INSTALL_PRODUCTIZED_VERSION=true CHANNEL=preview INSTALL_RHCL=false INSTALL_SERVICE_MESH=true USE_OCP_INGRESS=true ./config/openshift/deploy_openshift.sh
```

## Summary by CodeRabbit

* **New Features**
  * Adds automatic Kuadrant instance creation during deployment.

* **Chores**
  * Deployment defaults updated to disable MCP and Connectivity Link installation by default and add new configuration options.
  * Connectivity Link installation now runs only when explicitly enabled.
  * MCP Gateway deployment made more configurable with clearer, explicit parameters and conditional catalog/source handling.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Kuadrant/mcp-gateway/pull/887)

### DIFF
--- a/config/openshift/deploy_openshift.sh
+++ b/config/openshift/deploy_openshift.sh
@@ -6,9 +6,14 @@ MCP_GATEWAY_VERSION="${MCP_GATEWAY_VERSION:-local}"
 MCP_GATEWAY_HOST="${MCP_GATEWAY_HOST:-mcp.apps.$(oc get dns cluster -o jsonpath='{.spec.baseDomain}')}"
 MCP_GATEWAY_NAMESPACE="${MCP_GATEWAY_NAMESPACE:-mcp-system}"
 GATEWAY_NAMESPACE="${GATEWAY_NAMESPACE:-gateway-system}"
-INSTALL_RHCL="${INSTALL_RHCL:-true}"
+INSTALL_RHCL="${INSTALL_RHCL:-false}"
 INSTALL_SERVICE_MESH="${INSTALL_SERVICE_MESH:-true}"
 USE_OCP_INGRESS="${USE_OCP_INGRESS:-true}"
+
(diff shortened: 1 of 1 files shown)


config/openshift/deploy_openshift.sh
```diff
@@ -103,6 +116,15 @@ until oc get csv -n $MCP_GATEWAY_NAMESPACE -l operators.coreos.com/mcp-gateway.$
 done
 echo "MCP Gateway Controller installed via OLM"
 
+# Create RHCL Instance if not already exists
```
maleck13, line 122: might be worth waiting for the csv to be established before creating the kuadrant resources

```
echo "Waiting for Kuadrant CRD to be established..."
  until oc wait crd/kuadrants.kuadrant.io --for condition=established &>/dev/null; do sleep 5; done
```

trepel: added by force-push

maleck13 (approved):
Changes look ok here. Didn't try this one locally as the change is OCP specific

PR #888, opened 2026-05-06
use maps.Clone in copy-on-write mutations

Aman-Cool:
## Summary

My earlier PR #872 fixed the concurrent map write panic using `sync.Map` + copy-on-write and has since been merged into main. This PR builds on top of that by applying the `maps.Clone` suggestion from the review:

- `AddSession`: replaces `make` + manual copy loop with `maps.Clone`
- `RemoveServerSession`: replaces the filtered copy loop with `maps.Clone` + `delete`

The branch is rebased onto current main.

### DIFF
--- a/internal/session/cache.go
+++ b/internal/session/cache.go
@@ -3,6 +3,7 @@ package session
 import (
 	"context"
 	"errors"
+	"maps"
 	"sync"
 
 	redis "github.com/redis/go-redis/v9"
@@ -73,9 +74,9 @@ func (c *Cache) AddSession(ctx context.Context, key, mcpServerID, mcpSession str
 		if val, ok := c.inmemory.Load(key); ok {
 			existing = val.(map[string]string)
 		}
-		next := make(map[string]string, len(existing)+1)
-		for k, v := range existing {
-			next[k] = v
+		next := maps.Clone(existing)
+		if next == nil {
+			next = map[string]string{}
 		}
 		next[mcpServerID] = mcpSession
 		c.inmemory.Store(key, next)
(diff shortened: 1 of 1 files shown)


Aman-Cool:
@david-martin @maleck13, `sync.Map` looks thread-safe so the bug hides well, but it only protects the outer index; the `map[string]string` value comes back as a live reference. Two concurrent tool calls to different backends in the same session both get a pointer to the same inner map, both write, process dies. 
Swapped to a plain map + `RWMutex` and made `GetSession` return a copy so callers can't accidentally poke live state. Redis path was fine, in-memory only.

Would love a second pair of eyes on this when you get a chance, happy to adjust if I missed something.

internal/session/cache.go
```diff
@@ -30,27 +34,39 @@ func (c *Cache) KeyExists(ctx context.Context, key string) (bool, error) {
...
 func (c *Cache) GetSession(ctx context.Context, key string) (map[string]string, error) {
-	if c.inmemory != nil {
-		val, ok := c.inmemory.Load(key)
+	if c.extClient == nil {
+		c.mu.RLock()
+		inner, ok := c.sessions[key]
+		var result map[string]string
 		if ok {
-			return val.(map[string]string), nil
+			result = make(map[string]string, len(inner))
```
maleck13: nit: I think we could change this to use `maps.Clone()` that would do the same thing but keep it a little more readable. Prob other places we could use this also in this file

maleck13 (dismissed):
minor nit to use `maps.Copy` instead of the manual copy

maleck13:
@Aman-Cool thanks again for the work. This will need a rebase. I think if your ok to make the change, lets move to use `maps.Copy` instead ?

maleck13 (commented):
needs rebasing

Aman-Cool:
Hey @maleck13, just wanted to give some context on this PR exists alongside #872.

When I opened #872 I kept the scope tight: just fix the race condition, nothing else. The copy-on-write logic in `AddSession` and `RemoveServerSession` still had the manual `make` + loop pattern, but changing that in the same PR would've mixed a bug fix with a cleanup, making it harder to review and bisect later if something went wrong.

This PR is the intentional follow-up for exactly that cleanup; replacing those manual copy loops with `maps.Clone` as you suggested in the review. It's already rebased onto current main so there are no conflicts. Happy to squash or adjust anything if needed...😁

maleck13 (approved):

maleck13:
@Aman-Cool Np. Thanks again.

PR #906, opened 2026-05-07
Review and improve issue-triage workflow (#650)

### LINKED ISSUE #650: Review and improve issue-triage workflow
## Summary

Review and improve our `.github/workflows/issue-triage.yaml` workflow, taking inspiration from the [coraza-kubernetes-operator issue-management workflow](https://github.com/networking-incubator/coraza-kubernetes-operator/blob/main/.github/workflows/issue-management.yml) and evaluating how well our current automation is serving us based on recent refinement calls.

## Current State

Our workflow uses inline JavaScript via `actions/github-script@v8` with these rules:
- No milestone: ensure `triage/needs-triage`, remove `triage/accepted` and priority labels
- Has milestone: ensure `triage/accepted`, default `priority/normal`
- Other triage label added: remove `triage/needs-triage`

### Known Issues (as of 2026-03-10)

**Issues missing triage labels entirely** (7 open issues):
- #601, #598, #596, #445, #422, #411, #404 — all have no milestone and no `triage/*` label

**Issues with duplicate priority labels** (workflow doesn't enforce single priority):
- #649 — has both `priority/normal` and `priority/low`
- #467 — has both `priority/normal` and `priority/low`

These indicate gaps in the current automation.

## What We Can Learn from coraza-kubernetes-operator

Their approach differs in several ways worth evaluating:

| Area | Our Workflow | Coraza Workflow |
|------|-------------|-----------------|
| **Implementation** | Inline JS in YAML (~150 lines) | Dedicated Go CLI tool checked into the repo |
| **PR skip** | Runtime check in JS | Job-level `if` condition (cleaner) |
| **Action pinning** | `actions/github-script@v8` (tag) | Pinned to full SHA (more secure) |
| **Top-level permissions** | Not set | `permissions: read-all` (least privilege) |
| **Close declined** | Not implemented | Auto-closes issues labeled `triage/declined` |
| **Testability** | Difficult (inline script) | Go tool can have unit tests |

## Proposed Investigation

- [ ] **Fix the PR skip**: Use job-level `if` condition instead of runtime check
- [ ] **Pin action versions**: Use SHA-pinned actions for supply chain security
- [ ] **Add top-level `permissions: read-all`**: Follow least-privilege principle
- [ ] **Enforce single priority label**: When a new `priority/*` label is added, remove others
- [ ] **Evaluate `triage/declined` + auto-close**: Decide if we want a declined flow that auto-closes issues
- [ ] **Backfill missing labels**: Run a one-time fix for the 7 issues missing triage labels
- [ ] **Fix duplicate priority labels**: Clean up #649 and #467
- [ ] **Consider extracting to a Go tool**: Evaluate whether the logic has grown enough to warrant a testable CLI tool (like coraza) vs keeping inline JS
- [ ] **Review trigger events**: Confirm the current event list covers all needed scenarios
- [ ] **Add `closed` event handling**: Consider removing triage/priority labels when issues are closed, or adding a `triage/done` label

## Context

Reference workflow: https://github.com/networking-incubator/coraza-kubernetes-operator/blob/main/.github/workflows/issue-management.yml

vibhor-5:
This PR improves the issue triage workflow based on the recommendations in #650 and inspirations from the coraza-kubernetes-operator.

### Changes:
- **Pin action versions**: Pinned actions/github-script to full SHA for security.
- **Fix PR skip**: Moved PR skip logic to a job-level if condition for cleaner execution.
- **Enforce single priority**: Added logic to automatically remove existing priority labels when a new one is added.
- **Auto-close declined**: Issues labeled triage/declined are now automatically closed.
- **Closed event handling**: Automatically removes triage/needs-triage when an issue is closed.

Closes #650

## Summary by CodeRabbit

* **Improvements**
  * Expanded issue triage automation to respond to more issue lifecycle events, with clearer event-aware logging.
  * Improved workflow safety using execution controls and tighter scope for triage on non–pull request issues.
  * Enforced a single active `priority/*` label by automatically removing conflicting priorities.
  * Updated declined-issue handling to better respect issue state, including appropriate issue closure behavior and avoiding further milestone/triage label changes afterward.
  * Refined label update logic to prevent conflicting add/remove operations.


vibhor-5:
@coderabbitai review

vibhor-5:
@coderabbitai review

david-martin:

.github/workflows/issue-triage.yaml
```diff
@@ -3,23 +3,26 @@ name: Issue Triage
 on:
   issues:
     types:
-      [opened, reopened, edited, labeled, unlabeled, milestoned, demilestoned]
+      [opened, reopened, edited, labeled, unlabeled, milestoned, demilestoned, closed]
+
+permissions: read-all
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
```
david-martin: should this be `github.event.issue.number` as this only triggers on issues, not prs?

vibhor-5: yes it should be, changed it.

vibhor-5:
Revisiting the PR just now , I ended up not choosing going with a dedicated go tool for this , it seemed overkill 
Some arguements for and against using a dedicated tool :

### Pros (Why use a Go CLI)

  • Highly Testable: You can write standard go test unit tests to catch logic bugs for CI changes before merging
  • Language Consistency: Developers don't have to context-switch to JavaScript; they can use the language they already write the project in.
  • Scalability: Much easier to add complex rules later (e.g., parsing  OWNERS  files, querying other APIs, interacting with the local filesystem) without creating a massive, unreadable inline script.

  ### Cons (Why keep inline JavaScript)

  • Increased Workflow Duration: Compiling a Go binary during CI adds more time, whereas  actions/github-script  executes instantly.
  • Higher Boilerplate: Requires manual setup for CLI flags, OAuth token handling, and API client initialization, whereas GitHub Actions injects an authenticated octokit client automatically into inline JS.
  • Maintenance Overhead: Introduces a new set of Go dependencies that must be maintained and tracked by dependabot.
  
  Even though there are strong pros the increased workflow duration seemed like an unnecesary compromise, what are your thoughts on this @david-martin.

Patryk-Stefanski (changes requested):
Thanks for the improvements — the SHA pin, permission tightening, and single-priority enforcement are all solid additions worth keeping.

Before this can merge, there are two bugs in the `triage/declined` auto-close rule, and I'd also ask you to pull out a couple of things that add noise without clear value.

---

**Bug 1 — `triage/declined` close fires on every event, not just when the label is added**

The rule currently checks:
```js
if (currentLabels.includes('triage/declined') && issue.state === 'open') {
```

This fires on *every* event that hits a declined open issue — `edited`, `milestoned`, `unlabeled`, etc. — issuing a redundant close API call each time. Combined with `cancel-in-progress: true`, two concurrent runs can both read `state === 'open'` from their respective payloads and both call close, generating duplicate closed events.

Fix: guard it so it only fires at the moment the label is applied:
```js
if (action === 'labeled' && eventLabel === 'triage/declined' && issue.state === 'open') {
```

---

**Bug 2 — milestoned `triage/declined` issues end up labelled `triage/accepted` after auto-close**

When `triage/declined` is added to an issue that has a milestone, the auto-close `await` fires first (correct), but then execution falls through to the `action !== 'closed'` block (true — this is a `labeled` event), enters the milestone path, adds `triage/accepted`, and removes `triage/declined`. The closed issue ends up labelled `triage/accepted` — the opposite of the intent.

Fix: return early after the auto-close so no further label rules run:
```js
if (action === 'labeled' && eventLabel === 'triage/declined' && issue.state === 'open') {
  await github.rest.issues.update({ ... state: 'closed', state_reason: 'not_planned' });
  return;
}
```

---

**Please remove from this PR:**

- **`concurrency: cancel-in-progress: true`** — this workflow mutates issue labels via a sequential loop of individual API calls. If a run is cancelled mid-loop, the issue is left in a partially-modified label state (some labels removed, others not). For a workflow where correctness of state matters more than avoiding overlap, `cancel-in-progress: false` (the default) is safer.

- **Remove `triage/needs-triage` on close** — nice-to-have with very little practical impact. Issues in `triage/needs-triage` state disappear from view once closed anyway. Pulling it out keeps the scope of this PR focused on things that matter.

vibhor-5:
@Patryk-Stefanski addresed the review. PTAL

.github/workflows/issue-triage.yaml
```diff
@@ -54,89 +59,120 @@ jobs:
...
+                issue_number: issueNumber,
+                state: 'closed',
+                state_reason: 'not_planned'
+              });
+              return;
+            }
 
-            } else {
-              // ensure 'triage/accepted' and remove other triage labels
-              const triageLabels = getLabelsStartingWith('triage/');
```
Patryk-Stefanski, line 108: The early `return` here skips the entire label cleanup phase. When an issue has `triage/needs-triage` (or any other triage/priority label), those are never removed — the issue ends up closed but still wearing stale labels.

`labelsToRemove` is empty at this point because the milestone block (which queues labels for removal) comes after this block and is never reached.

Two options:

**Option A** — remove the `return` and let the script continue. The milestone block will queue `triage/needs-triage` for removal naturally (no-milestone path, `remainingTriageLabels` will be non-empty due to `triage/declined` still being present):

```js
if (action === 'labeled' && eventLabel === 'triage/declined' && issue.state === 'open') {
  console.log('Issue marked as declined -> closing');
  await github.rest.issues.update({
    owner: context.repo.owner,
    repo: context.repo.repo,
    issue_number: issueNumber,
    state: 'closed',
    state_reason: 'not_planned'
  });
  // no return — fall through to label cleanup
}
```

**Option B** — explicitly queue stale labels before returning:

```js
getLabelsStartingWith('triage/').forEach(label => {
  if (label !== 'triage/declined') labelsToRemove.push(label);
});
getLabelsStartingWith('priority/').forEach(label => labelsToRemove.push(label));
// then move the apply-changes block here, or remove the return
```

Option A is simpler. Note that if you go with Option A you should also ensure the milestone block handles the `closed` action correctly (see separate comment).

.github/workflows/issue-triage.yaml
```diff
@@ -54,89 +59,120 @@ jobs:
...
-                  labelsToRemove.push(label);
-                  console.log(`Has milestone -> removing ${label}`);
+            // Rule: Milestone-based triage
+            // Skip milestone-based triage rules for closed events to avoid label mutation.
+            if (action !== 'closed') {
+              if (!hasMilestone) {
+                // 1. If an issue has no milestone:
+                
+                // a. Remove 'triage/accepted' (e.g., milestone was removed)
+                if (currentLabels.includes('triage/accepted')) {
```
Patryk-Stefanski, line 118: The PR description says "Automatically removes triage/needs-triage when an issue is closed" but this guard means the script is a no-op for `closed` events.

When `action === 'closed'`:
- Priority block: skipped (requires `action === 'labeled'`)
- Declined block: skipped (requires `action === 'labeled'`)
- This block: skipped entirely

`labelsToRemove` stays empty, the script logs "No label changes needed", and `triage/needs-triage` stays on the closed issue.

A dedicated closed handler is needed, for example placed before this block:

```js
// Rule: Issue closed -> remove triage/needs-triage
if (action === 'closed') {
  if (currentLabels.includes('triage/needs-triage')) {
    labelsToRemove.push('triage/needs-triage');
    console.log('Issue closed -> removing triage/needs-triage');
  }
}
```

Alternatively, restructure so the label cleanup runs for closed events too and only the "add labels" side is suppressed.

.github/workflows/issue-triage.yaml
```diff
@@ -3,23 +3,25 @@ name: Issue Triage
...
     types:
-      [opened, reopened, edited, labeled, unlabeled, milestoned, demilestoned]
+      [opened, reopened, edited, labeled, unlabeled, milestoned, demilestoned, closed]
+
+permissions: read-all
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.issue.number || github.ref }}
 
 jobs:
```
Patryk-Stefanski, line 11: Minor: `permissions: read-all` at the workflow level grants read on all scopes — contents, pull-requests, actions, packages, security-events, etc. The job already overrides to `issues: write` + `contents: read`, so this workflow-level grant is redundant and adds unnecessary surface area.

Consider dropping `permissions: read-all` and keeping only the job-level permissions, or replacing it with an explicit minimal set:

```yaml
permissions:
  issues: write
  contents: read
```

.github/workflows/issue-triage.yaml
```diff
@@ -3,23 +3,25 @@ name: Issue Triage
...
   triage:
+    if: github.event.issue.pull_request == null
     runs-on: ubuntu-latest
```
Patryk-Stefanski, line 14: Minor: the `|| github.ref` fallback is semantically wrong here. `github.event.issue.number` is always a positive integer for issue events, so this fallback never activates in practice. But if it ever did (e.g. a future trigger where issue number is absent), `github.ref` resolves to the default branch ref (`refs/heads/main`), collapsing all concurrent runs into one group and causing unrelated issues to cancel each other.

A safer fallback would be a static string:

```yaml
concurrency:
  group: ${{ github.workflow }}-${{ github.event.issue.number || 'default' }}
```

Or just remove the fallback entirely since issue events always have a number.

.github/workflows/issue-triage.yaml
vibhor-5: fixed this by queuing the stale labels for removal and setting action = 'closed' to skip the milestone block.

.github/workflows/issue-triage.yaml
vibhor-5: good catch, removed it.

.github/workflows/issue-triage.yaml
vibhor-5: removed the fallback since the issue number is always present here.

Patryk-Stefanski:
All previous review comments are addressed except one — the `closed` event not removing `triage/needs-triage` (https://github.com/Kuadrant/mcp-gateway/pull/906#discussion_r3499487799). That thread has no response yet. Once that's fixed this is good to go.

.github/workflows/issue-triage.yaml
vibhor-5: added this back in to fully address the feedback.

vibhor-5:
@Patryk-Stefanski addresed all comments now. kindly review

Patryk-Stefanski (approved):

PR #908, opened 2026-05-07
remove .md and docs file from codestyle ignored paths

Patryk-Stefanski:
The code-style.yaml workflow had paths-ignore entries for **.md and docs/**, meaning PRs that only touched markdown or docs files skipped the workflow entirely. But make spell runs cspell  against the entire repo. So a doc-only PR could merge with an unknown word (e.g. myprefix in docs/guides/tool-discovery.md), and then the next PR touching .go files would trigger the workflow, CSpell would scan the whole repo, find that word, and fail — even though that PR didn't introduce it.

### DIFF
--- a/.github/workflows/code-style.yaml
+++ b/.github/workflows/code-style.yaml
@@ -4,8 +4,6 @@ on:
   pull_request:
     branches: [ 'main',  ]
     paths-ignore:
-      - '**.md'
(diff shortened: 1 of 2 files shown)


Patryk-Stefanski:
closing in favour of https://github.com/Kuadrant/mcp-gateway/pull/845

PR #910, opened 2026-05-08
fix(broker): guard against nil gatewayServer to prevent panic

### LINKED ISSUE #882: broker: nil pointer panic in findToolConflicts causes CrashLoopBackOff on upstream connect

Abdul-Moiz31:
**Problem**

When a `tools/list_changed` notification fires before the broker has fully
wired up the gateway server, `manage()` calls into `findToolConflicts()`,
`removeAllTools()`, and the `AddTools`/`DeleteTools` sync block on a nil
`gatewayServer`. This causes a nil pointer dereference panic, crashing the
broker into CrashLoopBackOff.

**Solution**

Add nil guards at the three call sites that dereference `gatewayServer`:

- `findToolConflicts()` — return nil early (no conflicts to check)
- `removeAllTools()` — skip the `DeleteTools` call
- `manage()` — skip the `AddTools`/`DeleteTools` sync block

This is preferable to rejecting nil at construction time because nil is a
valid state for managers that do not exercise tool sync. Many existing tests
rely on this intentionally.

Also fixes a pre-existing parameter name typo: `gatewaySever` → `gatewayServer`.

**Testing**

Added `TestMCPManager_NilGatewayServer_NoPanic` as a regression test that
calls `manage()` with a nil gateway server and asserts no panic. All
existing tests continue to pass.

Fixes #882

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed potential crashes in the upstream manager when operating without a configured gateway server
  * Enhanced stability of tool management operations to handle missing gateway components gracefully

### DIFF
--- a/internal/broker/upstream/manager.go
+++ b/internal/broker/upstream/manager.go
@@ -102,14 +102,14 @@ const DefaultTickerInterval = time.Minute * 1
 // NewUpstreamMCPManager creates a new MCPManager for managing a single upstream MCP server.
 // The addTools and removeTools callbacks are used to update the gateway's tool registry.
 // The tickerInterval controls how often the manager checks backend health (use 0 for default).
-func NewUpstreamMCPManager(upstream MCP, gatewaySever ToolsAdderDeleter, logger *slog.Logger, tickerInterval time.Duration, policy mcpv1alpha1.InvalidToolPolicy) *MCPManager {
+func NewUpstreamMCPManager(upstream MCP, gatewayServer ToolsAdderDeleter, logger *slog.Logger, tickerInterval time.Duration, policy mcpv1alpha1.InvalidToolPolicy) *MCPManager {
 	if tickerInterval <= 0 {
 		tickerInterval = DefaultTickerInterval
 	}
 
 	return &MCPManager{
 		MCP:               upstream,
-		gatewayServer:     gatewaySever,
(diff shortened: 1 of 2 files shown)


david-martin:
Thanks for this. The nil gatewayServer scenario is now handled at construction time on main: https://github.com/Kuadrant/mcp-gateway/blob/d8b40ed6b59c7e0dd01e70f5e09551b603fb0667/internal/broker/upstream/manager.go#L119

NewUpstreamMCPManager returns an error if gatewayServer is nil, so the runtime guards are no longer needed. Closing as superseded.

Abdul-Moiz31:
Hey @david-martin  Thankyou so much for letting me know . 
Im very much intrested and contributed in here 

Can you please assign me and let me know if there is somthing pending to implement . even bug or new feature i will go and research and wrok on it properly . 

Thankyou !!

Looking forward

PR #918, opened 2026-05-08
fix: avoid recursive RLock deadlock in ValidateAllServers

Aman-Cool:
Fourth concurrency issue I've found in this area since #872. This one is in the broker itself rather than the session cache, but same class of problem, a lock being acquired in a way the Go runtime explicitly says it doesn't support.

`ValidateAllServers` takes `m.mcpLock.RLock()` at the top of the function, then immediately calls `m.RegisteredMCPServers()`, which takes `m.mcpLock.RLock()` again internally. The Go docs for `sync.RWMutex` call this out directly: if any goroutine is waiting for the write lock between those two read-lock calls, the second `RLock` blocks. The first read lock never gets released because the goroutine is stuck waiting. The writer never unblocks because it's waiting for that first reader. Permanent deadlock.

The failure window is: a readiness probe or monitoring system hits `/status` (which calls `ValidateAllServers`) at the same moment `OnConfigChange` is trying to acquire the write lock for a config reload. In a Kubernetes deployment with a liveness probe running every ~10s and a controller reconciling `MCPServerRegistration` objects on any cluster change, this race is just a matter of time. When it hits, the process stays up but `m.mcpLock` is stuck forever; and since `GetServerInfo` and `ToolAnnotations` both take `m.mcpLock.RLock()`, every subsequent tool call deadlocks too. No panic, no log line, just silent routing death.

Fix is a one-liner: `ValidateAllServers` already holds the outer `RLock`, so the call to `RegisteredMCPServers()` is redundant; just iterate `m.mcpServers` directly. `RegisteredMCPServers()` keeps its own lock for any external callers.

## Summary by CodeRabbit

* **Chores**
  * Improved evaluation workflow by adding model initialization step to ensure Ollama model is ready before tests run.
  * Updated agent configuration type for enhanced system compatibility.
  * Optimized internal server validation mechanism.

### DIFF
--- a/.github/workflows/gevals.yaml
+++ b/.github/workflows/gevals.yaml
@@ -114,6 +114,13 @@ jobs:
 
           echo "Ollama setup complete using model $MODEL_NAME"
 
+      - name: Warm up Ollama model
+        if: env.OPENAI_API_KEY == 'ollama'
+        run: |
+          echo "Warming up model to ensure it is loaded before eval..."
+          curl -s -X POST http://localhost:11434/api/generate \
+            -d '{"model":"qwen2.5:1.5b","prompt":"hi","stream":false}' > /dev/null || true
+
       - name: Run mcpchecker
         run: |
           echo "Installing mcpchecker..."
--- a/evals/gemini-agent/agent.yaml
+++ b/evals/gemini-agent/agent.yaml
@@ -5,7 +5,7 @@ builtin:
   # Using openai-agent type as a generic LLM client that supports OpenAI API format
   # Set MODEL_BASE_URL to point to Gemini (e.g. https://generativelanguage.googleapis.com/v1beta/openai/)
   # or any other OpenAI-compatible endpoint.
-  type: "openai"
+  type: "llm-agent"
 
   model: "openai:qwen2.5:1.5b"
   params:
--- a/evals/gemini-agent/agent.yaml
+++ b/evals/gemini-agent/agent.yaml
@@ -5,7 +5,7 @@ builtin:
   # Using openai-agent type as a generic LLM client that supports OpenAI API format
   # Set MODEL_BASE_URL to point to Gemini (e.g. https://generativelanguage.googleapis.com/v1beta/openai/)
   # or any other OpenAI-compatible endpoint.
-  type: "openai-acp"
+  type: "openai"
 
   model: "openai:qwen2.5:1.5b"
   params:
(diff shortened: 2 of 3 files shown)


Aman-Cool:
Hey @maleck13, continuing the same thread as #872, #888, and #911; found another one while tracing through the broker locking.

`ValidateAllServers` holds `m.mcpLock.RLock()` and then calls `RegisteredMCPServers()` which tries to grab `m.mcpLock.RLock()` again. That's fine until any concurrent `OnConfigChange` is waiting for the write lock ; at that point the second `RLock` blocks, the first never releases, and it's a permanent deadlock. Since `GetServerInfo` uses the same lock, every new tool call stops routing too. No panic, process looks alive, just silently dead. Readiness probe + controller reconciliation in Kubernetes is enough to hit it.

Fix is a one-liner ; `ValidateAllServers` already holds the outer lock so just iterate `m.mcpServers` directly instead of going through `RegisteredMCPServers()`. 
Happy to adjust if anything looks off...

Aman-Cool:
@maleck13, Just pushed a second commit to fix the gevals check. Two things were wrong:

The agent type `"openai-acp"` is deprecated as of `mcpchecker` v0.0.16 ; it throws a warning and its tool-calling behaviour isn't reliable anymore. Switched to `"openai"` which is the current supported type.

The other issue is specific to the Ollama fallback path. When the model cache is warm the check passes; when it's cold (fresh pull, no prior run on the runner) `qwen2.5:1.5b` makes 0 tool calls and the `minToolCalls: 1` assertion fails. The model just isn't loaded into memory yet. Added a single warmup inference call right before `mcpchecker` kicks off ; only runs when falling back to Ollama, no-ops if it fails, keeps the happy path untouched...

Aman-Cool:
@maleck13, Pushed another fix, my bad on the last one.

I assumed `"openai"` was the obvious replacement for the deprecated `"openai-acp"` but that type doesn't exist at all in `mcpchecker` v0.0.16, which is why it hard-crashed with `unknown builtin type: openai` instead of just running and failing the assertion like before. Dug into the `mcpchecker` source and the actual canonical type is `"llm-agent"` ; that's what both `"openai-acp"` and `"openai-agent"` alias to internally. Switched to that. Should be good now.

david-martin:
@Aman-Cool do you mind if we close this and follow up with a separate PR for the gevals change? (i've merged a similar fix for the deadlock in #921 which also included a test)

I think we might actually drop gevals from CI soon anyways as it's started failing a couple times now and we're not overly familiar with it (it was added by someone who is not active on the project now, so there's maintenance overhead if we want to keep it). The value of having it is not clear to me at least.

PR #919, opened 2026-05-08
guard span.SetAttributes with IsRecording in router hot path

shiavm006:
CLAUDE.md says:

> Guard span attributes: `if span.IsRecording()` before `span.SetAttributes(...)`

The pattern is followed in `internal/otel/logging.go:38` and
`internal/mcp-router/server.go:214`, but missed in the per-request hot path:

- `internal/mcp-router/request_handlers.go` — 11 unguarded calls
- `internal/mcp-router/server.go:244` — 1 unguarded
- `internal/mcp-router/tracing.go` — `recordError` helper unguarded

## Summary by CodeRabbit

* **Chores**
  * Optimized internal tracing instrumentation to reduce overhead and improve system performance.

### DIFF
--- a/internal/mcp-router/request_handlers.go
+++ b/internal/mcp-router/request_handlers.go
@@ -215,13 +215,19 @@ func (s *ExtProcServer) RouteMCPRequest(ctx context.Context, mcpReq *MCPRequest)
 	s.Logger.DebugContext(ctx, "HandleMCPRequest ", "session id", mcpReq.GetSessionID())
 	switch {
 	case mcpReq.isElicitationResponse():
-		span.SetAttributes(attribute.String("mcp.route", "elicitation-response"))
+		if span.IsRecording() {
+			span.SetAttributes(attribute.String("mcp.route", "elicitation-response"))
+		}
 		return s.HandleElicitationResponse(ctx, mcpReq)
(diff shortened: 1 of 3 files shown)


david-martin:
Thanks for the contribution. After digging into this, I don't think we need these guards. The OTel Go SDK already handles this internally:

- When tracing is disabled, the span is a `nonRecordingSpan` whose `SetAttributes` is an empty no-op.
- When tracing is enabled, `recordingSpan.SetAttributes` already checks `isRecording()` internally and returns early if the span has ended.

The only overhead without the guard is constructing the `attribute.KeyValue` structs before the call, but these are small stack-allocated value types with no heap allocation. The cost is a few nanoseconds, roughly equivalent to the `IsRecording()` branch check itself.

The one scenario where caller-side guards help is expensive argument construction (e.g., `fmt.Sprintf`), but that only appears in `recordError` which is an error path, not the hot path.

We had guidance in our CLAUDE.md recommending this pattern, but that was overly broad. We'll be updating it. Closing this, but appreciate you taking the time.

david-martin:
follow up #969

shiavm006:
Thanks  @david-martin , makes sense. i see  #969  useful to know the SDK already covers the non-recording case. Will keep that in mind for future tracing work.

PR #922, opened 2026-05-09
fix: eliminate data race between LoadConfig and broker OnConfigChange

Aman-Cool:
`MCPServersConfig` has a `lock sync.RWMutex` (it's right there on the struct), and every read accessor uses it correctly ; `GetServerConfigByName` takes `RLock`, `RegisterObserver` takes `Lock`, `Notify` takes `RLock`. But `LoadConfig` in `main.go` writes `mcpConfig.Servers` and `mcpConfig.VirtualServers` directly as raw field assignments, completely bypassing the lock. The reason it was never caught is that lock is unexported, so nothing outside the config package can call it, and a locked setter method was never added. The field itself is public so the write compiles fine and looks innocent.

The race fires during a hot-reload when the broker's `OnConfigChange` goroutine from a previous `Notify` is still running ; which is realistic because that goroutine connects to backend servers, and dialling + MCP initialize can take a few seconds. While it's inside the deregistration loop reading `conf.Servers` through `slices.ContainsFunc`, `LoadConfig` sets `mcpConfig.Servers = []*config.MCPServer{}` to reset before unmarshalling. `slices.ContainsFunc` over an empty slice always returns false, so every server ID fails the membership check and every manager gets `Stop()`-ped: connections closed, tools removed from the gateway. In Kubernetes this is normal usage ; the controller writes the config Secret, the pod remounts it, and `fsnotify` delivers two close-spaced events (an unlink and a create) for one logical update. The broker goroutine from event one is still connecting when event two fires `LoadConfig`. No panic, no error log, just a silent window where `tools/list` returns empty and every `tools/call` gets "tool not found".

The fix adds `SetServers` to `MCPServersConfig` as the proper locked write path, along with `ListServers` and `ListVirtualServers` for locked snapshots on the read side. `LoadConfig` now unmarshals into local variables and calls `SetServers` once the full new config is ready, so the shared struct is never in a half-written state. The broker's `OnConfigChange` takes a snapshot of both slices at the very top of the function, before acquiring `mcpLock`, so everything it does after that is working off an immutable local copy that `LoadConfig` can't touch mid-loop.

### DIFF
--- a/cmd/mcp-broker-router/main.go
+++ b/cmd/mcp-broker-router/main.go
@@ -385,26 +385,26 @@ func LoadConfig(path string) {
 	if err != nil {
 		log.Fatalf("Error reading config file: %s", err)
 	}
-	// reset the servers to avoid old configs being written to
-	mcpConfig.Servers = []*config.MCPServer{}
-	err = viper.UnmarshalKey("servers", &mcpConfig.Servers)
+	var newServers []*config.MCPServer
+	err = viper.UnmarshalKey("servers", &newServers)
 	if err != nil {
 		log.Fatalf("Unable to decode server config into struct: %s", err)
 	}
-	mcpConfig.VirtualServers = []*config.VirtualServer{}
+	var newVirtualServers []*config.VirtualServer
 	// Load virtualServers if present - this is optional
 	if viper.IsSet("virtualServers") {
-		err = viper.UnmarshalKey("virtualServers", &mcpConfig.VirtualServers)
+		err = viper.UnmarshalKey("virtualServers", &newVirtualServers)
 		if err != nil {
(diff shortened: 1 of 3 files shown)


Aman-Cool:
Hey @maleck13, same class of concurrency issue as #872, #888, #911, and #918 but this one's in the config write path; `LoadConfig` was writing directly to `mcpConfig.Servers` without the lock that's already on the struct, so a broker goroutine mid-reload could see an empty server list and stop everything. 
Happy to adjust anything if needed...

Aman-Cool:
@coderabbitai , review this please...

Aman-Cool:
@coderabbitai any updates?

maleck13 (approved):
Thanks

PR #924, opened 2026-05-10
fix(session): bound JWTManager.Terminate cache deletion with a 5s timeout

SarthakB11:
## Summary

Symmetric fix to #911 on the JWT-manager side of session teardown.

`JWTManager.Terminate` is invoked by the upstream MCP server library when the gateway needs to close a session. The `SessionIdManager` interface from `mark3labs/mcp-go` does not pass a `context.Context`, so the implementation creates one internally, but it used `context.TODO()`. That has no cancellation and no deadline, so if `sessionDeleter` is the Redis-backed cache and Redis is unreachable, the goroutine driving session expiry blocks indefinitely on the cache deletion call. That leaks the goroutine and can hold the upstream library's session-manager lock.

Replace `context.TODO()` with `context.WithTimeout(context.Background(), terminateTimeout)` and defer the cancel. `terminateTimeout` is a package-level constant set to 5 seconds so the value is visible at one site and easy to tune. Same shape as #911, just on the other end of the session lifecycle.

`TestTerminate` gains a subtest that wires `JWTManager` with a `Deleter` whose `DeleteSessions` blocks until the supplied context is canceled. Without the fix, `Terminate` blocks forever; with it, `Terminate` returns within the deadline plus a small buffer.

## Verification

```
go build ./...
go vet ./...
go test -race -count=1 -timeout 60s ./internal/session/...
make lint
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Session termination operations now have a bounded timeout to prevent indefinite blocking.

* **Tests**
  * Added test coverage for session termination timeout behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Kuadrant/mcp-gateway/pull/924)

### DIFF
--- a/internal/session/jwt.go
+++ b/internal/session/jwt.go
@@ -211,8 +211,7 @@ func (m *JWTManager) GetExpiresIn(tokenValue string) (time.Time, error) {
 	return nd.Time, nil
 }
 
-// terminateTimeout bounds the cache deletion call made by Terminate so a stalled
-// backing store cannot block the upstream session manager indefinitely.
+// bounds cache deletion in Terminate so a stalled store cannot block forever
 const terminateTimeout = 5 * time.Second
 
 // Terminate part of the SessionIDManager interface. Will remove the associated sessions from cache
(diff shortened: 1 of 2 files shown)


SarthakB11:
@coderabbitai review

david-martin:
Good fix. This will need a minor rebase (typo fix landed on main in the same file). After rebasing, this should be good to go.

SarthakB11:
Rebased onto main. Picked up the `errors` import that landed alongside the typo fix; no behavior changes from my side. Build + race + lint clean locally.

david-martin:

jasonmadigan (approved):
lgtm. bounds the stalled-deleter path correctly, test proves the timeout fires.

PR #930, opened 2026-05-10
fix: use singleflight to prevent duplicate backend sessions on concurrent tool calls

Aman-Cool:
#911 cleaned up `sessionCloser` inside `initializeMCPSeverSession`. While I was tracing through that function I missed something: the initialization path itself has the same class of TOCTOU problem that #872 fixed in the cache mutations.

When a client hits a backend server for the first time, the router checks the session cache for an existing backend session ID and, finding nothing, drops into `initializeMCPSeverSession`. That function does a second cache check before calling `InitForClient` ; but the check is a bare `sync.Map.Load` with no exclusion around it; two concurrent goroutines on the same gateway session can both pass it clean, both call `InitForClient`, and both race toward `AddSession`.

`AddSession` is safe ; #872 took care of that ; but by the time the loser hits the write, its `clientHandle` is already open. The second call silently overwrites the first session ID in the map, and the first `clientHandle` is now unreachable: connection open, mcp-go SSE reader goroutine alive, backend session held, none of it reachable until `time.AfterFunc` fires at JWT expiry. On top of that, both goroutines registered their own `sessionCloser`, so there are two cleanup timers for the same session. The first one to fire calls `DeleteSessions` and wipes the whole session map for that client; any tool call landing in the gap between the two firings sees an empty cache, re-initializes, and potentially repeats the cycle.

The realistic trigger is any agent framework dispatching multiple tool calls in parallel on the same session before a backend session exists ; which is the normal first-turn pattern when an LLM calls more than one tool. The race window is the full `InitForClient` round-trip (TCP + MCP handshake, 50–200ms on a remote backend), so it's wide enough to hit consistently under load.

`singleflight` fix wraps the creation block in `singleflight.Do` keyed on `gatewaySessionID + ":" + serverName`. Concurrent initializations for the same pair collapse into one; all callers get the same session ID; exactly one `clientHandle`, one `time.AfterFunc`.

`golang.org/x/sync` is already an indirect dep ; no new module dependency.

---

**Additional fixes included in this PR:**

**Missing `ExtProcServer` fields for URL elicitation** : the merge with main brought in references to `ElicitationEnabled` and `TokenElicitationMap` throughout `request_handlers.go` and `response_handlers.go`, but those fields were absent from the `ExtProcServer` struct, causing compilation failures. Added `ElicitationEnabled bool` and `TokenElicitationMap elicitation.Map` to the struct.

**Hairpin init request uses cached user token** : for servers with `tokenURLElicitation` configured, the router correctly injects the cached per-user token into the Envoy `HeadersBuilder` for the forwarded tool call, but `initializeMCPSeverSession` was building `passThroughHeaders` from the original client request headers. Those headers carry the client's gateway-bound JWT, not the upstream API key, so the hairpin `initialize` request reached the backend with the wrong credential and was rejected with 401. Fixed by looking up the cached user token inside `initializeMCPSeverSession` and overwriting `passThroughHeaders["authorization"]` before calling `InitForClient`.


Aman-Cool:
@maleck13, Same area as #911 ; was poking around `initializeMCPSeverSession` after that fix and spotted this one. Pretty confident in the root cause and the `singleflight` approach but would appreciate a second pair of eyes, especially on whether the flight key granularity looks right to you. 
Happy to adjust anything.

david-martin:

Aman-Cool:

PR #931, opened 2026-05-11
Url elicitation combined

### LINKED ISSUE #831: Documentation for URL elicitation
## Documentation for URL elicitation

Write user-facing guide and update security docs for the URL elicitation feature.

### Files
- `docs/guides/url-elicitation.md` — user-facing guide covering both broker page and external URL patterns
- `docs/design/security-architecture.md` — update with token data boundaries and isolation
- `docs/reference/mcpserverregistration.md` — update API reference for `credentialURLElicitation` field

### Acceptance Criteria
- [ ] Guide covers both broker page and external URL patterns
- [ ] Security doc covers token isolation and known risks
- [ ] API reference updated

### Jira
Part of CONNLINK-998

maleck13:
This is a big PR.  I have attempted to break each piece into a logical commit to help with reviewing.

 **Important** Elicitation design change: The broker now handles elicitation. Router only routes and pull tokens from the cache

## Summary
 Add URL token elicitation feature: 
  - clients without a cached token receive a -32042 error with a token page URL, submit their token via URLElicitation, and retry to tool call successfully
  - Router detects missing tokens and routes to broker `/mcp/elicitation` endpoint which returns the expected SSE error response force elicitation flow
  - Token page (`/tokens`) serves an HTML form, validates and caches submitted tokens per session
  - Cached tokens are injected by the router on subsequent tool calls
  - Non-elicitation-capable clients get an `isError` result instead of -32042

### Docs

Docs added:
  - docs/guides/url-elicitation.md — user-facing guide
  - demos/url-elicitation-github/ — demo script and sample manifests

### E2E test cases

  1. URL elicitation triggers on missing token — elicitation-capable client gets -32042 with token page URL
  2. Full round-trip — trigger → GET form → POST token → retry succeeds
  3. Cached token reused — subsequent calls skip elicitation
  4. Non-elicitation client gets standard error — isError result, not -32042
  5. Server without tokenURLElicitation unaffected — no elicitation flow, direct tool call works

## Manual verification

Trying the demo

  Prerequisites

  - make local-env-setup completed
  - make auth-example-setup-no-vault completed (sets up Keycloak)
  - A GitHub PAT (export GITHUB_PAT=ghp_...) with read:user scope

  Run

  ./demos/url-elicitation-github/demo.sh

  This sets up a GitHub MCP server behind the gateway with per-user token collection. When a
  tool call requires a token, the gateway returns an elicitation URL instead of failing.

  Connect Claude Code

  NODE_TLS_REJECT_UNAUTHORIZED=0 claude mcp add mcp-gateway --transport http
  http://mcp.127-0-0-1.sslip.io:8001/mcp
  NODE_TLS_REJECT_UNAUTHORIZED=0 claude
  run /mcp and authorize

  Test the flow

  1. Ask Claude to use a GitHub tool, e.g. "use mcp-gateway and call github_get_me"
  2. Claude prompts you to open a URL in your browser
  3. Open the URL — you'll be redirected to Keycloak to login (user: mcp-user / password:
  password)
  4. After login you'll see a form asking for your GitHub token
  5. Paste your GitHub PAT and click Submit
  6. Claude retries the tool call automatically — you should see your GitHub user data
  7. Subsequent calls use the cached token

  Cleanup

  ./demos/url-elicitation-github/demo.sh cleanup

  Known limitations

  The demo uses manual AuthPolicies instead of OIDCPolicy due to three issues discovered
  during development:

  1. OIDCPolicy post-auth redirect drops the port — redirects to http://host/ instead of
  http://host:8001/
  2. Target cookie drops the query string — loses ?elicitation_id=... after auth
  3. Rego cookie parser breaks on = in values — split("=") fails for cookies containing =

## Summary by CodeRabbit

* **New Features**
  * URL-based per-user token elicitation with a /tokens token submission page and SSE elicitation flow.
  * Per-gateway (urlElicitation) and per-server (tokenURLElicitation) configuration to enable/override elicitation.
  * Router/broker support to trigger elicitation, inject cached user tokens into upstream calls, and handle retries.
  * Session cache support for storing/encrypting per-user upstream tokens.

* **Documentation**
  * Complete guides, design, and reference docs for URL token elicitation.

* **Chores**
  * Increased E2E timeout from 15m to 20m.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Kuadrant/mcp-gateway/pull/931?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


cmd/mcp-broker-router/main.go
```diff
@@ -181,7 +184,16 @@ func main() {
 		}
 	}
 
-	sessionCache, err := session.NewCache(session.WithRedisClient(redisClient))
+	var sessionCacheOpts []func(*session.Cache)
+	sessionCacheOpts = append(sessionCacheOpts, session.WithRedisClient(redisClient))
+	if redisClient != nil {
+		encKey, encErr := session.DeriveEncryptionKey([]byte(jwtSigningKeyFlag))
```
maleck13, line 182: we should prob rename from jwt to just gatewaySigningkey

cmd/mcp-broker-router/main.go
```diff
@@ -314,7 +337,7 @@ func main() {
 	}
 }
 
-func setUpHTTPServer(address string, mcpBroker broker.MCPBroker, sessionManager *session.JWTManager, writeTimeoutSecs int64) (*http.Server, *server.StreamableHTTPServer) {
+func setUpHTTPServer(address string, mcpBroker broker.MCPBroker, sessionManager *session.JWTManager, writeTimeoutSecs int64, tokenHandler http.Handler, elicitationHandler http.Handler) (*http.Server, *server.StreamableHTTPServer) {
```
maleck13, line 331: signature here is getting a bit out of hand. Needs some attention

jasonmadigan:
👀

internal/broker/tokens.go
```diff
@@ -0,0 +1,169 @@
...
+
+	w.Header().Set("Content-Type", "text/html; charset=utf-8")
+	w.WriteHeader(http.StatusOK)
+	_, _ = w.Write([]byte(renderTemplate("token_form.html", tokenFormData{
+		ServerName:    entry.ServerName,
+		ElicitationID: elicitationID,
+	})))
+}
+
+func (h *TokenHandler) handlePost(w http.ResponseWriter, r *http.Request) {
```
jasonmadigan, line 103: no csrf protection on this POST handler. the design doc specifies `csrfToken` in the elicitation entry schema but it's not implemented. the `Entry` struct has no csrf field either. textbook csrf if an attacker obtains or guesses the elicitation id (which is in the url, so it leaks via referer headers and browser history).

internal/broker/tokens.go
```diff
@@ -0,0 +1,169 @@
...
+	if err := r.ParseForm(); err != nil {
```
jasonmadigan, line 105: `ParseForm()` called without `http.MaxBytesReader` -- no body size limit. attacker can POST an arbitrarily large body. gosec G120 flags this too. wrap with `r.Body = http.MaxBytesReader(w, r.Body, maxSize)` before parsing.

internal/mcp-router/response_builder.go
```diff
@@ -109,6 +109,9 @@ func (rb *ResponseBuilder) WithImmediateResponse(statusCode int32, message strin
 // WithImmediateJSONRPCResponse adds an immediate response, typically JSON-RPC format,
 // that terminates request processing
 func (rb *ResponseBuilder) WithImmediateJSONRPCResponse(statusCode int32, setHeaders []*basepb.HeaderValueOption, message string) *ResponseBuilder {
+	allHeaders := append(setHeaders, &basepb.HeaderValueOption{
```
jasonmadigan: `allHeaders := append(setHeaders, ...)` -- if a caller ever passes a slice with spare capacity, this mutates the caller's backing array. both current call sites pass fresh literals so it's safe today, but the pattern is a footgun. safer to copy first:
```go
allHeaders := make([]*basepb.HeaderValueOption, 0, len(setHeaders)+1)
allHeaders = append(allHeaders, setHeaders...)
allHeaders = append(allHeaders, &basepb.HeaderValueOption{...})
```

internal/mcp-router/request_handlers.go
```diff
@@ -630,3 +679,75 @@ func (s *ExtProcServer) HandleNoneToolCall(ctx context.Context, mcpReq *MCPReque
...
+	sessionID := mcpReq.GetSessionID()
+
+	token, ok, err := s.SessionCache.GetUserToken(ctx, sessionID, serverInfo.Name)
+	if err != nil {
+		s.Logger.ErrorContext(ctx, "user token cache lookup failed", "error", err)
+		return nil, fmt.Errorf("user token lookup: %w", err)
+	}
+	if ok {
+		s.Logger.DebugContext(ctx, "found cached user token", "server", serverInfo.Name)
+		headers.WithAuth(token)
```
jasonmadigan, line 857: token injected as-is without `Bearer ` prefix. if a user pastes a raw PAT (e.g. `ghp_abc`) into the form, this sends `Authorization: ghp_abc` which most services will reject. the form label says "Token or API Key" with no guidance on format. either normalise here (prepend `Bearer ` if not present) or make the expected format explicit in the form and docs.

cmd/mcp-broker-router/main.go
```diff
@@ -138,6 +140,7 @@ func main() {
 	flag.BoolVar(&enforceCapabilityFilteringFlag, "enforce-capability-filtering", false, "when enabled an x-mcp-authorized header will be needed to return any capabilities (tools, prompts)")
 	flag.StringVar(&invalidToolPolicyFlag, "invalid-tool-policy", "FilterOut", "policy for upstream tools with invalid schemas: FilterOut (default) or RejectServer")
 	flag.IntVar(&maxRequestBodySize, "max-request-body-size", 5242880, "max request body size in bytes for the ext_proc router. Default 5MB.")
+	flag.BoolVar(&enableURLElicitationFlag, "enable-url-elicitation", true, "enable URL elicitation for per-user credential collection")
```
jasonmadigan: defaulting to `true` for a new feature is surprising. combined with no AuthPolicy = no identity check on the token form, so anyone with the elicitation url can inject tokens into another user's session. should default to `false` so operators opt in.

cmd/mcp-broker-router/main.go
```diff
@@ -353,23 +376,27 @@ func setUpHTTPServer(address string, mcpBroker broker.MCPBroker, sessionManager
 
 	mux.HandleFunc("/status", mcpBroker.HandleStatusRequest)
 	mux.HandleFunc("/status/", mcpBroker.HandleStatusRequest)
+	mux.Handle("/tokens", tokenHandler)
```
jasonmadigan: this route is registered unconditionally regardless of the `--enable-url-elicitation` flag. the flag gates the elicitation logic in the router/broker, but the `/tokens` http endpoint is always live. gate the registration behind the flag.

internal/broker/elicitation_handler.go
```diff
@@ -0,0 +1,91 @@
...
+func (h *ElicitationHandler) buildElicitationURL(serverName, elicitationID string, r *http.Request) string {
+	serverConfig, err := h.Config.GetServerConfigByName(serverName)
+	if err == nil && serverConfig.TokenURLElicitation != nil && serverConfig.TokenURLElicitation.URL != "" {
+		return serverConfig.TokenURLElicitation.URL + "?elicitation_id=" + elicitationID
+	}
+	scheme := r.Header.Get("X-Forwarded-Proto")
+	if scheme == "" {
+		scheme = "https"
+	}
+	host := r.Host
```
jasonmadigan: fallback to `r.Host` for url construction. if the broker is ever exposed directly (not behind envoy), an attacker controls this value and can point the elicitation url at a phishing page. the primary path uses the configured url which is fine, but this fallback is reachable when no custom url is configured.

internal/broker/tokens.go
```diff
@@ -0,0 +1,169 @@
...
+}
+
+func (h *TokenHandler) handleGet(w http.ResponseWriter, r *http.Request) {
+	elicitationID := r.URL.Query().Get("elicitation_id")
+	if elicitationID == "" {
+		h.sendError(w, http.StatusBadRequest, "missing elicitation_id parameter")
+		return
+	}
+
+	entry, ok, err := h.elicitationMap.Lookup(r.Context(), elicitationID)
```
jasonmadigan, line 66: `Lookup` is non-destructive -- the elicitation id stays valid after the form is rendered. only the POST uses `Claim`. an attacker who obtains the id from browser history, server logs, or referer headers can race to submit before the legitimate user. also missing `Referrer-Policy: no-referrer` on this response to prevent leakage.

api/v1alpha1/types.go
```diff
@@ -60,8 +60,23 @@ type MCPServerRegistrationSpec struct {
...
+	// +optional
+	TokenURLElicitation *TokenURLElicitationConfig `json:"tokenURLElicitation,omitempty"`
+}
+
+// TokenURLElicitationConfig configures per-user token collection via URL elicitation.
+type TokenURLElicitationConfig struct {
+	// url overrides the default broker token page URL.
+	// When set, users are directed to this external URL (e.g. a Vault UI) instead of the broker's built-in page.
+	// +optional
+	URL string `json:"url,omitempty"`
```
jasonmadigan, line 80: no validation on this field -- no CEL rule, no `+kubebuilder:validation:Pattern`, no `format: uri`. malformed or malicious urls (`javascript:`, `data:`) silently accepted and end up in the -32042 error response sent to clients.

internal/session/crypto.go
```diff
@@ -0,0 +1,81 @@
...
+	internaljwt "github.com/Kuadrant/mcp-gateway/internal/jwt"
+	"golang.org/x/crypto/hkdf"
+)
+
+// DeriveEncryptionKey derives a 32-byte AES-256 key from the session signing key using HKDF.
+func DeriveEncryptionKey(signingKey []byte) ([]byte, error) {
+	if len(signingKey) < 16 {
+		return nil, fmt.Errorf("signing key too short: need at least 16 bytes")
+	}
+	r := hkdf.New(sha256.New, signingKey, nil, []byte("mcp-gateway-user-token-encryption"))
```
jasonmadigan, line 22: nil salt here contradicts the design doc which says "using a context-specific salt". the info param provides domain separation which is good, but rfc 5869 recommends a non-nil salt. a fixed string like `[]byte("mcp-gateway-v1")` would be strictly better and match what the doc claims.

jasonmadigan (changes requested):
🤖 some things

cmd/mcp-broker-router/main.go
maleck13: Yeah ok that makes sense. Our guide can call this out. 1) put authpolicy in place 2) enable

maleck13:
Thanks for the initial review @jasonmadigan . I will work through these items 👍

internal/broker/tokens.go
maleck13: I don't thinks CSRF is a concern here as we don't use any ambient credentials. But the Referrer-Policy makes sense to avoid leaking the elicitation id as this *is* the auth if you don't apply an auth policy (which ofc you should be doing).

internal/mcp-router/request_handlers.go
maleck13: I am thinking we allow a scheme to be added under tokenURLElicitation:

```yaml
tokenURLElicitation:
    headerName: Authorization        # default
    headerValueFormat: "Bearer {token}"  # default
```

WDYT

internal/broker/elicitation_handler.go
maleck13: yep 👍 should only use the h.Config.GetExternalHostname

internal/broker/tokens.go
maleck13: Yes I have added a Referrer policy. The token has a 5 minute lifetime to try reduce the chances of this happening also

internal/session/crypto.go
maleck13: went back and forwards on this one. Our session-signing-key is already strong, So rather than add another flag, I think this is ok

maleck13:
@jasonmadigan most covered there now:

 Comment: Referrer-Policy + MaxBytesReader
  Fix: Added both to token handler
  Commit: c65928a
  ────────────────────────────────────────
  Comment: append footgun
  Fix: Copy slice before appending
  Commit: 8e9fefd
  ────────────────────────────────────────
  Comment: Default true + route gating
  Fix: Default false, gate routes behind flag, update guide/demo
  Commit: da2af71
  ────────────────────────────────────────
  Comment: r.Host fallback
  Fix: Use configured hostname only
  Commit: a0afdc8
  ────────────────────────────────────────
  Comment: URL validation + HKDF salt doc
  Fix: Pattern validation on CRD, fix doc
  Commit: 793d13f

  Still open as future work (documented in design doc):
  - Bearer prefix / headerName+headerValueFormat — design doc updated, implementation
  deferred

Main decision point is this last one. I think it makes sense to allow specifying the header and format

jasonmadigan:
👀

jasonmadigan:
this looks good now with the other changes. a failing e2e test but other than that, think this is almost ready to merge

cmd/mcp-broker-router/main.go
maleck13: https://github.com/Kuadrant/mcp-gateway/issues/968

internal/mcp-router/request_handlers.go
maleck13: https://github.com/Kuadrant/mcp-gateway/issues/970

maleck13:
I have started to spec out how we want to handle the OAuth flow for browser. I hadn't considered this deeply enough initially. 

- OIDCPolicy could be an option 
- Allowing configuring the broker also an option

internal/broker/tokens.go
maleck13: That said. We do need to use a cookie with browser as the JWT will be set by keycloak as a cookie. Re-evaluating

maleck13: I have addressed this need in a later commit

jasonmadigan:
👀

jasonmadigan:
> Open the URL — you'll be redirected to Keycloak to login (user: mcp-user / password: password)

had to use `mcp` / `mcp`


```
⏺ mcp-gateway - Get my user profile (MCP)

──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
  MCP server “mcp-gateway” — waiting for completion

  Authorization is required to access this service.

  http://mcp.127-0-0-1.sslip.io:8001/tokens?elicitation_id=75d5296d-5682-4b22-bc2f-f19ef025e572

  Waiting for the server to confirm completion…
```


Pasted  (valid) token into http://mcp.127-0-0-1.sslip.io:8001/tokens:

```
Token Stored
Your token for mcp-test/github has been stored. You can close this window and retry the tool call.
```

But then after `Reopen URL` in claude:

```
{
"error": "invalid or expired elicitation_id"
}
```

Trying `Retry now`:

```
⏺ The call failed — the mcp-gateway returned a transport error suggesting the GitHub MCP server behind it might be using
  legacy SSE transport rather than the streamable HTTP that the gateway expects. This is an upstream connection issue between
  the gateway's broker and the GitHub MCP server, not an auth problem.

  Is this a known configuration, or would you like me to dig into the gateway logs to see what's happening?
```

Wasn't able to get further.

maleck13:
But then after Reopen URL in claude:

```
{
"error": "invalid or expired elicitation_id"
}
```

This would be expected as the token is single user and times out after 2 mins

The re-try should have worked though

maleck13:
ok there was a regression from a rebase. Believe I have it sorted now

maleck13:
@jasonmadigan ok should all be good. I have given it a good internal review and fixed the rebase issue. Ready for a final review and verification

demos/url-elicitation-github/demo.sh
```diff
@@ -0,0 +1,235 @@
...
+  echo ""
+  echo "Get a token at: https://github.com/settings/tokens/new"
+  echo "Required permissions: read:user"
+  exit 1
+fi
+
+if [[ ! "$GITHUB_PAT" =~ ^ghp_ ]]; then
+  echo "Warning: GITHUB_PAT should start with 'ghp_' (Personal Access Token)"
+  echo "Current value starts with: ${GITHUB_PAT:0:4}..."
+  read -p "Continue anyway? (y/N) " -n 1 -r
```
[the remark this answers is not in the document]
maleck13: This is purely a demo script not intended to be used beyond that

jasonmadigan:
thanks @maleck13 giving it another spin

internal/config/types.go
```diff
@@ -67,6 +67,11 @@ func (config *MCPServersConfig) Notify(ctx context.Context) {
 	}
 }
 
+// GetExternalHostname returns the public hostname of the gateway
+func (config *MCPServersConfig) GetExternalHostname() string {
+	return config.MCPGatewayExternalHostname
+}
```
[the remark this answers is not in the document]
maleck13: MCPGatewayExternalHostname is set once at startup (line 246)
  before any goroutines read it, and is never written again

jasonmadigan:
steps and manually verified. will dispatch some 🤖 to see if we see any remaining issues.

internal/session/cache.go
```diff
@@ -133,6 +137,100 @@ func (c *Cache) GetClientElicitation(ctx context.Context, gatewaySessionID strin
...
+		next := maps.Clone(existing)
+		if next == nil {
+			next = map[string]string{}
+		}
+		next[field] = token
+		c.inmemory.Store(sessionID, next)
+		return nil
+	}
+	value := token
+	if c.encryptionKey != nil {
```
jasonmadigan, line 159: q: do we document that we only encrypt these in redis?

api/v1alpha1/types.go
jasonmadigan: nit: should probably set a MaxLength on this

jasonmadigan (dismissed):
this is in good shape

internal/session/cache.go
maleck13: Yes. There is little value in encrypting in memory as the key is also stored in memory and it need to be decrypted in memory anyway before sending it to the backend

jasonmadigan (approved):

PR #932, opened 2026-05-11
feat(broker): add /healthz endpoint and readiness initialDelaySeconds

Aman-Cool:
Two small gaps noticed while reviewing #848.

**`/healthz`** : the static manifest in `deployment-broker.yaml` already references it as a liveness probe path, but the broker HTTP server never registered it. Added a plain `200` handler in `setUpHTTPServer`. No logic, no state ; the process being alive is the only thing worth signalling here. Means the next person who wants to wire it to a liveness probe in `broker_router.go` doesn't need a code change on the broker side first.

**`initialDelaySeconds: 5`** : the controller-generated deployment in `broker_router.go` was building the readiness probe with no initial delay (defaulting to `0`), so Kubernetes starts polling `/readyz` the moment the container starts. With `failureThreshold: 3` × `periodSeconds: 10` there's 30 seconds of tolerance before the pod gets marked unready ; usually fine, but tight if upstream connections are slow. The static manifest already had `initialDelaySeconds: 5`; this just brings the controller-generated deployment in line. Updated the test baseline in `deployment_test.go` to match since `deploymentNeedsUpdate` uses `DeepEqual`.


Aman-Cool:
@david-martin, Follow-up from #848 ; two small gaps noticed during review.

`/healthz` was referenced in the manifest as a liveness probe path but never actually registered in the broker. `initialDelaySeconds` was missing from the controller-generated readiness probe while the static manifest already had it.  Just syncing things up.

david-martin (approved):
Thanks @Aman-Cool 

Good catch

Issue #916, opened 2026-05-08
Bug:  concurrent manage() calls cause unsynchronised access to serverTools


aniketpandey05:
## Bug

`manage()` can be called concurrently from two different goroutines:

1. The ticker loop in `Start()`  `case <-man.ticker.C`
2. The `OnNotification` callback in `registerCallbacks()`  this fires from mcp-go's streamable HTTP transport which dispatches notifications in its own 
goroutine (`streamable_http.go:445`)

When both fire at the same time, they race on `man.serverTools` specifically the writes at lines 269 and 273, and the read in `shouldFetchTools()` at 
line 289  none of which are covered by `toolsLock`.
 
code to test and reproduce the bug 

```go
func TestMCPManager_manage_ConcurrentRace(t *testing.T) {
	mock := newMockMCP("test-server", "test_")
	gateway := newThreadSafeMockToolsAdderDeleter()
	logger := slog.New(slog.NewTextHandler(os.Stdout, nil))
	manager := NewUpstreamMCPManager(mock, gateway, logger, 0, mcpv1alpha1.InvalidToolPolicyFilterOut)

	ctx, cancel := context.WithTimeout(context.Background(), 2*time.Second)
	defer cancel()

	var wg sync.WaitGroup
	for i := 0; i < 10; i++ {
		wg.Add(2)
		go func() {
			defer wg.Done()
			manager.manage(ctx, eventTypeTimer)
		}()
		go func() {
			defer wg.Done()
			manager.manage(ctx, eventTypeNotification)
		}()
	}
	wg.Wait()
}
```

I am considering adding a mutex around the entire `manage()` call to serialize execution so ticker and notification goroutines can't run it simultaneously 
happy to implement this if this fix is good , any suggestions

PR #936, opened 2026-05-12
fix: serialize concurrent manage calls in MCPManager

Aman-Cool:
The previous fixes had me reading through the broker and manager code longer than I planned. The comment on line 187 of `manager.go` says "manage should be the only entry point that triggers changes to tools"; it is, but that's not the same as being called from one goroutine.

`Start` drives the ticker loop and calls `manage` on every tick. `registerCallbacks` wires up `OnNotification`, which calls `manage` directly from the mcp-go SSE-reader goroutine whenever the backend sends `notifications/tools/list_changed`. Nothing serialises them.

The race is a read-compute-write: both callers snapshot `man.tools` under `RLock`, call `MCP.ListTools()` with the lock released, compute a diff against their snapshot, then take `toolsLock.Lock()` to commit. If the notification goroutine finishes first with a correct `servedToolsMap ; {A, B, C}`, and the timer goroutine, which started from a slightly staler snapshot of `[A, B]` ; acquires the write lock second, it overwrites: `man.tools = [A, B]`, `servedToolsMap = {A, B}`. C is gone from the lookup table even though the mcp-go gateway server has it registered.

What keeps it stuck: `shouldFetchTools` only lets timer ticks re-fetch when `len(man.serverTools) == 0`. After the race, `serverTools` has duplicates and is non-zero, so every subsequent tick returns early. No periodic recovery. The only escape is another `notifications/tools/list_changed` from the backend, which won't arrive if nothing else changes.

From a client's view: `tools/list` shows C, `tools/call` on C returns "Tool not found". Nothing in the logs surfaces the divergence between the gateway server's registry and `servedToolsMap`.

Fix is one mutex at the top of `manage`.

## Release Notes

* **Bug Fixes**
  * Enhanced stability of tool management operations by preventing conflicts between concurrent operations.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Kuadrant/mcp-gateway/pull/936)

### DIFF
--- a/internal/broker/upstream/manager.go
+++ b/internal/broker/upstream/manager.go
@@ -85,6 +85,8 @@ type MCPManager struct {
 	servedToolsMap map[string]*mcp.Tool
 	// toolsLock protects tools, serverTools
 	toolsLock sync.RWMutex
+	// manageMu serializes concurrent manage calls from the ticker and notification goroutines
+	manageMu sync.Mutex
 
 	logger *slog.Logger
 
@@ -186,6 +188,8 @@ func (man *MCPManager) registerCallbacks(ctx context.Context) func() {
 
 // manage should be the only entry point that triggers changes to tools
 func (man *MCPManager) manage(ctx context.Context, event eventType) {
+	man.manageMu.Lock()
+	defer man.manageMu.Unlock()
 	man.logger.Debug("managing connection", "upstream mcp server", man.MCP.ID(), "event type", event)
 	var numberOfTools = 0
 	// during connect the client will validate the protocol. So we don't have a separate validate requirement currently. If a client already exists it will be re-used.


Aman-Cool:
@maleck13, The only file touched is `manager.go` ; two lines added, no API changes. Worth checking `shouldFetchTools` while you're in there; that's the part that makes the stale state stick around rather than self-correcting on the next tick.

maleck13:
Appreciate the effort here, the window and opportunity for this to cause an issue is quite narrow, but it would difficult to debug if it were to happen 👍 . Each MCP gets its own manager routine and as mentioned the notification trigger comes from a separate routine so happy with the lock at the manage level not causing any issues.

maleck13 (approved):
I believe this closes #916

PR #937, opened 2026-05-12
test: add regression test for concurrent manage() data race

aniketpandey05:
Adds a regression test for the data race reported in #916 and fixed in #936.

The test calls `manage()` concurrently from multiple goroutines to simulate the ticker loop and notification callback firing simultaneously. Running with `go test -race` confirms the race, ensuring the fix in #936 won't regress.

The existing CI already runs with `-race` via `make test-unit`, so this test will be caught automatically on future PRs.

## Summary by CodeRabbit

* **Tests**
  * Added a concurrency/race-focused test to improve coverage around concurrent operations, increasing confidence in thread-safety and reliability under concurrent load.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Kuadrant/mcp-gateway/pull/937)

### DIFF
--- a/internal/broker/upstream/manager_test.go
+++ b/internal/broker/upstream/manager_test.go
@@ -986,3 +986,35 @@ func TestMCPManager_manage_AllValidTools(t *testing.T) {
 	assert.Empty(t, status.InvalidToolList)
 	assert.Len(t, gateway.tools, 2)
 }
+func TestMCPManager_manage_ConcurrentRace(t *testing.T) {
+	mock := newMockMCP("test-server", "test_")
+	gateway := NewMockGatewayServer()
+	logger := slog.New(slog.NewTextHandler(os.Stdout, nil))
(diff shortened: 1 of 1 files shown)


david-martin:
Thanks for this. A few things have changed on main since this was opened that you'll want to be aware of:

- #936 added `manageMu` to serialize concurrent `manage()` calls
- #943 refactored the manager to use a channel-based event loop, which changes how concurrency works in this area

This PR will need a rebase against main. There's still value in having a concurrent regression test for `manage()` to guard against future changes accidentally removing the serialization. After rebasing, please verify the test still exercises the right code paths given the new event loop design.

david-martin:

jasonmadigan:
thanks @rogueslasher, closing -- #943 refactored the manager to a channel-based event loop so manage() is only called from a single goroutine now, making this test moot.

PR #940, opened 2026-05-12
fix: prevent concurrent map write panic in Cache.AddSession and Remov…

### LINKED ISSUE #939: bug: concurrent map write panic in Cache.AddSession and RemoveServerSession
## Bug

`Cache.AddSession()` and `Cache.RemoveServerSession()` perform a read-modify-write on an inner `map[string]string` that is stored inside a `sync.Map`. While `sync.Map` protects individual `Load`/`Store` operations, it does NOT protect the sequence as a whole. Two concurrent calls can both load the same map, modify it, and store back — causing a concurrent map write panic.

This is directly reachable in production — `AddSession` is called from `request_handlers.go:581` and `RemoveServerSession` from `response_handlers.go:42`, both in concurrent request handling paths.

## Reproducer

```go
func TestInMemoryCache_AddSession_Concurrent(t *testing.T) {
	ctx := context.Background()
	cache, err := NewCache()
	require.NoError(t, err)

	var wg sync.WaitGroup
	for i := 0; i < 100; i++ {
		wg.Add(1)
		go func(i int) {
			defer wg.Done()
			serverName := fmt.Sprintf("server%d", i)
			_, err := cache.AddSession(ctx, "gateway-session-1", serverName, "upstream-session")
			require.NoError(t, err)
		}(i)
	}
	wg.Wait()
}
```

Running this panics with a fatal concurrent map write at `cache.go:71`.

## Suggested Fix

Add a `sync.Mutex` per gateway session key, or replace the inner `map[string]string` with a `sync.Map`, or use a single `sync.RWMutex` 
protecting the entire read-modify-write sequence in `AddSession` and `RemoveServerSession`.

Happy to make a pr 

@maleck13

aniketpandey05:
Fixes #939

Cache.AddSession() and Cache.RemoveServerSession() were performing a read-modify-write on an inner map[string]string stored inside a sync.Map. While sync.Map protects individual Load/Store calls, it does not protect the sequence as a whole two concurrent callers could both load the same map, modify it, and store back, causing a concurrent map write panic.

Fix adds a sync.Mutex to Cache to serialize the read-modify-write sequence in both methods for the in-memory path.

Verified with go test -race including a new concurrent reproducer test that previously panicked and now passes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved cache thread safety by adding proper synchronization to prevent race conditions during concurrent operations.

* **Tests**
  * Added test coverage for concurrent cache operations to validate stability under high concurrency.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Kuadrant/mcp-gateway/pull/940)

### DIFF
--- a/internal/session/cache.go
+++ b/internal/session/cache.go
@@ -14,6 +14,7 @@ const clientElicitationPrefix = "clientelicitation:"
 type Cache struct {
 	inmemory  *sync.Map
 	extClient *redis.Client
+	mu        sync.Mutex
 }
 
 // KeyExists checks if a key exists in the cache
@@ -64,6 +65,8 @@ func (c *Cache) DeleteSessions(ctx context.Context, key ...string) error {
 // AddSession will add a session under the key. If the key exists it will append that session
 func (c *Cache) AddSession(ctx context.Context, key, mcpServerID, mcpSession string) (bool, error) {
 	if c.inmemory != nil {
+		c.mu.Lock()
+		defer c.mu.Unlock()
 		session, err := c.GetSession(ctx, key)
 		if err != nil {
 			return false, err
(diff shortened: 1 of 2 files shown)


david-martin:
Thanks for looking into this. This race was already fixed in #872, which landed with a more comprehensive approach: copy-on-write semantics for the inner map and mutex protection on AddSession, RemoveServerSession, and DeleteSessions. Closing as superseded.

PR #941, opened 2026-05-12
fix: clean up elicitation idmap entries on interrupted ext_proc streams

shiavm006:
While reading through the elicitation flow I noticed the cleanup contract in `docs/design/notifications.md:339`:

> Mappings persist for the tool call duration and are removed when the response stream ends. In Redis-backed deployments, a 1-hour TTL acts as a safety net for orphaned entries.

The "removed when the response stream ends" half doesn't hold for all the ways a stream can end.

In `internal/mcp-router/server.go:Process()`, `rewriter.Flush(ctx)` is called from exactly one site — line 285, inside `ProcessingRequest_ResponseBody` when `endOfStream` is true. After the rewriter is created at line 253 there are several `return err` paths that bypass it:

- `stream.Recv()` error on the next iteration (client disconnect / Envoy stream close)
- `stream.Send` error on the routing response
- `stream.Send` error on the response body

Each abandoned stream then leaks N entries in `ElicitationMap`, where N is the number of `elicitation/create` messages already buffered before the stream died.

For the in-memory `idmap` (`internal/idmap/inmemory.go`) this is unbounded — unlike `redis.go` it has no `entryTTL`, no safety net. Memory grows with streaming-tool-call traffic that includes elicitations. The Redis-backed deployments still hold the orphaned entries until the 1-hour TTL kicks in.

The fix is two small pieces: a function-level `defer` in `Process` that calls `rewriter.Flush` on any exit path, plus making `Flush` idempotent (clear `w.gatewayIDs` after the Remove loop) so the happy-path call followed by the defer is a clean no-op.

Added a subtest under `TestSSERewriter_Flush` covering the idempotency contract. All `internal/` tests pass with `-race`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved resource cleanup during stream processing errors. The system now properly handles orphaned entries when stream receive/send errors occur, ensuring cleanup happens regardless of exit conditions. Made resource cleanup operations idempotent to prevent duplicate processing attempts.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Kuadrant/mcp-gateway/pull/941)

### DIFF
--- a/internal/mcp-router/elicitation.go
+++ b/internal/mcp-router/elicitation.go
@@ -58,12 +58,14 @@ func (w *sseRewriter) Process(ctx context.Context, chunk []byte) []byte {
 
 // Flush flushes the buffer and cleans up any gatewayIDs created by the rewriter
 // This allows us to deal with orphaned elicitation id mappings
+// Safe to call multiple times; subsequent calls are no-ops
 func (w *sseRewriter) Flush(ctx context.Context) []byte {
 	remaining := w.buf
 	w.buf = nil
 	for _, id := range w.gatewayIDs {
 		w.idMap.Remove(ctx, id) // tool request + response finished, no need to hold onto the mappings any more
 	}
+	w.gatewayIDs = nil
 	return remaining
 }
 
--- a/internal/mcp-router/elicitation_test.go
+++ b/internal/mcp-router/elicitation_test.go
@@ -269,6 +269,36 @@ func TestSSERewriter_Flush(t *testing.T) {
 		flushed := w.Flush(ctx)
 		require.Nil(t, flushed)
 	})
+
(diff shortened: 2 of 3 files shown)


shiavm006:
@maleck13  spotted this while reading the elicitation cleanup path. Reasonably confident it's a real leak in the in memory variant since `inmemory.go` has no TTL safety net, but happy to add an integration-level test if you'd prefer that over the idempotency unit test. No rush given the review load.

shiavm006:
cc @david-martin

david-martin:

jasonmadigan (approved):
lgtm. defer catches all exit paths, Flush idempotency is well handled (nil gatewayIDs after cleanup). good catch from the design doc.

Issue #659, opened 2026-03-12
Custom TLS configuration for broker-to-backend connections


jasonmadigan:
**Is your feature request related to a problem? Please describe.**

The MCP broker connects to upstream MCP servers using Go's default HTTP transport, which only trusts publicly-trusted CAs. This works for external services like GitHub's MCP server and in-cluster MCP servers with valid public certificates (#450 verified this, including full stack on OpenShift 4.20 with Let's Encrypt wildcard certs), but in-cluster MCP servers can often use private CAs (e.g. OpenShift service-serving CA, cert-manager, self-signed) which the broker rejects with a certificate verification error. There's no way to provide a custom CA bundle, override the TLS server name, or skip verification for development.

**Describe the solution you'd like**

I am not firm on an API, but something that exposes options like below that configures how the broker connects to HTTPS backends would be welcome:

```yaml
spec:
  tls:
    caCertSecretRef:
      name: my-ca-bundle    # Secret containing CA PEM
    insecureSkipVerify: false # maybe? not sure we need this if we can inject bundles tbh
```

This only affects the broker path (init, tools/list). The tool call path flows through Envoy.

### Open questions

- **`insecureSkipVerify` -- is it needed?** If providing a CA cert in a Secret is sufficient for development, skip-verify feels redundant?
- **`serverName` override -- defer?** Could be useful for SNI routing or IP-only endpoints, but niche and bypasses hostname verification. Bit of a shotgun otherwise.
- **Policy separation.** MCPServerRegistration already combines backend declaration + credential reference. Adding TLS override means one RBAC grant covers all three. Worth considering whether TLS config should live on a separate policy resource imho

**Describe alternatives you've considered**

- Inlining CA cert PEM in the MCPServerRegistration spec: simpler but doesn't support rotation without re-applying the CR, and bloats the resource
- Mounting CA bundles into the broker pod via volume mounts: works but requires Helm chart changes per-cluster and doesn't allow per-backend CA config?

**Additional context**

- #450 confirmed that HTTPS works today for publicly-trusted CAs (verified with GitHub MCP server on Kind and full stack on OpenShift 4.20 with Let's Encrypt)

david-martin:
Suggestion from @maleck13 and @jasonmadigan , let's *not* add the `insecureSkipVerify` flag.
The secretRef can be mounted into the right pod, and the client should use it (if it's on the right path)

justin212407:
[...]

Issue #450, opened 2025-12-17
Testing and defining support HTTPS for MCP Server backends


david-martin:
Jira: https://issues.redhat.com/browse/CONNLINK-813

Although there is provision for using a protocol of https in an MCP server endpoint config (https://github.com/Kuadrant/mcp-gateway/blob/9a1f463a9dd3fa38d95162616e4e62e90200aa9b/pkg/controller/mcpserver_controller.go#L523), this hasn't been tested and verified as working.

Also, there is no provision for adding a custom CA, or allowing insecure servers for development, when connecting to a HTTPS enabled MPC Server.

In general, the story for HTTPS MCP Servers 'in cluster' is not well defined.

We should look to support HTTPS servers, and add e2e tests for them. 
It should be possible for users to configure the MCP Gateway (i.e. the mcp-broker) with a CA for use when doing an init/tool list with backend MCP Servers.
We should also consider supporting an insecure flag in the MCPServer resource, mainly for development.

For tool calls, this shouldn't be a concern as requests don't hit the broker, only the router.
For that scenario, the https configuration should be handled at the Gateway & HTTPRoute, possibly via BackendTLSPolicy.

jasonmadigan:
Have done the initial part of this issue, checking that HTTPS backend MCP servers in cluster actually work with the broker (was untested - they work fine).

Some steps for others if we need to verify again:

# MCP Gateway and in-cluster, HTTPS backends

Verifies that the MCP broker works end-to-end over TLS: client connects via HTTPS, initialises a session, lists federated tools, and calls a tool routed. I tested with a Prow cluser (which uses a real LE cert on a ROSA cluster).

## Setup & Deploy

```bash
bash config/openshift/deploy_openshift.sh

# test servers
kubectl apply -k config/test-servers/
kubectl get deploy -n mcp-test -o name | while read d; do
  name=$(kubectl get "$d" -n mcp-test -o jsonpath='{.spec.template.spec.containers[0].name}')
  kubectl patch "$d" -n mcp-test -p "{\"spec\":{\"template\":{\"spec\":{\"containers\":[{\"name\":\"$name\",\"imagePullPolicy\":\"IfNotPresent\"}]}}}}"
done
kubectl wait --for=condition=Available deployment -n mcp-test --all --timeout=120s
kubectl apply -f config/samples/mcpserverregistration-test-servers-base.yaml

# broker HTTPRoute (not created by Helm chart)
[...]

PR #942, opened 2026-05-12
fix(router): use https scheme for hair-pinning when listener is HTTPS

### LINKED ISSUE #917: Router hairpin uses wrong scheme/port for HTTPS Gateway listeners
## Bug

When the Gateway listener uses HTTPS on port 443, the router constructs the hairpin URL with the `http://` scheme and port 443:

```
http://my-gateway-istio.gateway-system.svc.cluster.local:443/mcp
```

This fails because the Envoy service expects TLS on that port:

```
failed to create client: transport error: failed to send request: Post "http://...svc.cluster.local:443/mcp": context canceled
```

### Root Cause

Two things combine to produce this:

1. [`InternalHost()`](https://github.com/Kuadrant/mcp-gateway/blob/41aefb09/api/v1alpha1/mcpgatewayextension_types.go#L235-L244) uses the Gateway listener port directly — for an HTTPS listener, that's 443.

2. [`internal/clients/clients.go:30`](https://github.com/Kuadrant/mcp-gateway/blob/41aefb09/internal/clients/clients.go#L30) hardcodes the `http://` scheme:
   ```go
   url := fmt.Sprintf("http://%s%s", gatewayHost, mcpPath)
   ```

The result is plain HTTP to a TLS-only port.

### How to Reproduce

Create a Gateway with an HTTPS listener and route MCP traffic through it:

```yaml
apiVersion: gateway.networking.k8s.io/v1
kind: Gateway
metadata:
  name: my-gateway
  namespace: gateway-system
spec:
  gatewayClassName: example
  listeners:
    - name: mcp
      hostname: mcp.example.com
      port: 443
      protocol: HTTPS
      tls:
        mode: Terminate
        certificateRefs:
          - kind: Secret
            name: gateway-tls
      allowedRoutes:
        namespaces:
          from: All
    - name: backend-mcp
      hostname: backend-mcp.example.com
      port: 443
      protocol: HTTPS
      tls:
        mode: Terminate
        certificateRefs:
          - kind: Secret
            name: gateway-tls
      allowedRoutes:
        namespaces:
          from: All
```

```yaml
apiVersion: gateway.networking.k8s.io/v1
kind: HTTPRoute
metadata:
  name: mcp-gateway-route
  namespace: mcp-system
spec:
  parentRefs:
    - name: my-gateway
      namespace: gateway-system
      sectionName: mcp
  hostnames:
    - mcp.example.com
  rules:
    - matches:
        - path:
            type: PathPrefix
            value: /mcp
      backendRefs:
        - name: mcp-gateway
          port: 8080
```

```yaml
apiVersion: gateway.networking.k8s.io/v1
kind: HTTPRoute
metadata:
  name: backend-server-route
  namespace: mcp-system
  labels:
    mcp-server: "true"
spec:
  parentRefs:
    - name: my-gateway
      namespace: gateway-system
      sectionName: backend-mcp
  hostnames:
    - backend-mcp.example.com
    - my-gateway-istio.gateway-system.svc.cluster.local
  rules:
    - matches:
        - path:
            type: PathPrefix
            value: /mcp
      backendRefs:
        - name: backend-service
          namespace: apps
          port: 8080
```

```yaml
apiVersion: mcp.kuadrant.io/v1alpha1
kind: MCPGatewayExtension
metadata:
  name: mcp-gateway
  namespace: mcp-system
spec:
  targetRef:
    kind: Gateway
    name: my-gateway
    namespace: gateway-system
    sectionName: mcp
  publicHost: mcp.example.com
  httpRouteManagement: Disabled
```

A `tools/call` request will fail with the `context canceled` error above.

### Expected Behavior

The hairpin URL should use a scheme that matches the listener protocol — either `https://...:443` or `http://` with an appropriate HTTP port.

### Workaround

Set [`spec.privateHost`](https://github.com/Kuadrant/mcp-gateway/blob/41aefb09/api/v1alpha1/mcpgatewayextension_types.go#L236-L238) on the MCPGatewayExtension to override the computed internal host with an address that accepts plain HTTP. For example, if the Envoy service also exposes an HTTP port internally:

```yaml
apiVersion: mcp.kuadrant.io/v1alpha1
kind: MCPGatewayExtension
metadata:
  name: mcp-gateway
  namespace: mcp-system
spec:
  privateHost: "my-gateway-istio.gateway-system.svc.cluster.local:8080"
  targetRef:
    kind: Gateway
    name: my-gateway
    namespace: gateway-system
    sectionName: mcp
```

This only works if the Envoy service has an HTTP listener available on that port. If the Gateway only has HTTPS listeners, adding an HTTP listener on a separate port is needed to unblock the hairpin path.

See the [`privateHost` API reference](https://github.com/Kuadrant/mcp-gateway/blob/41aefb09/docs/reference/mcpgatewayextension.md) for details.

Bhuvanesh66:
## Summary

Fixes #917.

When the Gateway listener uses `HTTPS` on port `443`, the broker hair-pins the
backend `initialize` request through Envoy using the URL it gets via
`--mcp-gateway-private-host`. That URL was hard-coded to `http://` in
`internal/clients/clients.go`, while the controller passed only the
`host:port` (with `port=443` for an HTTPS listener). The broker therefore sent
plain HTTP to a TLS-only port:

This PR makes `--mcp-gateway-private-host` scheme-aware following the
implementation approach suggested by @david-martin in the issue: plumb the
listener protocol through `ListenerConfig`, let the controller decide the
scheme, and have the broker's URL builder respect an explicit scheme prefix
when it sees one.

Related: #450, #659.

## Root cause

Two things combined:

1. `internalHost()` (controller) uses the Gateway listener port directly, so
   an HTTPS listener emits `host:443`.
2. `buildHairpinURL` (broker client) prepended `http://` unconditionally.

Result: HTTP request hitting Envoy on the TLS port, which never responds with
HTTP, and the broker context times out.

## What this PR does

- **`api/v1alpha1/mcpgatewayextension_types.go`**
  Added a `Protocol` field to the internal `ListenerConfig` so downstream code
  can branch on `HTTP` vs `HTTPS`. Updated the doc comment on
  `MCPGatewayExtensionSpec.PrivateHost` to describe the new scheme behaviour
  and the override semantics (operator-supplied value is honoured verbatim).
- **`internal/controller/mcpgatewayextension_controller.go`**
  `findListenerConfigByName` now populates `ListenerConfig.Protocol` from
  `gatewayv1.Listener.Protocol`.
- **`internal/controller/broker_router.go`**
  New `derivePrivateHost(mcpExt, listenerConfig)` helper, called from
  `reconcileBrokerRouter`:
  - If `spec.privateHost` is set, pass it through unchanged (so users keep
    full control of scheme/host/port and can override either direction).
  - Else, if `listenerConfig.Protocol == "HTTPS"` (case-insensitive), prepend
    `https://` to the computed internal host.
  - Else, leave it bare (no scheme), preserving today's behaviour for HTTP
    listeners.
- **`internal/clients/clients.go`**
  Factored the URL composition into `buildHairpinURL(gatewayHost, mcpPath)`:
  - If `gatewayHost` already has an `http://` / `https://` prefix, use it
    as-is.
  - Otherwise fall back to `http://` (preserves the current behaviour for
    bare hosts -> backwards compatible).
- **Docs**
  `docs/reference/mcpgatewayextension.md` updated to match the new
  `privateHost` semantics.

## Backwards compatibility

- HTTP listeners produce the same `--mcp-gateway-private-host` value as
  before (no scheme prefix). No effect on existing HTTP-only deployments.
- Existing `spec.privateHost` overrides are honoured verbatim and never
  silently rewritten.
- The `--mcp-gateway-private-host` flag surface is unchanged; the controller
  just emits an `https://`-prefixed value when appropriate, and the broker's
  URL builder accepts either form.
- Operators with a private-CA HTTPS listener will still need #659 (custom CA
  bundle for the broker) for the TLS handshake to succeed. After this PR they
  get a clear cert-validation error instead of the misleading
  `context canceled` symptom, which makes #659 actionable.

## Out of scope

- Custom CA bundle / `insecureSkipVerify` for backend TLS — tracked in #659.

## Test plan

Automated (run locally on Windows):

- [x] `go build ./...` clean
- [x] `go vet ./...` clean
- [x] `go test -race -count=1` on
      `./internal/clients/...`, `./internal/controller/...`,
      `./api/v1alpha1/...`, `./internal/mcp-router/...`,
      `./internal/session/...`, `./internal/broker/...`,
      `./internal/config/...` — all `ok`
- [x] New unit tests:
  - `TestBuildHairpinURL` — bare host, `https://` prefix, `http://` prefix,
    custom path
  - `TestDerivePrivateHost` — HTTP listener, HTTPS listener, mixed-case
    protocol, `PrivateHost` override with and without scheme, TCP fallback
  - `TestFindListenerConfig` extended to assert `Protocol` is populated

Manual / e2e (suggested for reviewers, since I can't reproduce the OpenShift
setup locally):

- [ ] Apply the reproducer manifest from #917 (Gateway with HTTPS listener on
      `:443`). Confirm `kubectl get deploy mcp-gateway -o yaml` now shows
      `--mcp-gateway-private-host=https://my-gateway-istio.gateway-system.svc.cluster.local:443`
      and that `tools/call` succeeds.
- [ ] Confirm an HTTP-listener deployment yields the unchanged
      (no-scheme-prefix) value.
- [ ] Confirm a deployment with `spec.privateHost` set is unchanged.

---

cc @david-martin (thanks for the implementation pointer in #917) and
@maleck13 for visibility.

## Summary by CodeRabbit

* **New Features**
  * `privateHost` now accepts full URLs (explicit schemes like https://) and custom ports for gateway configuration.

* **Bug Fixes**
  * Correctly preserves/chooses HTTP vs HTTPS in generated hairpin URLs for gateway listeners.

* **Documentation**
  * Clarified that `privateHost` is honored verbatim, including schemes and ports; CRD docs updated.

* **Tests**
  * Added coverage for URL construction and private-host derivation to ensure correct behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Kuadrant/mcp-gateway/pull/942)

### DIFF
--- a/internal/controller/broker_router.go
+++ b/internal/controller/broker_router.go
@@ -279,13 +279,35 @@ func derivePublicHost(listenerConfig *mcpv1alpha1.ListenerConfig, annotationOver
 	return hostname, nil
 }
 
+// derivePrivateHost returns the value passed to --mcp-gateway-private-host,
+// which the broker uses when hairpinning the internal initialize request back
+// through the gateway. When the user explicitly sets spec.privateHost, that
+// value is honoured verbatim (so an operator can override scheme, host, and
+// port). Otherwise the host is computed from the targetRef and listener port,
+// and an https:// scheme prefix is added when the listener is HTTPS so the
+// broker hairpin doesn't send plain HTTP to a TLS-only port (issue #917).
+func derivePrivateHost(mcpExt *mcpv1alpha1.MCPGatewayExtension, listenerConfig *mcpv1alpha1.ListenerConfig) string {
+	host := mcpExt.InternalHost(listenerConfig.Port)
+	// honour an operator-supplied PrivateHost as-is. They may already have
+	// embedded a scheme, or chosen a separate HTTP port, and we should not
+	// silently rewrite either.
+	if mcpExt.Spec.PrivateHost != "" {
+		return host
+	}
+	// listener.Protocol is the Gateway API protocol string, e.g. "HTTPS".
+	if listenerConfig != nil && strings.EqualFold(listenerConfig.Protocol, "HTTPS") {
+		return "https://" + host
+	}
+	return host
+}
+
 func (r *MCPGatewayExtensionReconciler) reconcileBrokerRouter(ctx context.Context, mcpExt *mcpv1alpha1.MCPGatewayExtension, listenerConfig *mcpv1alpha1.ListenerConfig) (bool, error) {
 	// derive values from listener config before building resources
 	publicHost, err := derivePublicHost(listenerConfig, mcpExt.Spec.PublicHost)
 	if err != nil {
 		return false, newValidationError(mcpv1alpha1.ConditionReasonInvalid, err.Error())
 	}
-	internalHost := mcpExt.InternalHost(listenerConfig.Port)
+	internalHost := derivePrivateHost(mcpExt, listenerConfig)
 
 	// reconcile service account (must exist before deployment)
 	serviceAccount := r.buildBrokerRouterServiceAccount(mcpExt)
--- a/internal/controller/broker_router.go
+++ b/internal/controller/broker_router.go
@@ -287,13 +287,10 @@ func derivePublicHost(listenerConfig *mcpv1alpha1.ListenerConfig, annotationOver
 // and an https:// scheme prefix is added when the listener is HTTPS so the
 // broker hairpin doesn't send plain HTTP to a TLS-only port (issue #917).
 func derivePrivateHost(mcpExt *mcpv1alpha1.MCPGatewayExtension, listenerConfig *mcpv1alpha1.ListenerConfig) string {
-	host := mcpExt.InternalHost(listenerConfig.Port)
-	// honour an operator-supplied PrivateHost as-is. They may already have
-	// embedded a scheme, or chosen a separate HTTP port, and we should not
-	// silently rewrite either.
 	if mcpExt.Spec.PrivateHost != "" {
-		return host
+		return mcpExt.Spec.PrivateHost
 	}
+	host := mcpExt.InternalHost(listenerConfig.Port)
 	// listener.Protocol is the Gateway API protocol string, e.g. "HTTPS".
 	if listenerConfig != nil && strings.EqualFold(listenerConfig.Protocol, "HTTPS") {
 		return "https://" + host
--- a/internal/controller/broker_router.go
+++ b/internal/controller/broker_router.go
@@ -292,7 +292,7 @@ func derivePrivateHost(mcpExt *mcpv1alpha1.MCPGatewayExtension, listenerConfig *
 	}
 	host := mcpExt.InternalHost(listenerConfig.Port)
 	// listener.Protocol is the Gateway API protocol string, e.g. "HTTPS".
-	if listenerConfig != nil && strings.EqualFold(listenerConfig.Protocol, "HTTPS") {
+	if strings.EqualFold(listenerConfig.Protocol, "HTTPS") {
 		return "https://" + host
 	}
 	return host
--- a/api/v1alpha1/mcpgatewayextension_types.go
+++ b/api/v1alpha1/mcpgatewayextension_types.go
@@ -64,7 +64,11 @@ type MCPGatewayExtensionSpec struct {
 	PublicHost string `json:"publicHost,omitempty"`
 
 	// privateHost overrides the internal host used for hair-pinning requests
-	// back through the gateway. Defaults to <gateway>-istio.<ns>.svc.cluster.local:<port>.
+	// back through the gateway. Defaults to
+	// <gateway>-istio.<ns>.svc.cluster.local:<port>, with an https:// scheme
+	// prefix when the targeted Gateway listener uses the HTTPS protocol. The
+	// value supplied here is honoured verbatim, so an operator can include a
+	// scheme (e.g. "https://my-gw:443") or pin to a different port.
 	// +optional
 	PrivateHost string `json:"privateHost,omitempty"`
 
@@ -257,4 +261,8 @@ type ListenerConfig struct {
 	Hostname string `json:"hostname,omitempty"`
 	// name is the listener name (sectionName)
 	Name string `json:"name,omitempty"`
+	// protocol is the Gateway listener protocol (e.g. HTTP, HTTPS).
+	// Used to determine whether the broker-router hairpin URL should use
+	// http:// or https://.
+	Protocol string `json:"protocol,omitempty"`
 }
--- a/bundle/manifests/mcp.kuadrant.io_mcpgatewayextensions.yaml
+++ b/bundle/manifests/mcp.kuadrant.io_mcpgatewayextensions.yaml
@@ -78,7 +78,11 @@ spec:
               privateHost:
                 description: |-
                   privateHost overrides the internal host used for hair-pinning requests
-                  back through the gateway. Defaults to <gateway>-istio.<ns>.svc.cluster.local:<port>.
+                  back through the gateway. Defaults to
+                  <gateway>-istio.<ns>.svc.cluster.local:<port>, with an https:// scheme
+                  prefix when the targeted Gateway listener uses the HTTPS protocol. The
+                  value supplied here is honoured verbatim, so an operator can include a
+                  scheme (e.g. "https://my-gw:443") or pin to a different port.
                 type: string
               publicHost:
                 description: |-
@@ -247,9 +251,3 @@ spec:
     storage: true
     subresources:
       status: {}
-status:
-  acceptedNames:
-    kind: ""
-    plural: ""
-  conditions: null
-  storedVersions: null
--- a/bundle/manifests/mcp.kuadrant.io_mcpgatewayextensions.yaml
+++ b/bundle/manifests/mcp.kuadrant.io_mcpgatewayextensions.yaml
@@ -251,3 +251,9 @@ spec:
     storage: true
     subresources:
       status: {}
+status:
+  acceptedNames:
+    kind: ""
+    plural: ""
+  conditions: null
+  storedVersions: null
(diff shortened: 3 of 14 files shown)


david-martin (dismissed):
@Bhuvanesh66 The change looks clean. Thanks for taking this on.

I think we should look to cover the https listener scenario in e2e tests, but doing that for #659 makes sense rather than in this change.

@maleck13 before merging, are you happy with this fix approach as well?

internal/controller/broker_router.go
```diff
@@ -285,13 +271,35 @@ func derivePublicHost(listenerConfig *mcpv1alpha1.ListenerConfig, annotationOver
...
 
+// derivePrivateHost returns the value passed to --mcp-gateway-private-host,
+// which the broker uses when hairpinning the internal initialize request back
+// through the gateway. When the user explicitly sets spec.privateHost, that
+// value is honoured verbatim (so an operator can override scheme, host, and
+// port). Otherwise the host is computed from the targetRef and listener port,
+// and an https:// scheme prefix is added when the listener is HTTPS so the
+// broker hairpin doesn't send plain HTTP to a TLS-only port (issue #917).
+func derivePrivateHost(mcpExt *mcpv1alpha1.MCPGatewayExtension, listenerConfig *mcpv1alpha1.ListenerConfig) string {
+	host := mcpExt.InternalHost(listenerConfig.Port)
```
maleck13: This already defaults to the private host. So we can change this to

``` go
func derivePrivateHost(mcpExt *mcpv1alpha1.MCPGatewayExtension, listenerConfig *mcpv1alpha1.ListenerConfig) string {
      if mcpExt.Spec.PrivateHost != "" {
          return mcpExt.Spec.PrivateHost
      }
      host := mcpExt.InternalHost(listenerConfig.Port)
      if listenerConfig != nil && strings.EqualFold(listenerConfig.Protocol, "HTTPS") {
          return "https://" + host
      }
      return host
  }
```

Bhuvanesh66:
Hi @david-martin and @maleck13 — I want to be transparent about where I am stuck on the bundle check.
What is actually failing
check-bundle-crd-sync runs make bundle and then diffs the result against what is committed. The CRD sync itself is passing (✅ CRDs are synchronized) — the only remaining failure is that the committed bundle/ tree does not match a fresh make bundle output. Specifically two files are consistently drifting:

bundle/manifests/mcp-gateway.clusterserviceversion.yaml — contains createdAt timestamp churn
bundle/metadata/dependencies.yaml — real content drift, not just a timestamp

The Makefile's bundle-ignore-createdAt step only suppresses lines matching ^ createdAt:. Because dependencies.yaml also differs, the ignore rule does not fire, git checkout ./bundle does not run, and the tree still differs from the index — so the check fails.
What blocked me from fixing it cleanly
I tried fixing this both in GitHub Codespaces and through Docker and hit the same set of environment issues each time:

Windows + Docker mount — tools under bin/ like yq were sometimes Windows-built binaries on the shared volume. Inside Linux containers they either do not run or behave oddly, causing make generate-all and sync-helm-rbac to fail until they were removed and reinstalled inside the container.
CRLF line endings — hack/sync-helm-rbac.sh had Windows CRLF line endings which produced bash\r / env: bash\r failures in Linux until normalized.
Partial commits — bundle-ignore-createdAt makes it easy to mistake timestamp-only noise for real drift. Committing only part of bundle/ like the CSV but not dependencies.yaml means the check still fails.
Push authentication — even when commits were created locally in Docker, pushing to my fork required my GitHub auth which could not be completed reliably from inside the container environment.

I think below idea would fix it:
A single commit that includes every file make bundle changes, particularly bundle/metadata/dependencies.yaml and the CSV, run with:

```bash
make SHELL=/bin/bash bundle
git add bundle/
git commit -s -m "chore: sync OLM bundle"
git push
```

Would either of you be able to help unblock this?
Could you advise whether this is a known drift issue with dependencies.yaml in this repo, or whether it would be easier for you to run make bundle and push the result? I don't want to keep dismissing your review with noisy commits that don't actually resolve it. Happy to do whatever helps most.

david-martin:
@Bhuvanesh66 It looks like the bundle issue is sorted already on this PR branch?
`make check-bundle-crd-sync` exits 0 for me locally.

The only items left are the comment about the private host, and a rebase from main.

Bhuvanesh66:
Hi @david-martin , @maleck13  — rebased from main, applied the derivePrivateHost simplification @maleck13 suggested, and addressed inline comment on the header filtering. Ready for re-review when you get a chance.

internal/controller/broker_router.go
```diff
@@ -285,13 +285,32 @@ func derivePublicHost(listenerConfig *mcpv1alpha1.ListenerConfig, annotationOver
...
+// port). Otherwise the host is computed from the targetRef and listener port,
+// and an https:// scheme prefix is added when the listener is HTTPS so the
+// broker hairpin doesn't send plain HTTP to a TLS-only port (issue #917).
+func derivePrivateHost(mcpExt *mcpv1alpha1.MCPGatewayExtension, listenerConfig *mcpv1alpha1.ListenerConfig) string {
+	if mcpExt.Spec.PrivateHost != "" {
+		return mcpExt.Spec.PrivateHost
+	}
+	host := mcpExt.InternalHost(listenerConfig.Port)
+	// listener.Protocol is the Gateway API protocol string, e.g. "HTTPS".
+	if listenerConfig != nil && strings.EqualFold(listenerConfig.Protocol, "HTTPS") {
```
david-martin: We can drop the nil check here.
Already accessing listenerConfig above, and it's checked earlier in the stack (validateGatewayTarget)

david-martin:
Thanks @Bhuvanesh66 

We're nearly there. 1 more small thing and this should be good to merge.

david-martin:
Apologies for the churn here @Bhuvanesh66 
There has been a number of PRs landing today after updates and reviews.
The gevals CI has been removed on main, so the change for echo.yaml is no longer needed. Another rebase should sort that out.

david-martin (approved):
LGTM - fix correctly handles HTTPS scheme for hairpin initialization. Tests pass.

PR #943, opened 2026-05-12
Refactor/manager channel event loop

maleck13:
**Summary**

Move to channel based synchronisation rather than mutex. Based on recent PRs , I was prompted to re-think how the manager behaves. Not code I had looked at for some time. But looking at it fresh, decided to make a change to use channels rather than mutex on the manger.

- Uses a buffer to manage if multiple notifications arrive at once rather than queuing many calls to manager behind the lock
- All mutation flows through one goroutine's select loop now. 
- toolsLock only exists for external readers
- Aligns the pattern for notification with the ticker that also uses a channel
- Sending on the channel, ensures notification handling is none blocking for the caller

## Summary by CodeRabbit

* **Refactor**
  * Manager startup is now asynchronous and returns an active server handle with explicit Stop/inspect methods; teardown and event processing moved to an internal event loop for safer concurrency and shutdown.

* **Tests**
  * Added tests covering notification-driven updates, concurrent reads during updates, and stop-during-update shutdown races.

* **Documentation**
  * Added a “Concurrency” guideline recommending channel-based coordination before defaulting to mutexes.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Kuadrant/mcp-gateway/pull/943)

### DIFF
--- a/internal/broker/upstream/manager.go
+++ b/internal/broker/upstream/manager.go
@@ -93,10 +93,10 @@ type MCPManager struct {
 
 	// events funnels notifications into the Start() loop. Buffer of 1 coalesces
 	// rapid notifications; safe because manage() always does a full tool sync.
-	events   chan eventType
-	stopOnce sync.Once     // ensures Stop() is only executed once
-	done     chan struct{} // triggers the exit of the select and routine
-	status   ServerValidationStatus
+	events chan eventType
+	cancel context.CancelFunc // cancels the internal context to stop the event loop
+	done   chan struct{}      // closed when the event loop exits
+	status ServerValidationStatus
 }
 
 // DefaultTickerInterval is the default interval for backend health checks
@@ -138,39 +138,41 @@ func (man *MCPManager) MCPName() string {
 // registers notification callbacks to handle tool list changes. This method blocks
 // until Stop is called or the context is cancelled.
 func (man *MCPManager) Start(ctx context.Context) {
+	ctx, man.cancel = context.WithCancel(ctx)
 	man.ticker.Reset(man.tickerInterval)
 	man.manage(ctx, eventTypeTimer)
 
 	for {
 		select {
 		case <-ctx.Done():
-			man.Stop()
+			man.ticker.Stop()
+			if err := man.MCP.Disconnect(); err != nil {
+				man.logger.Error("failed to disconnect during stop", "upstream mcp server", man.MCP.ID(), "error", err)
+			}
+			man.removeAllTools()
+
+			close(man.done)
+			man.logger.Debug("manager stopped", "upstream mcp server", man.MCP.ID())
+			return
 		case <-man.ticker.C:
 			man.logger.Debug("health check tick", "upstream mcp server", man.MCP.ID())
 			man.manage(ctx, eventTypeTimer)
 		case evt := <-man.events:
 			man.logger.Debug("received event", "upstream mcp server", man.MCP.ID(), "event", evt)
 			man.manage(ctx, evt)
-		case <-man.done:
-			man.logger.Debug("shutting down manager", "upstream mcp server", man.MCP.ID())
-			return
 		}
 	}
 }
 
-// Stop gracefully shuts down the manager. It stops the ticker, removes all tools
-// from the gateway, disconnects from the upstream server, and waits for the Start
-// goroutine to complete. Safe to call multiple times.
+// Stop cancels the event loop and waits for teardown to complete.
+// All cleanup (remove tools, disconnect) happens on the event loop goroutine.
+// Safe to call before Start (no-op) and multiple times.
 func (man *MCPManager) Stop() {
-	man.stopOnce.Do(func() {
-		man.ticker.Stop()
-		man.removeAllTools()
-		if err := man.MCP.Disconnect(); err != nil {
-			man.logger.Error("failed to disconnect during stop", "upstream mcp server", man.MCP.ID(), "error", err)
-		}
-		close(man.done)
-		man.logger.Debug("manager stopped", "upstream mcp server", man.MCP.ID())
-	})
+	if man.cancel == nil {
+		return
+	}
+	man.cancel()
+	<-man.done
 }
 
 func (man *MCPManager) registerCallbacks() func() {
--- a/internal/broker/upstream/manager.go
+++ b/internal/broker/upstream/manager.go
@@ -149,9 +149,9 @@ func (man *MCPManager) MCPName() string {
 func (man *MCPManager) Start(ctx context.Context) ActiveMCPServer {
 	ctx, cancel := context.WithCancel(ctx)
 	man.ticker.Reset(man.tickerInterval)
-	man.manage(ctx, eventTypeTimer)
 
 	go func() {
+		man.manage(ctx, eventTypeTimer)
 		for {
 			select {
 			case <-ctx.Done():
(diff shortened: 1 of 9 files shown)


maleck13:
@jasonmadigan @Aman-Cool After some thought and PRs around this area. I decided to move to a channel rather than the mutex.

internal/broker/upstream/manager.go
```diff
@@ -255,31 +261,27 @@ func (man *MCPManager) manage(ctx context.Context, event eventType) {
...
+	man.serverTools = append(man.serverTools, toAdd...)
+	man.toolsLock.Unlock()
+
 	man.logger.Debug("updating gateway tools", "upstream mcp server", man.MCP.ID(), "adding", len(toAdd), "removing", len(toRemove))
 	if len(toRemove) > 0 {
 		man.gatewayServer.DeleteTools(toRemove...)
 	}
 	if len(toAdd) > 0 {
 		man.gatewayServer.AddTools(toAdd...)
 	}
```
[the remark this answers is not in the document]
maleck13: The window is tiny here and wont cause an significant impact:

Add direction: servedToolsMap has the new tool, but the gateway hasn't
  registered it. Clients can't discover or call it yet (discovery goes through
   the gateway's tool list, not the manager's map). A GetServedManagedTool hit
   here would return metadata for a tool that clients can't invoke yet — no
  harm.


 Remove direction: servedToolsMap no longer has the tool, but the gateway
  still does briefly. A client could call it, GetServedManagedTool returns
  nil, and the broker returns "tool not found." This already exists if a client has tool/list and hasn't yet reacted to or received a change notification

jasonmadigan:
@maleck13 this looks good to me. 

Maybe add something explicit to the project's CLAUDE.md (or some doc linked off from it) about this new principal, kind of like the [performance](https://github.com/Kuadrant/mcp-gateway/blob/main/CLAUDE.md#performance) ones ones?

Aman-Cool:
@maleck13, Nice refactor; the buffer-of-1 coalescing is a clean solve for the notification thundering-herd problem.

Two things I noticed that might be worth a follow-up PR, curious what you think:

**`status` field looks like a data race** : `GetStatus()` is called from HTTP handlers on arbitrary goroutines while `setStatus()` writes from the `Start` loop with no synchronisation between them. A small `statusMu sync.RWMutex` would cover it, or is there something I'm missing about how it's used?

**`getTools()` dropping the read lock** : the "Must only be called from the `Start()` event loop" invariant holds for ticker/notification paths, but `broker.go` calls `man.Stop()` directly from `OnConfigChange()` (a different goroutine). That means `Stop()` -> `removeAllTools()` can write `man.tools` under `toolsLock` concurrently with `manage()` -> `getTools()` reading it without any lock. Same applies to the `len(man.serverTools)` reads in `shouldFetchTools()` and `setStatus()`. Would restoring the `RLock` in `getTools()` work, or do you have a different invariant in mind?

Also worth adding a test that calls `Stop()` from a separate goroutine while `manage()` is in flight ; `-race` should catch both of the above.

maleck13:
> @maleck13, Nice refactor; the buffer-of-1 coalescing is a clean solve for the notification thundering-herd problem.
> 
> Two things I noticed that might be worth a follow-up PR, curious what you think:
> 
> **`status` field looks like a data race** : `GetStatus()` is called from HTTP handlers on arbitrary goroutines while `setStatus()` writes from the `Start` loop with no synchronisation between them. A small `statusMu sync.RWMutex` would cover it, or is there something I'm missing about how it's used?

It is expected that this endpoint will be called repeatedly at the moment. So is eventually consistent. I have added a small update to write the status in one go

> 
> **`getTools()` dropping the read lock** : the "Must only be called from the `Start()` event loop" invariant holds for ticker/notification paths, but `broker.go` calls `man.Stop()` directly from `OnConfigChange()` (a different goroutine). That means `Stop()` -> `removeAllTools()` can write `man.tools` under `toolsLock` concurrently with `manage()` -> `getTools()` reading it without any lock. Same applies to the `len(man.serverTools)` reads in `shouldFetchTools()` and `setStatus()`. Would restoring the `RLock` in `getTools()` work, or do you have a different invariant in mind?

While there are races here the impact is negligible. That said it would cause an issue with the race detector. So I will add something for these to allow us to keep the race detection and avoid regression.

> 
> Also worth adding a test that calls `Stop()` from a separate goroutine while `manage()` is in flight ; `-race` should catch both of the above.

maleck13:
I am going to expand this refactor a little. I am going to have the manger return a Start() that returns an `ActiveMCP` interface. So Start must be called to get an ActiveMCP. An ActiveMCP can be queried and Stopped()

internal/broker/upstream/manager.go
```diff
@@ -128,114 +140,129 @@ func NewUpstreamMCPManager(upstream MCP, gatewayServer ToolsAdderDeleter, logger
...
+			case evt := <-man.events:
+				man.logger.Debug("received event", "upstream mcp server", man.mcp.ID(), "event", evt)
+				man.manage(ctx, evt)
+			}
 		}
-	}
+	}()
+
+	return &activeMCP{manager: man, cancel: cancel}
 }
```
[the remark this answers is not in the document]
maleck13: Yes I missed this 👍

maleck13:
@jasonmadigan ready for review again now. I have fully isolated the manager from the broker via interface and made it so a start function is returned to the broker for each new mcp manager. The start function kicks off the routine and returns an ActiveMCP interface that the broker can then use to query the MCP info or Stop the manager

internal/broker/upstream/manager_test.go
```diff
@@ -1017,3 +1031,109 @@ func TestMCPManager_NewUpstreamMCPManager_nilGateway(t *testing.T) {
...
+	mock.hasToolsCap = false
+	// slow down ListTools so manage() is mid-flight when Stop() fires
+	mock.listToolsDelay = 100 * time.Millisecond
+	gateway := NewMockGatewayServer()
+	manager, err := NewUpstreamMCPManager(mock, gateway, logger, time.Hour, mcpv1alpha1.InvalidToolPolicyFilterOut)
+	require.NoError(t, err)
+
+	active := manager.Start(context.Background())
+	// let Start complete its initial manage() call
+	time.Sleep(200 * time.Millisecond)
```
jasonmadigan, line 1126: could we `.Eventually` these?

internal/broker/upstream/manager_test.go
```diff
@@ -1017,3 +1031,109 @@ func TestMCPManager_NewUpstreamMCPManager_nilGateway(t *testing.T) {
...
+
+	// trigger another manage() on the event loop via the events channel;
+	// ListTools will block for 100ms giving us time to call Stop()
+	manager.events <- eventTypeNotification
+
+	time.Sleep(10 * time.Millisecond)
```
jasonmadigan, line 1132: ^

jasonmadigan (approved):

internal/broker/upstream/manager_test.go
maleck13: We could but not sure of the advantage. The server in the background in a  mock that has a 100 ms delay so I don't see these being flakey

maleck13:
@jasonmadigan I don't think the sleeps in this test case will create flakey tests. The listToolDelay is deterministic and the sleep is just ensuring that the manage loop has time to execute. If we see flakiness in these I would be surprised and happy to refactor

PR #948, opened 2026-05-12
feat: implement exponential backoff for MCP health checks

### LINKED ISSUE #467: Enhancement. Change the broker MCP health check to a exponential backoff  on failure rather than set interval

vibhor-5:
Implements exponential backoff for MCP health checks using k8s.io/apimachinery/pkg/util/wait.

This addresses the feedback from PR #805:
- Uses a standard Kubernetes utility instead of a custom one.
- Includes jitter to prevent thundering herd.
- Resets backoff immediately on success.
- Applies backoff immediately after the first failure.

Closes #467

## Summary by CodeRabbit

* **New Features**
  * Added exponential backoff to upstream manager error handling for improved failure recovery.

* **Bug Fixes**
  * Enhanced status reporting to reflect current tool count when tool fetching is skipped.
  * Added error logging for connection failures.

* **Tests**
  * Added test coverage for backoff behavior under various failure scenarios.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Kuadrant/mcp-gateway/pull/948)

### DIFF
--- a/internal/broker/upstream/manager.go
+++ b/internal/broker/upstream/manager.go
@@ -13,6 +13,8 @@ import (
 	"sync"
 	"time"
 
+	"k8s.io/apimachinery/pkg/util/wait"
+
 	mcpv1alpha1 "github.com/Kuadrant/mcp-gateway/api/v1alpha1"
 	"github.com/Kuadrant/mcp-gateway/internal/config"
 	"github.com/mark3labs/mcp-go/mcp"
@@ -105,7 +107,11 @@ type MCPManager struct {
 	ticker *time.Ticker
 	// tickerInterval is the interval between backend health checks
 	tickerInterval time.Duration
-	gatewayServer  ToolsAdderDeleter
+	// backoff is used to calculate the next interval on failure
+	backoff wait.Backoff
+	// baseBackoff stores the initial backoff configuration for resets
+	baseBackoff   wait.Backoff
+	gatewayServer ToolsAdderDeleter
 	// serverTools is an internal copy that contains the managed MCP's tools with prefixed names. It is these that are externally available via the gateway
 	serverTools []server.ServerTool
 	// tools is the original set from MCP server with no prefix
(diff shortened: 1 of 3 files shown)


vibhor-5:
@david-martin The PR is a follow up on previously closed #805 using the apimachinery/wait package as suggested , kindly review

david-martin:
Thanks for taking this up. The manager was refactored to a channel-based event loop in #943 (merged), so this will need a rebase. Once rebased and the backoff logic is adapted to the new event loop, this should be good to go.

vibhor-5:
@david-martin Done. Rebased the branch and made minor tweaks to adapt to the event loop.

david-martin (approved):
Looks good. Clean use of `k8s.io/apimachinery/pkg/util/wait.Backoff`, backoff progression is sensible (1s → 2s → 4s → ... → cap at tickerInterval), and reset-on-success is correct.

Two minor suggestions for follow-up (non-blocking):

1. **`setStatus` side effects**: `setStatus` now modifies the ticker interval via `applyBackoff`/`resetBackoff`. Consider moving those calls into `manage()` directly so `setStatus` stays a pure status setter.

2. **Test coupling**: `TestMCPManager_Backoff` asserts on `backoff.Steps` internal state. A more robust check would verify the ticker was reset or assert on status/readiness, which is less coupled to the `wait.Backoff` implementation.

The `happy_path_test.go` fix is already on main (9dbe0ab) so that commit is redundant, but Git resolves it cleanly so no issue merging as-is.

vibhor-5:
> Two minor suggestions for follow-up (non-blocking):
> 
> 1. **`setStatus` side effects**: `setStatus` now modifies the ticker interval via `applyBackoff`/`resetBackoff`. Consider moving those calls into `manage()` directly so `setStatus` stays a pure status setter.
> 2. **Test coupling**: `TestMCPManager_Backoff` asserts on `backoff.Steps` internal state. A more robust check would verify the ticker was reset or assert on status/readiness, which is less coupled to the `wait.Backoff` implementation.

Made a follow up PR following the suggested changes at #982

PR #923, opened 2026-05-10
fix(broker): take RLock in findServerByName and use MCPName() in log

SarthakB11:
## Summary

Found another concurrency hole in the broker, in the same family as the recent `mcpServers` / session-cache fixes (#860, #872, #888, #918, #921).

`findServerByName` in `internal/broker/filtered_tools_handler.go` iterates `broker.mcpServers` without taking `broker.mcpLock`. `OnConfigChange` mutates the same map under the write lock at `broker.go:175,190,197`, so a `tools/list` request that lands during a config reload races on the iteration. The existing helpers `ToolAnnotations` and `GetServerInfo` already RLock around the same map, so this fix just brings `findServerByName` into line.

While I was in the file I also fixed an adjacent log line in `filterToolsByServerMap`. It passed the method value `upstream.MCPName` instead of calling it (`MCPName()`), so slog rendered a method-value formatter string in place of the actual server name. Happy to split that into a separate PR if preferred.

`TestFilterTools_DataRace` exercises `FilterTools` and a mutator that mirrors `OnConfigChange` under `broker.mcpLock` from two goroutines. Pattern matches `manager_test.go::TestMCPManager_status_DataRace`. Run with `go test -race ./internal/broker/...`.

## Verification

```
go build ./...
go vet ./...
go test -race -count=1 -timeout 60s ./internal/broker/...
make lint
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed race condition in tool filtering that could occur during concurrent configuration updates.
  * Corrected server identification logging in tool filtering operations.

* **Tests**
  * Added a concurrency test to validate thread safety during simultaneous tool filtering and configuration mutations.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Kuadrant/mcp-gateway/pull/923)

PR #951, opened 2026-05-12
fix: add missing mcpLock.RLock in findServerByName to prevent data race

### LINKED ISSUE #947: bug: data race in findServerByName - missing lock on mcpServersbug
## Bug

`findServerByName` in `internal/broker/filtered_tools_handler.go` iterates `broker.mcpServers` with no lock held:

```go
func (broker *mcpBrokerImpl) findServerByName(name string) *upstream.MCPManager {
    for _, upstream := range broker.mcpServers {
        if upstream.MCPName() == name {
            return upstream
        }
    }
    return nil
}
```

Every other method that reads `mcpServers` acquires `mcpLock.RLock()` first. `findServerByName` is called from `filterToolsByServerMap` → `FilterTools` which fires on every `tools/list` request concurrently, while `OnConfigChange` writes to `mcpServers` under `mcpLock`.

## Reproducer

`go test -race` confirms the race with a concurrent reader/writer test.

## Fix

Add `mcpLock.RLock()/RUnlock()` to `findServerByName`.

Happy to open a PR for this.

aniketpandey05:
Fixes #947

`findServerByName` in `filtered_tools_handler.go` was iterating `broker.mcpServers` with no lock held. Every other method that reads `mcpServers` acquires `mcpLock.RLock()` first — this was a clear oversight.

`findServerByName` is called from `filterToolsByServerMap` → `FilterTools` which fires on every concurrent `tools/list` request, while `OnConfigChange` writes to `mcpServers` under `mcpLock.Lock()`. This is a data race in production.

Fix adds `mcpLock.RLock()/RUnlock()` to `findServerByName` and includes a regression test that confirms the race is gone with `go test -race`.

## Bug Fixes
* Improved concurrent access handling during server lookup operations to enhance system stability under simultaneous requests.

## Tests
* Added concurrency test validating server configuration behavior during active concurrent lookups.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Kuadrant/mcp-gateway/pull/951)

### DIFF
--- a/internal/broker/filtered_tools_handler.go
+++ b/internal/broker/filtered_tools_handler.go
@@ -135,6 +135,8 @@ func (broker *mcpBrokerImpl) parseAuthorizedCapabilitiesJWT(headerValues []strin
 }
 
 func (broker *mcpBrokerImpl) findServerByName(name string) *upstream.MCPManager {
+	broker.mcpLock.RLock()
+	defer broker.mcpLock.RUnlock()
 	for _, upstream := range broker.mcpServers {
 		if upstream.MCPName() == name {
 			return upstream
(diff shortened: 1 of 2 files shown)


david-martin:
Thanks for this. #923 addresses the same findServerByName race and also fixes a logging bug in filterToolsByServerMap. Closing as a duplicate of #923.

Issue #950, opened 2026-05-12
Investigate support for Roots


david-martin:
Split from #257.

Investigate support for MCP [Roots](https://modelcontextprotocol.io/specification/2025-06-18/client/roots) capability.

The roots capability is currently declared in mcp.go but has no implementation behind it.

## Subtasks

- [ ] Investigate whether Roots fits the mcp-gateway use cases
- [ ] Implement Roots support (if applicable)

## Notes from community (#257)

For roots, extracting roots from the downstream client initialize params, storing them in the session cache, and responding to roots/list requests from upstream using those stored roots seems like the right approach.

aniketpandey05:
@david-martin  I started investigating the implementation. 

Storing the roots is straightforward during the initialize response, we can extract roots from the downstream client's params and store them in the session cache (keyed by gateway session ID). I've implemented this part.

The challenge is the roots handler on the upstream client side. The upstream broker client (in internal/broker/upstream/mcp.go) needs to respond to roots requests from upstream servers, but it doesn't have access to the session cache or the gateway session ID to look up the stored roots. These live in different layers the session cache is in the router layer, the upstream client is in the broker layer.

mcp-go does provide a WithRootsHandler ClientOption that handles roots at the transport level, but wiring the session cache into the broker layer would require passing it down through several layers.

What's the preferred approach here? Should we pass the session cache reference down to the broker layer, or is there a cleaner way to handle this?

david-martin:
@rogueslasher Thanks for taking the initiative here.

Before we consider the implementation, I'd like to better understand 'roots' as a feature of MCP servers, and what it's purpose is.
Does it make sense for http based mcp servers, or is it moreso intended for stdio?
I really don't know at this point.
So, it would be good to understand the value of 'roots' from an MCP gateway point of view, for users of the gateway (agents/clients, platform admins etc..).
Perhaps some example 'job stories' so we can make a call on if this is a feature that makes sense at the gateway. 

Perhaps even a draft design doc to further the conversation with the team (see https://github.com/Kuadrant/mcp-gateway/blob/main/docs/design/CLAUDE.md)

However, if you'd prefer not to take on something of this size, there's no obligation. No worries.

manik3160:
Hey @david-martin ! I have started a draft design doc in #959 based on the discussion above.

[...]

PR #959, opened 2026-05-13
design: add MCP Roots support investigation

manik3160:
Design doc for #950

The gateway currently advertises the roots capability in every upstream initialize handshake (mcp.go:118) but has no implementation behind it. This doc proposes storing roots from the downstream client's initialize params in the session cache, and wiring up a roots/list handler on the upstream broker client that reads them back. The gateway stays transparent — it doesn't generate or interpret roots, just holds them per session and returns them when upstream asks.

### DIFF
--- a/docs/design/roots/roots-design.md
+++ b/docs/design/roots/roots-design.md
@@ -1,0 +1,97 @@
+# MCP Roots Support
+
+## Problem
+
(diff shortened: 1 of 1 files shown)


david-martin:

Patryk-Stefanski:
Thanks for the design investigation @manik3160. Per [CONTRIBUTING.md](https://github.com/Kuadrant/mcp-gateway/blob/main/CONTRIBUTING.md), feature and speculative work should reference an existing issue so the design can be discussed before implementation. Could you file an issue for MCP Roots support first? That way maintainers can weigh in on the approach before the design doc lands.

david-martin:
FYI, roots support is deprecated in the upcoming mcp spec.
As such I've closed the investigation issue https://github.com/Kuadrant/mcp-gateway/issues/950

PR #964, opened 2026-05-14
fix: acquire manageMu in Stop() to prevent tool re-registration race

Aman-Cool:
I was reading through the manager code and noticed something subtle with how `Stop()` and `manage()` interact.

When `Stop()` is called ; say, because a hot-reload removed a server; it grabs `toolsLock`, calls `removeAllTools()` to delete the tools from the broker, then disconnects and closes `done`. Clean enough on its own.

The problem is `manage()` doesn't know any of this is happening. It holds `manageMu` for its entire lifetime, but when it gets to the `getTools()` call it's just sitting there waiting on a network response from the backend. `Stop()` doesn't touch `manageMu` at all, so it runs right through ; clears the tools, releases `toolsLock`, disconnects the backend, closes `done`; and exits.

Then the network response comes back. `manage()` picks up where it left off, acquires `toolsLock` (nobody's holding it anymore), computes the diff between what it had before and what it just fetched, and calls `gatewayServer.AddTools()` with the result. The tools are back in `listeningMCPServer`.

But the manager itself is already deleted from `m.mcpServers` by `OnConfigChange`. So now `tools/list` shows those tools, but `GetServerInfo()` can't find the server config for them because it only looks in `m.mcpServers`. Every `tools/call` for those tools returns "Tool not found". And since no future code path will call `DeleteTools()` for a stopped, untracked manager, they stay there until the process restarts.

This is most likely to bite on a server's first `manage` cycle ; when `man.tools` is still empty, `toAdd` equals the full fetched list, so `AddTools()` always fires if the race hits.

The fix is small: `Stop()` acquires `manageMu` before calling `removeAllTools()`. It blocks until any running `manage()` finishes, then clears the tools. By that point the ticker is already stopped, `done` is about to close, so no new `manage()` can sneak in after. Three lines, no new fields, no new locks.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed a race condition during shutdown that could cause tools to be improperly re-added after cleanup, improving overall shutdown reliability.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Kuadrant/mcp-gateway/pull/964)

### DIFF
--- a/internal/broker/upstream/manager.go
+++ b/internal/broker/upstream/manager.go
@@ -159,7 +159,12 @@ func (man *MCPManager) Start(ctx context.Context) {
 func (man *MCPManager) Stop() {
 	man.stopOnce.Do(func() {
 		man.ticker.Stop()
+		// Acquire manageMu to wait for any in-flight manage() call to complete before
+		// clearing tools. Without this, manage() can re-add tools via AddTools() after
+		// removeAllTools() has deleted them, leaving the broker with un-routable tools.
+		man.manageMu.Lock()
 		man.removeAllTools()
+		man.manageMu.Unlock()
 		if err := man.MCP.Disconnect(); err != nil {
 			man.logger.Error("failed to disconnect during stop", "upstream mcp server", man.MCP.ID(), "error", err)
 		}


Aman-Cool:
@david-martin @maleck13, Found this while reading through the manager shutdown path ; `Stop()` never acquires `manageMu`, so if a `manage()` call is mid-flight on a `ListTools()` network request when `Stop()` runs, it can come back and re-add tools to the broker after `removeAllTools()` already deleted them. Since the manager is gone from `mcpServers` by then, those tools show up in `tools/list` but always 404 on `tools/call`. Grabbing `manageMu` in `Stop()` before the cleanup lets any in-flight `manage()` drain first. Tiny change, no new fields.

happy to adjust the approach if you'd prefer handling it differently.

david-martin:
Thanks for this fix. PR #943 (merged) refactored the manager's event loop and touched the same file, so this will need a rebase.

After rebasing, please verify that the race you identified (manage() re-adding tools after Stop() clears them) is still present with the current code on main. The manageMu/manage() pattern is still in place, so the fix likely still applies, but worth confirming against the latest.

maleck13:
This is not an issue with the linked PR. The stop function uses channel based approach now this is the same select that calls the manage method.

PR #966, opened 2026-05-14
chore: fix methodInitialized constant spelling and use it in isInitializeRequest

### LINKED ISSUE #958: chore: methodInitialized constant has wrong spelling and is never used
## Description

The constant `methodInitialized` in `internal/mcp-router/request_handlers.go`
is defined with the wrong method name and is never referenced anywhere in the codebase.

```go
const methodInitialized = "notification/initialized"   // singular — WRONG
```

The correct MCP spec method name is `"notifications/initialized"` (plural).

`isInitializeRequest()` hardcodes the correct string directly, bypassing
the constant entirely:

```go
func (mr *MCPRequest) isInitializeRequest() bool {
    return mr.Method == "initialize" || mr.Method == "notifications/initialized"
}
```

## Risk

Any future code written using `methodInitialized` (e.g. `mr.Method == methodInitialized`)
will silently fail to match because the constant is misspelled. The constant
exists to prevent exactly this class of mistake, but currently makes it
more likely.

## Fix

Either fix and use the constant:
```go
const methodInitialized = "notifications/initialized"  // plural — correct
```

And update `isInitializeRequest()` to use it:
```go
func (mr *MCPRequest) isInitializeRequest() bool {
    return mr.Method == methodInitialize || mr.Method == methodInitialized
}
```

Or remove the unused constant entirely.

## Code Reference

`internal/mcp-router/request_handlers.go`
Constant: `methodInitialized`
Function: `isInitializeRequest()`

Prachi194agrawal:
## Summary

Fixes #958

The constant `methodInitialized` in `internal/mcp-router/request_handlers.go` was defined with the wrong MCP spec method name (`"notification/initialized"` — singular), and was never referenced in the codebase. `isInitializeRequest()` bypassed the constant entirely by hardcoding the correct string directly.

## Changes

1. Fixed the constant value from `"notification/initialized"` to `"notifications/initialized"` (plural, per MCP spec).
2. Updated `isInitializeRequest()` to use `methodInitialized` instead of the hardcoded string literal.

## Before / After

**Constant:**
```go
// Before
const methodInitialized = "notification/initialized"   // wrong — singular

// After
const methodInitialized = "notifications/initialized"  // correct — plural
```

**Function:**
```go
// Before
func (mr *MCPRequest) isInitializeRequest() bool {
    return mr.Method == "initialize" || mr.Method == "notifications/initialized"
}

// After
func (mr *MCPRequest) isInitializeRequest() bool {
    return mr.Method == methodInitialize || mr.Method == methodInitialized
}
```

## Why This Matters

The constant exists to prevent silent string-mismatch bugs. With the old misspelled value, any future code using `methodInitialized` directly would silently fail to match. This fix ensures the constant is both correct and actively used, making the codebase safer to extend.

## Code Reference

**File:** `internal/mcp-router/request_handlers.go`  
**Constant:** `methodInitialized`  
**Function:** `isInitializeRequest()`

### DIFF
--- a/internal/mcp-router/request_handlers.go
+++ b/internal/mcp-router/request_handlers.go
@@ -62,7 +62,7 @@ func NewRouterErrorf(code int32, format string, args ...any) *RouterError {
 const (
 	methodToolCall    = "tools/call"
 	methodInitialize  = "initialize"
-	methodInitialized = "notification/initialized"
+	methodInitialized = "notifications/initialized"
 
 	elicitationResultAction  = "action"
 	elicitationActionAccept  = "accept"
@@ -128,7 +128,7 @@ func (mr *MCPRequest) isToolCall() bool {
 
 // isInitializeRequest returns true if the method is initialize or initialized
 func (mr *MCPRequest) isInitializeRequest() bool {
-	return mr.Method == "initialize" || mr.Method == "notifications/initialized"
+	return mr.Method == methodInitialize || mr.Method == methodInitialized
 }
 
 // clientSupportsElicitation checks if an initialize request declares elicitation support


david-martin:

jasonmadigan:
thanks @Prachi194agrawal, closing -- the methodInitialized constant was removed on main and the correct string is now used directly. this would conflict.

PR #967, opened 2026-05-14
fix: rewrite elicitation IDs in sseRewriter.Flush() for final partial SSE lines

### LINKED ISSUE #957: bug: sseRewriter.Flush() skips elicitation ID rewrite for partial final SSE line
## Description

`Flush()` in `internal/mcp-router/elicitation.go` returns the leftover
buffer contents raw, without passing them through `maybeRewriteElicitation`.

`Process()` is intentionally line-buffered — it only processes complete
lines (ending with `\n`) and holds partial lines in `w.buf` for the next
chunk. However, `Flush()` — which is called at `EndOfStream` — never
attempts to rewrite whatever remains in the buffer:

```go
func (w *sseRewriter) Flush(ctx context.Context) []byte {
    remaining := w.buf
    w.buf = nil
    for _, id := range w.gatewayIDs {
        w.idMap.Remove(ctx, id)
    }
    return remaining  // returned raw — maybeRewriteElicitation never called
}
```

## Scenario That Triggers It

1. Backend sends a final SSE chunk **without a trailing `\n`**:
data: {"jsonrpc":"2.0","method":"elicitation/create","id":"backend-123"}
2. `Process()` appends it to `w.buf` but `bytes.IndexByte(w.buf, '\n')`
returns `-1`, so the loop breaks — the line stays in `w.buf` unprocessed.
3. Envoy signals `EndOfStream = true` → `server.go` calls `Flush()`.
4. `Flush()` returns the raw bytes — the backend ID `"backend-123"` is
**never rewritten** to a gateway ID.
5. The client receives `elicitation/create` with the raw backend ID and
sends it back as an elicitation response — but the gateway has no
mapping for `"backend-123"`, so the response is **silently dropped**.

## Impact

Elicitation silently breaks for the last event in any SSE stream where
the final chunk lacks a trailing `\n`. The client sends an elicitation
response the gateway cannot route.

## Suggested Fix

```go
func (w *sseRewriter) Flush(ctx context.Context) []byte {
 remaining := w.buf
 w.buf = nil

 if len(remaining) > 0 {
     if bytes.HasPrefix(bytes.TrimSpace(remaining), dataPrefix) {
         remaining = w.maybeRewriteElicitation(ctx, remaining)
     }
 }

 for _, id := range w.gatewayIDs {
     w.idMap.Remove(ctx, id)
 }
 return remaining
}
```

This mirrors the same check already present in `Process()` and ensures
the final partial line gets the same rewrite treatment as all other lines.

## Code Reference

- File: `internal/mcp-router/elicitation.go`
- Functions: `Flush()`, `Process()`

Prachi194agrawal:
## Summary

Fixes #957

`Flush()` in `internal/mcp-router/elicitation.go` was returning the leftover buffer raw, without passing it through `maybeRewriteElicitation()`. Since `Process()` is line-buffered and only handles complete lines (ending with `\n`), a final SSE chunk without a trailing newline would sit in `w.buf` unprocessed — and `Flush()` would return it with the raw backend ID unrewritten.

## Root Cause

`Process()` breaks out of its loop when `bytes.IndexByte(w.buf, '\n')` returns `-1`, leaving the partial line in `w.buf`. When Envoy signals `EndOfStream = true`, `Flush()` is called — but the original code returned `remaining` immediately without attempting a rewrite:

```go
// Before — raw return, no rewrite attempted
func (w *sseRewriter) Flush(ctx context.Context) []byte {
    remaining := w.buf
    w.buf = nil
    for _, id := range w.gatewayIDs {
        w.idMap.Remove(ctx, id)
    }
    return remaining
}
```

## Fix

Added the same `dataPrefix` guard already used in `Process()` before returning `remaining`, so the final partial line gets identical rewrite treatment:

```go
// After
func (w *sseRewriter) Flush(ctx context.Context) []byte {
    remaining := w.buf
    w.buf = nil
    if len(remaining) > 0 {
        if bytes.HasPrefix(bytes.TrimSpace(remaining), dataPrefix) {
            remaining = w.maybeRewriteElicitation(ctx, remaining)
        }
    }
    for _, id := range w.gatewayIDs {
        w.idMap.Remove(ctx, id)
    }
    return remaining
}
```

## Impact Without This Fix

If the backend sends a final SSE chunk without a trailing `\n` (e.g. `data: {"method":"elicitation/create","id":"backend-123"}`), the client receives the raw backend ID. When the client sends back an elicitation response using that ID, the gateway has no mapping for it and silently drops the response — breaking elicitation for the last event in any such stream.

## Code Reference

**File:** `internal/mcp-router/elicitation.go`  
**Functions:** `Flush()`, `Process()`

### DIFF
--- a/internal/mcp-router/elicitation.go
+++ b/internal/mcp-router/elicitation.go
@@ -61,6 +61,11 @@ func (w *sseRewriter) Process(ctx context.Context, chunk []byte) []byte {
 func (w *sseRewriter) Flush(ctx context.Context) []byte {
 	remaining := w.buf
 	w.buf = nil
+	if len(remaining) > 0 {
+		if bytes.HasPrefix(bytes.TrimSpace(remaining), dataPrefix) {
+			remaining = w.maybeRewriteElicitation(ctx, remaining)
+		}
+	}
 	for _, id := range w.gatewayIDs {
 		w.idMap.Remove(ctx, id) // tool request + response finished, no need to hold onto the mappings any more
 	}


david-martin:

jasonmadigan (approved):
lgtm. correct fix, guards match Process(). would be nice to add a test case for partial elicitation lines in Flush but not blocking.

PR #976, opened 2026-05-14
fix(router): prevent duplicate backend sessions on concurrent tool calls

Aman-Cool:
While I was in `initializeMCPSeverSession` for #911, I was focused on the `sessionCloser` context leak and didn't look hard enough at the initialization path itself.

The cache check at line 622 is a bare `GetSession` read with nothing around it. Two goroutines on the same gateway session hitting the same backend for the first time both read an empty map, both clear the check, both fall into `InitForClient`. That call does a full TCP + MCP handshake; easily 50–200ms on any remote backend, so the race window is wide. Both goroutines race to `AddSession`; the copy-on-write there (#872) means no corruption, but last writer wins: the first `clientHandle` is now orphaned. Its mcp-go SSE reader goroutine stays alive, the backend session stays open, and the `time.AfterFunc` cleanup still fires; but so does the second goroutine's. Two timers, same `gatewaySessionID`. The first one to call `DeleteSessions` wipes the whole session map for that client.

The realistic trigger is any agent framework that fans out parallel tool calls on a fresh session before a backend session exists, which is basically every first-turn multi-tool call.

Fix wraps the creation block in `singleflight.Do` keyed on `gatewaySessionID + "/" + serverName`. Concurrent calls collapse into one; all callers get the same session ID back; exactly one `clientHandle`, one timer. `golang.org/x/sync` is already an indirect dep so no new module dependency.

## Bug Fixes
* Backend session initialization now properly serializes concurrent requests per gateway session and backend server, preventing duplicate sessions and connection leaks.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Kuadrant/mcp-gateway/pull/976)

### DIFF
--- a/internal/mcp-router/request_handlers.go
+++ b/internal/mcp-router/request_handlers.go
@@ -665,7 +665,7 @@ func (s *ExtProcServer) initializeMCPSeverSession(ctx context.Context, mcpReq *M
 			}
 			passThroughHeaders["user-agent"] = "mcp-router"
 		}
-		s.Logger.DebugContext(ctx, "initializing target as no mcp-session-id found for client", "server ", mcpReq.serverName, "with passthrough headers", passThroughHeaders)
+		s.Logger.DebugContext(ctx, "initializing target as no mcp-session-id found for client", "server", mcpReq.serverName, "passthrough header count", len(passThroughHeaders))
 
 		// check if the original client declared elicitation support
 		if !mcpReq.clientElicitation {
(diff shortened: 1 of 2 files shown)


Aman-Cool:
@david-martin and @maleck13, would love a review when you get a chance!

maleck13 (approved):
Happy with this change. 👍  . It highlights a gap in our testing. I have created a follow up for a new e2e test
https://github.com/Kuadrant/mcp-gateway/issues/989

Issue #477, opened 2026-01-12
E2E tests for HTTPS MCP server backends


jasonmadigan:
#450 verified that HTTPS works for MCP server backends with publicly-trusted CAs (both in and off-cluster, with real certs). In-cluster backends with private CAs (self-signed, cert-manager, OpenShift service-serving CA) need the custom TLS config work first (#659).

The current hostname backendRef e2e test only verifies the controller processes the MCPServerRegistration correctly via a status condition. It doesn't test actual tool discovery because the controller always generates HTTPS URLs for external services, and our test servers only speak HTTP.

## Tests

Three categories of e2e test. None should run per-PR since they depend on external services.

- [ ] **In-cluster with real public certs** -- Needs a cluster with real TLS termination (e.g. OpenShift Route, load balancer with cert). Deploy test servers behind a gateway with a publicly-trusted wildcard cert. Verify initialize, tools/list, and tools/call all work over HTTPS end-to-end. Previously verified manually on OpenShift 4.20 with Let's Encrypt (#450), verification procedure documented there.

- [ ] **External service with public TLS (e.g. GitHub MCP)** -- Works today. Register GitHub MCP server as an external hostname backend. Verify broker discovers tools over HTTPS (43+ tools currently). Assert config contains `https://` URL for the backend. Requires a GitHub PAT stored as a CI secret.

- [ ] **In-cluster with private CA** -- Needs #659 (custom TLS config, `tls.caCertSecretRef`). Deploy a TLS-enabled test MCP server with a self-signed or cert-manager CA. Register with `tls.caCertSecretRef` pointing to the CA bundle. Verify broker connects over HTTPS and discovers tools. Negative case: same server without TLS config should fail with a cert verification error.

## Running

These tests hit external services or need cert/cluster infrastructure that isn't part of the standard Kind setup. They should run on-demand or on a schedule, not per-PR.

## Related

- #450: verified HTTPS works for public CAs, OpenShift verification procedure documented
- #659: custom TLS config, blocks the private CA test
- PR #435: external MCP server support ([discussion](https://github.com/Kuadrant/mcp-gateway/pull/435#discussion_r2681497713))

## Parent issue
- Part of #659

Issue #305, opened 2025-11-03
Benchmarking


shaneutt:
The purpose of this task is to provide a benchmarking tool for the MCP Gateway. The **main goal is to be able to track our performance from release to release** and protect against significant performance regressions.

### Acceptance Criteria

- [ ] Should have "scenarios" (e.g. different auth schemas, etc)
- [ ] Needs a README.md, and should be straightforward to run manually
- [ ] Hook it up in CI
  - [ ] Results should be reported, so they can be tracked over time
  - [ ] Before releases, performance reports should be automatically compared
    - [ ] Notify when performance was improved over the last release
    - [ ] Alert and block the release when performance degraded significantly
  - [ ] Report performance profiles for inspection

> **Note**: This doesn't need to be enormous we should keep it lean.

> **Note**: We don't necessarily need to start this until we're confident that we're intending to move to release with the current architecture.

PR #979, opened 2026-05-15
replace synchronous polling with fire and forget 

### LINKED ISSUE #629: Tech Debt - Improve how checking the status of the MCPServers with the gateway is handled

malladinagarjuna2:
closes #629
Remove ServerValidator and the synchronous HTTP probe from the reconciler
Set Accepted=True immediately after UpsertMCPServer succeeds — no blocking network call
Add disabled-state pre-flight: Spec.State == Disabled short-circuits before any upsert
Rename condition type Ready → Accepted across controller, API types, E2E verifiers and tests
Add bug condition exploration test and 4 preservation property tests

## Summary by CodeRabbit

* **Bug Fixes**
  * Removed synchronous broker polling that delayed MCPServerRegistration confirmation

* **Refactor**
  * MCPServerRegistration status conditions renamed from "Ready" to "Accepted"
  * Server configuration marked accepted immediately after write
  * kubectl display column updated to show "Accepted" status

* **Tests**
  * Added integration and preservation-property tests; updated e2e verifications to use Accepted semantics

* **CI / Conformance**
  * Conformance workflow and deploy steps now poll for route readiness and wait for Accepted condition

[Review Change Stack](https://app.coderabbit.ai/change-stack/Kuadrant/mcp-gateway/pull/979)


.golangci.yml
```diff
@@ -94,10 +94,9 @@ linters:
           - revive
           - staticcheck
         text: "dot.imports|ST1001"
-      # gosec false-positives in e2e test code: G101 test-only EC signing
-      # key, G109/G115 trusted int conversions in builders, G204 kubectl
-      # shell-outs with controlled inputs, G402 self-signed Kind certs.
-      # Other gosec rules still apply.
+      # gosec false-positives in e2e test code: self-signed Kind certs (G402),
```
Patryk-Stefanski: Is this related to the work done in this PR?

tests/e2e/kubectl_helpers.go
```diff
@@ -95,6 +96,51 @@ func RestartDeploymentAndWait(ctx context.Context, namespace, deploymentName str
 	return nil
 }
 
+// AddDeploymentCommandFlag appends a flag to a deployment's container command array.
+// The flag is not reverted by the controller if it is in the ignoredCommandFlags list.
+func AddDeploymentCommandFlag(ctx context.Context, namespace, deploymentName, flag string) error {
```
Patryk-Stefanski, line 100: These two new helpers (AddDeploymentCommandFlag, RemoveDeploymentCommandFlag) aren't called anywhere in this PR.

tests/e2e/kubectl_helpers.go
```diff
@@ -23,7 +24,7 @@ func ScaleDeployment(ctx context.Context, namespace, name string, replicas int)
 }
 
 // WaitForDeploymentReady waits for a deployment to be ready
-func WaitForDeploymentReady(ctx context.Context, namespace, name string) error {
+func WaitForDeploymentReady(ctx context.Context, namespace, name string, _ int) error {
```
Patryk-Stefanski: Compilation breakage: Changing this signature to add _ int breaks 3 callers that weren't updated:
   - [tests/e2e/suite_test.go:153](https://github.com/Kuadrant/mcp-gateway/blob/08f71a3f15466377a6668f8109207968f10b2459/tests/e2e/suite_test.go#L153)
   - [tests/e2e/url_elicitation_test.go:25](https://github.com/Kuadrant/mcp-gateway/blob/08f71a3f15466377a6668f8109207968f10b2459/tests/e2e/url_elicitation_test.go#L25)
   - [tests/e2e/url_elicitation_test.go:57](https://github.com/Kuadrant/mcp-gateway/blob/08f71a3f15466377a6668f8109207968f10b2459/tests/e2e/url_elicitation_test.go#L57)

tests/e2e/happy_path_test.go
```diff
@@ -843,24 +800,19 @@ var _ = Describe("MCP Gateway Registration Happy Path", func() {
 		testResources = append(testResources, registration2.GetObjects()...)
 		server2 := registration2.Register(ctx)
 
-		By("Verifying at least one MCPServerRegistration reports tool conflict in status")
-		// Both servers have tools like "time", "headers" etc.
-		// With the same prefix, they would produce "conflict_time", "conflict_headers" etc.
-		// At least one server should report a conflict.
+		By("Verifying both MCPServerRegistrations are Accepted=True (config written — conflict detection is broker-layer)")
```
Patryk-Stefanski: Same concern as the broken server test — users previously got a "conflict" message in the MCPServerRegistration status. Now both registrations show Accepted=True with no indication of the conflict.
 
The gateway tool list check is a weaker assertion since it doesn't tell the user which registration has a problem.

tests/e2e/happy_path_test.go
```diff
@@ -648,19 +607,17 @@ var _ = Describe("MCP Gateway Registration Happy Path", func() {
 		testResources = append(testResources, registration.GetObjects()...)
 		registeredServer := registration.Register(ctx)
 
-		By("Verifying MCPServerRegistration status reports connection failure")
+		By("Verifying MCPServerRegistration is Accepted=True immediately (fire-and-forget — no network probe)")
```
Patryk-Stefanski: The previous test verified that the controller reported "unsupported protocol version" in the MCPServerRegistration status, which was a meaningful user-facing signal. Now it just checks Accepted=True and verifies tools are absent.

With the fire-and-forget model, how should users discover that their MCP server has a protocol issue? Is there a follow-up planned (e.g. broker-side status reporting)? Worth documenting the trade-off here or in the PR description.

internal/controller/preservation_property_test.go
```diff
@@ -0,0 +1,315 @@
...
+//   They must pass on unfixed code to establish the baseline.
+//
+// EXPECTED OUTCOME AFTER FIX: ALL PASS
+//   After the fix, the condition type changes from "Ready" to "Accepted" on error paths,
+//   but the semantic (False + error message) is preserved. These tests are written to
+//   check the new "Accepted" condition type so they serve as regression guards post-fix.
+//
+// NOTE: The tests are written against the FIXED condition type ("Accepted") so they
+//   will FAIL on unfixed code (which uses "Ready"). This is intentional — they document
+//   the target behavior and will pass once the fix is applied.
```
Patryk-Stefanski: can we trim this down ?

internal/controller/mcpserverregistration_controller.go
```diff
@@ -621,8 +567,8 @@ func (r *MCPReconciler) updateStatus(
 		mcpsr.Status.Conditions = append(mcpsr.Status.Conditions, condition)
 		statusChanged = true
 	}
-	if mcpsr.Status.DiscoveredTools != toolCount {
-		mcpsr.Status.DiscoveredTools = toolCount
+	if mcpsr.Status.DiscoveredTools != 0 {
+		mcpsr.Status.DiscoveredTools = 0
```
Patryk-Stefanski: DiscoveredTools is now always set to 0, but the field still exists in the CRD status and appears as a kubectl print column (Tools). Users will see a Tools column that's always 0, which is confusing.
 
Two options:
 1. Remove the DiscoveredTools field and the Tools print column from the CRD entirely (since the controller can no longer report this).
 2. Keep it but populate it later (e.g. from the broker's async status reporting — if that's planned).

Either way, this is a user-visible regression in observability that should be called out.

config/e2e/auth/mcps-auth-policy.yaml
```diff
@@ -22,7 +22,7 @@ spec:
         patternMatching:
           patterns:
             - predicate: |
-                request.headers['x-mcp-toolname'] in (has(auth.identity.resource_access) && auth.identity.resource_access.exists(p, p == request.headers['x-mcp-servername']) ? auth.identity.resource_access[request.headers['x-mcp-servername']].roles : [])
+                ('tool:' + request.headers['x-mcp-toolname']) in (has(auth.identity.resource_access) && auth.identity.resource_access.exists(p, p == request.headers['x-mcp-servername']) ? auth.identity.resource_access[request.headers['x-mcp-servername']].roles : [])
```
Patryk-Stefanski, line 32: This changes the authorization predicate to prefix 'tool:' to the toolname header check. This is a behavioral change to auth logic that's unrelated to issue #629.

api/v1alpha1/types.go
Patryk-Stefanski: This should also be updated to `accepted`

Patryk-Stefanski (changes requested):
The core change — removing the synchronous broker probe — aligns with the issue ask and is the right direction, but the Ready → Accepted condition rename is a breaking change that wasn't requested in  the issue and could break user tooling or scripts watching for type: Ready; could we keep the condition name as Ready and just change the semantics? 

There are also several unrelated changes bundled  in (auth policy predicate change in mcps-auth-policy.yaml, unused AddDeploymentCommandFlag/RemoveDeploymentCommandFlag helpers, .golangci.yml comment reformatting) that should be split out, plus a  compilation breakage from the WaitForDeploymentReady signature change missing 3 callers.

malladinagarjuna2:
sure sir , will do that

malladinagarjuna2:
actually @Patryk-Stefanski  i was still working on this , thats why you are seeing so many unrelated changes, sorry for the mess.

Patryk-Stefanski:
@malladinagarjuna2 We would love to see this in 0.7.0, and we are cutting the release end of this week, so please let us know if you need any help.

malladinagarjuna2:
sure sir, i know this is important for issue 255  (one of the high priority issue)  will work on this

tests/e2e/happy_path_test.go
malladinagarjuna2: hey @Patryk-Stefanski you are right that this currently leaves a visibilty gap for runtime connection/protocol errors. The long term plan should be to have the broker handle runtime status reporting(eg: emitting or updating a seperate metrics/status endpoint when it fails to initialize the upstream session) rather than blocking the K8s reconcilation loop.?

Patryk-Stefanski: Agreed on the direction. A few things worth considering for the follow-up:
 
The broker already tracks this, MCPManager maintains a ServerValidationStatus per upstream (with Ready, Message, LastValidated, tool/prompt counts) and updates it on every health check tick. So the  data exists; it just isn't exposed externally yet.
  
For runtime status reporting, I'd recommend exposing this as Prometheus metrics (e.g. mcp_upstream_server_ready{server="..."} gauge, mcp_upstream_server_last_validated timestamp). It's the lightest  option — no K8s client or RBAC needed in the broker, integrates with existing alerting/dashboards, and avoids coupling the broker to the control plane.
 
For this PR:
 - Document the visibility gap as a known limitation in the PR description
 - File a follow-up issue for "expose broker upstream status as Prometheus metrics"
 - Consider whether the DiscoveredTools print column should be removed now (since it's always 0) or left for the metrics work to address — showing 0 is arguably more confusing than no column
 
@maleck13 does this sound reasonable to you, or should we leave this for the next release and address it as one larger PR?

malladinagarjuna2: sure sir

malladinagarjuna2:
i still have many doubts regarding this will address this pr later focussing currently on  the #305, #477

malladinagarjuna2:
hey @Patryk-Stefanski, I'm Back! had a bad time in passing the hefty testcases, could you please review it ?

malladinagarjuna2:
hey @Patryk-Stefanski @david-martin    can you please review it?

Patryk-Stefanski (commented):
## Follow-up Review

The core change is solid — removing the synchronous broker poll and renaming Ready→Accepted matches what #629 asks for. Most of the previous review comments were addressed. A few things remaining:

### 1. Revert the inline `SetupTrustedHeadersAuth`

The PR replaces the `SetupTrustedHeadersAuth(ctx, k8sClient)` call in `happy_path_test.go` with ~70 lines of inline code that's identical to the existing helper in `tests/e2e/builders.go:29-93`. Just keep using the helper — `tool_discovery_test.go` still calls it.

### 2. Don't expose `/status` through the gateway HTTPRoute

The `/status` path match added to `broker_router.go:627-632` (and `tests/e2e/builders.go:822-826`) makes the broker's status endpoint publicly accessible through Envoy with no auth and `Access-Control-Allow-Origin: *`. This leaks server names, namespaces, error messages, and tool counts to unauthenticated external clients.

The issue itself says the status endpoint should be "queried separately" — not routed through the public gateway. The e2e tests can port-forward directly to the broker service on port 8080 to reach `/status` without exposing it externally. Same for the conformance workflow polling.

### 3. Minor: `build/ci.mk:103`

Uses bare `kubectl` instead of `$(KUBECTL)`, inconsistent with surrounding lines.

malladinagarjuna2:
hey @Patryk-Stefanski glad that you have replied ! sure i will make the changes mentioned above!

jasonmadigan (changes requested):
needs a rebase -- merge conflicts with main.

malladinagarjuna2:
@Patryk-Stefanski  @jasonmadigan  kindly review this pr, after the rebase some extra changes were required to be done.

i have implemented all the three changes you asked for in place of port-forwarding i implemented get --raw, because of two reasons 
as once we removed the /status endpoint from public istio configuratiion then port forwarding will not work as envoyy requires a route and it will return 404

another reason for this port-forward is flaky in the tests.

malladinagarjuna2:
And apologies for the late work!

malladinagarjuna2:
hey @jasonmadigan  @Patryk-Stefanski  @david-martin  i have again rebased it 
 kindly review

malladinagarjuna2:
hey @jasonmadigan can you please review this ?

jasonmadigan:
@malladinagarjuna2 reviewers are pinged on changes, no need for all of these pings please.

malladinagarjuna2:
> @malladinagarjuna2 reviewers are pinged on changes, no need for all of these pings please.

sorry sir, actually its been high time on this pr and i have rebased it many times and then  passing of all the tests after the changes is a havoc

jasonmadigan:
Thanks for taking a look at this. This issue is already assigned to a maintainer, so we'll handle it internally. As a general rule, if an issue has an assignee it's probably best to check before starting work on it — saves you the effort if someone's already on it. Appreciate the contribution though.

malladinagarjuna2:
@jasonmadigan sir, if some part of my work is useful for solving this issue then can you please cherry-pick my commits ? and make me a coauthor ? i would be really grateful for this !

PR #985, opened 2026-05-17
fix: deleted MCPVirtualServer resources persist in broker after config change

### LINKED ISSUE #984: Bug: Deleted MCPVirtualServer resources persist in broker after config reconciliation`
### Summary

When an `MCPVirtualServer` resource is deleted from the cluster, it is not removed from the broker’s in-memory virtual server registry. Clients sending requests with the matching `X-Mcp-Virtualserver` header continue receiving filtered tool/prompt results for the deleted virtual server as if it still exists.

This causes stale runtime state and allows deleted virtual servers to remain active indefinitely.

---

## Root Cause

The issue exists at two layers.

### 1. Controller deletion path never rewrites virtual server config

`internal/controller/mcpvirtualserver_controller.go`

In the deletion reconciliation path, the controller removes the finalizer but never updates the generated virtual server config:

```go
if !mcpVS.DeletionTimestamp.IsZero() {
    if controllerutil.ContainsFinalizer(mcpVS, mcpGatewayFinalizer) {
        logger.Info("deleting mcpvirtualserver", ...)

        // TODO remove from config

        controllerutil.RemoveFinalizer(mcpVS, mcpGatewayFinalizer)
        ...
    }
    return ctrl.Result{}, nil
}
```

`generateVirtualServerConfig(...)` already excludes resources with a non-zero `DeletionTimestamp`, so the generated config logic itself is correct. However, deletion does not trigger a config rewrite, meaning the stale virtual server remains in the config secret until another reconciliation happens.

If no additional `MCPVirtualServer` changes occur, the stale config can persist indefinitely.

---

### 2. Broker config reconciliation never removes stale virtual servers

`internal/broker/broker.go`

Even when the config secret is eventually corrected, `OnConfigChange(...)` only adds or updates virtual servers:

```go
// register virtual servers
m.vsLock.Lock()
for _, vs := range virtualServers {
    m.virtualServers[vs.Name] = vs
}
m.vsLock.Unlock()
```

Entries absent from the new config are never removed from `m.virtualServers`.

This differs from the upstream MCP server reconciliation logic directly above, which properly removes decommissioned servers:

```go
for serverID := range m.mcpServers {
    if !slices.ContainsFunc(servers, func(s *config.MCPServer) bool {
        return serverID == s.ID()
    }) {
        man.Stop()
        delete(m.mcpServers, serverID)
    }
}
```

Virtual servers currently have no equivalent cleanup path.

---

## Impact

* Deleted `MCPVirtualServer` resources continue serving filtered responses indefinitely.
* Broker runtime state diverges from Kubernetes state.
* Stale virtual servers remain accessible through the `X-Mcp-Virtualserver` header.
* The in-memory `virtualServers` map can grow unbounded over time in environments with frequent create/delete cycles.

---

## Steps to Reproduce

1. Create an `MCPVirtualServer` resource.
2. Verify filtered tool/prompt responses work using the corresponding `X-Mcp-Virtualserver` header.
3. Delete the `MCPVirtualServer`.
4. Send another request with the same `X-Mcp-Virtualserver` header.
5. Observe that the deleted virtual server still returns filtered results.

---

## Expected Behavior

After deletion and reconciliation:

* the virtual server should be removed from the generated config
* the broker should remove it from `m.virtualServers`
* requests referencing the deleted virtual server should no longer resolve

---

## Actual Behavior

Deleted virtual servers remain active in the broker’s in-memory registry and continue serving filtered responses until broker restart.

### LINKED ISSUE #1084: Virtual servers never pruned on config change
## Description

`OnConfigChange` in `broker.go:342-347` adds new virtual servers to the map but never removes stale ones. MCP servers are properly deregistered (lines 289-299), but the `virtualServers` map only accumulates.

If a `MCPVirtualServer` CRD is deleted, its entry persists in the broker indefinitely. Clients can continue using a deleted virtual server name for tool scoping.

## Impact

Stale authorisation state. An admin who deletes a virtual server expects its tool scoping to stop working, but it continues indefinitely.

## Proposed Fix

Clear and rebuild the virtual servers map on each config change:

```go
m.vsLock.Lock()
m.virtualServers = make(map[string]*config.VirtualServer, len(virtualServers))
for _, vs := range virtualServers {
    m.virtualServers[vs.Name] = vs
}
m.vsLock.Unlock()
```

## Review Guidance

- `internal/broker/broker.go:342-347` -- the accumulate-only path
- Compare with MCP server deregistration pattern at lines 289-299

## Manual Verification

1. Create a virtual server, confirm it works
2. Delete the virtual server CRD
3. Before fix: virtual server name still scopes tools. After fix: lookup fails

ayuxsh009:
## Summary

  Fixes #984

  - Broker's `OnConfigChange` was append-only for virtual servers — deleted entries were never removed from `m.virtualServers`, causing stale virtual servers to remain active
   indefinitely
  - Controller's deletion path had a `// TODO remove from config` and never wrote the updated config, so the broker never received a pruned snapshot on deletion

  ## Changes

  **`internal/broker/broker.go`**
  Replace the append-only virtual server loop with a full map rebuild from the new config snapshot, mirroring the existing cleanup logic for upstream MCP servers.

  **`internal/controller/mcpvirtualserver_controller.go`**
  Call `generateVirtualServerConfig` + `WriteVirtualServerConfig` in the deletion path before removing the finalizer, so the config secret is updated immediately when a
  resource is deleted.

  ## Tests

  - `TestOnConfigChange_VirtualServerRemoval` — broker correctly removes virtual servers absent from the new config
  - `TestReconcile_VirtualServerDeletion_WritesConfig` — controller writes config excluding the deleted resource while keeping surviving ones
  - `TestReconcile_VirtualServerDeletion_EmptyConfigWhenLast` — deleting the last virtual server writes an empty config, not a no-op

## Summary by CodeRabbit

* **Bug Fixes**
  * Removed stale virtual server entries so deleted servers no longer persist in the active configuration.
  * Ensure virtual server configuration is generated and persisted during resource deletion, with proper handling of write conflicts and requeueing.

* **Tests**
  * Added tests validating virtual server removal, config writes during deletion, and empty config behavior when the last server is deleted.


ayuxsh009:
Hi @david-martin Can you please review it when you got a chance? Thanks!

david-martin:

ayuxsh009:
No worries, thanks for letting me know!


> Hey, thanks for the contribution. Just a heads up: reviews are currently being prioritised around the v0.7 milestone. PRs tied to v0.7 issues will be reviewed first. We'll get to this one, but it may take a bit longer than usual.

jasonmadigan (approved):

david-martin (approved):
Fix looks good, and missing impl in the controller is good too

PR #986, opened 2026-05-17
poc: add A2A federated agent card broker

Aman-Cool:
This is a proof-of-concept for the LFX mentorship project "Prototype A2A protocol support in the agentic gateway" ; put together to demonstrate my understanding of the codebase and the approach I'd take before the actual mentorship begins.

While going through the broker code trying to understand the project well enough to write a real implementation plan, something clicked when I was reading `OnConfigChange` in `broker.go`. The whole MCP tool federation story is essentially: fetch capabilities from upstream servers concurrently, apply a prefix to avoid naming collisions, merge everything into a single surface the client sees. At that point I started wondering ; what does A2A actually need that's different? And the honest answer for the discovery piece is: not much.

A2A agents advertise themselves via an `AgentCard` at `/.well-known/agent.json` the same way MCP servers advertise tools. A gateway that registers multiple upstream agents should fetch those cards, prefix the skill IDs, and serve a merged card at its own well-known endpoint. That's the same federation problem, just with a different wire format.

So that's what this does. `FederatedCard()` in `internal/a2a/broker.go` is the core ; it spins up one goroutine per registered agent, fetches the card, applies the `SkillPrefix`, and merges the results. If an upstream fails, we log it and skip it; the healthy agents still contribute. The `A2AAgent` config struct in `internal/config/a2a_types.go` mirrors `config.MCPServer` field for field because the controller will eventually resolve it the same way; `HTTPRoute` -> hostname -> URL.

I also added an `A2AAgentRegistration` CRD in `api/v1alpha1/` following the same pattern as `MCPServerRegistration`; same `TargetRef`, same `CredentialRef`, same status conditions shape. The deepcopy is hand-written for now since I can't run `make generate-all` without the CRD controller wired up; that gets replaced the moment this goes beyond PoC.

---

### What this PR is

A focused proof-of-concept showing that A2A agent card federation maps cleanly onto the existing MCP tool federation pattern; same concurrent fetch model, same prefix-based collision avoidance, same failure tolerance. The intent is to validate the architectural approach before building the router integration and controller on top of it.

---

### What this PR doesn't include

- **Router integration** : wiring `A2ABroker` into `ExtProcServer.Process()` to detect `/a2a` prefixed traffic and route it separately from MCP. That's the next layer and involves changes to `server.go`, `request_handlers.go`, and `headers.go`.
- **CRD controller** : an `A2AAgentRegistration` reconciler that watches `HTTPRoutes`, resolves backend URLs, and calls `SetAgents()` on the broker. The types are here; the controller is not.
- **Task proxying** : forwarding `message/send`, `tasks/get`, and other A2A JSON-RPC methods to the correct upstream agent after resolving the skill prefix.
- **Task ID mapping** : the equivalent of `idmap.Map` for A2A task IDs, which the gateway will need to rewrite to maintain task identity across upstream boundaries.
- **Session handling** : whether A2A requests reuse the existing `mcp-session-id` JWT or need their own session initialization.

---

### Open questions for mentors

**1. CRD group: `mcp.kuadrant.io` or a new `a2a.kuadrant.io`?**
I placed `A2AAgentRegistration` under `mcp.kuadrant.io` to stay consistent with existing types, but A2A is a distinct protocol and might warrant its own API group; especially if the intent is to eventually contribute this upstream to kube-agentic-networking SIG. This decision locks in the scheme registration, controller setup, and every YAML sample from day one, so I don't want to guess.

**2. Which A2A spec version should the implementation target?**
I wrote `AgentCard` based on what I could piece together from the spec, but there are real differences between versions; the well-known endpoint path, field names in `Skill`, whether capabilities are a flat struct or an enum. Building the card parser and the test server against the wrong version means redoing both. Is there a specific version the team has in mind, or should we just track latest?

**3. How should A2A traffic be discriminated in the `ext_proc` path?**
My plan uses the `:path` request header prefix (`/a2a`) as the protocol discriminator in `Process()`, set before the body is parsed; same approach as the elicitation path detection. But if the gateway is expected to accept A2A traffic on the same `/mcp` path (differentiated by JSON-RPC method instead), the detection logic is completely different and lives in the body phase, not the headers phase. This shapes the entire router implementation.

**4. Should A2A reuse `mcp-session-id` or have its own session?**
Right now the JWT only carries MCP session claims. If A2A reuses it, a client that completed MCP `initialize` automatically gets A2A access on the same session; convenient, but the access model is implicit. If A2A needs its own session initialization, there's a new handshake to implement and a new set of JWT claims to define. I can build either, but the session cache schema and the JWT manager changes are very different depending on the answer.

**5. Who owns the task ID, the gateway or the upstream?**
When a client sends `message/send`, does the gateway generate its own task ID and maintain a mapping to the upstream task ID (like `idmap.Map` does for elicitation request IDs), or does it just pass the upstream's task ID through to the client? If the gateway owns the ID, it can guarantee uniqueness across multiple upstream agents and the client never sees backend task IDs. If it passes through, the implementation is simpler but task IDs from different upstream agents could collide on the client side. Without knowing which model the team prefers, I can't design the task store correctly.

## Summary by CodeRabbit

* **New Features**
  * Introduced Agent-to-Agent (A2A) federation enabling the gateway to fetch and merge agent cards from multiple upstream agents while applying per-agent skill ID prefixes
  * Added `A2AAgentRegistration` Kubernetes custom resource definition for managing upstream A2A agent configurations including credentials and discovery endpoints

* **Documentation**
  * Added A2A module documentation with architectural overview and usage examples

[Review Change Stack](https://app.coderabbit.ai/change-stack/Kuadrant/mcp-gateway/pull/986?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


Aman-Cool:
Hey @david-martin, would love to get your thoughts on the open questions at the bottom; especially the CRD group and the task ID ownership ones, those two will shape most of what comes next...

david-martin:

Aman-Cool:
Closing this one out.., it's done its job. This PoC was where I first validated that A2A agent-card federation worked end to end, and a lot of what it surfaced fed straight into the design doc (#1114). The design has since moved past the merged-federated-card approach to path-per-agent discovery (per-agent cards + an API catalog), so the PoC itself is superseded now... but it did exactly what a PoC should: proved the concept and de-risked the direction before the real design work began. The work continues in the design doc (#1114) and the A2A test server (#1200). Thanks to everyone who took a look 🙏

Issue #804, opened 2026-04-24
Implement AccessPolicy CRD for tool-level authorization


david-martin:
## Summary

Implement, as experimental (not production use), [AccessPolicy CRD](https://github.com/kubernetes-sigs/kube-agentic-networking/blob/main/docs/proposals/0008-ToolAuthAPI.md) from the kube-agentic-networking project. AccessPolicy enables per-tool authorization for MCP servers — e.g., agent A can call `add` and `subtract` on `math-server`, while agent B can only call `subtract`.

This is a protocol-aware authorization layer that sits between the caller identity and the specific MCP tool being invoked, enforcing least-privilege access at the tool-function level rather than just the network endpoint.

## Motivation

MCP Gateway can currently handle authentication and tool based authorization using AuthPolicy/Authorino and keycloak. However, there is *no built* in mechanism for tool based authorization.
The kube-agentic-networking SIG is defining a standard API for this (`AccessPolicy` + `Backend` CRDs). MCP Gateway could be well-positioned to implement this because the ext-proc already parses MCP request bodies and exposes tool metadata via headers (see below).

## Current State (what we already have)

The MCP Gateway ext-proc (`internal/mcp-router/`) already:
- Parses JSON-RPC request bodies to extract MCP method (`tools/call`, `initialize`, etc.)
- Extracts the tool name from `tools/call` requests
- Sets headers that downstream filters can consume:
  - `x-mcp-method` — the MCP method (e.g. `tools/call`)
  - `x-mcp-toolname` — the tool being called (e.g. `add`)
  - `x-mcp-servername` — the target MCP server
  - `x-mcp-annotation-hints` — tool annotations (readOnly, destructive, etc.)
- Runs **before** AuthPolicy/Authorino in the filter chain (ext_proc runs first, then ext_authz)

This means Authorino already has access to the tool name and server name via request headers when it evaluates AuthPolicy rules. The missing piece is an AccessPolicy controller that translates tool-level authorization rules into the appropriate AuthPolicy/Authorino configuration.

## Design Considerations

This feature needs a design doc before implementation. Key questions to explore:

### 1. How does AccessPolicy map to AuthPolicy/Authorino?

The AccessPolicy CRD defines rules like "ServiceAccount X can call tools A, B on Backend Y". These need to be translated into Authorino authorization checks. Options include:

[...]

PR #995, opened 2026-05-18
Add MCP audit trail: identity headers, parameter logging, and access log

david-martin:
Implement the auditing feature from docs/design/auditing/.

Router changes:
- Parse W3C Baggage header for user.id and agent.id
- Identity fallback chain: baggage -> explicitly configured auth-layer headers (no defaults)
- Opt-in tool call parameter extraction (truncated to 1KB)
- Set audit data as ext_proc dynamic metadata (namespace `mcp.audit`) on ProcessingResponse, not as headers
- Dynamic metadata is consumed by Envoy access logs but not forwarded to upstream servers
- AuditConfig on ExtProcServer, configured via env vars

CRD changes:
- ParameterLoggingPolicy enum (Enabled/Disabled) on AuditConfig
- spec.audit field on MCPGatewayExtension
- identityHeaders with no defaults (operators must explicitly list their auth layer headers)
- Operator injects MCP_AUDIT_PARAMETER_LOGGING and MCP_AUDIT_IDENTITY_HEADERS

Operator changes:
- Access log ConfigPatch (NETWORK_FILTER/MERGE) added to EnvoyFilter when spec.audit is set
- JSON format uses `%DYNAMIC_METADATA(envoy.filters.http.ext_proc:mcp.audit:...)%` for audit fields
- Routing headers (`x-mcp-method`, `x-mcp-toolname`, `x-mcp-servername`) remain as `%REQ(...)%`

Security:
- Dynamic metadata eliminates header laundering (untrusted baggage re-emitted as authoritative headers) and parameter leakage (tool arguments in headers exposed to proxies, CDNs, upstream apps)
- No default identity headers forces operators to explicitly tie audit config to their auth setup


Rul1an:
This looks like a solid request-side audit MVP.

Two small things I’d decide before the contract settles:

- `x-mcp-tool-params` is written as a request header for logging. If params are only meant for the audit projection, a non-forwarded surface would reduce where sensitive arguments can travel.
- Some immediate-response paths appear to bypass audit header population, for example invalid session or tool/prompt not found. Those failed attempts are audit-relevant too, even if the row only has method/session/identity plus transport status.

The response-leg outcome split can stay a follow-on; these just seem worth tightening inside the MVP.

david-martin:
@Patryk-Stefanski @maleck13 @jasonmadigan @trepel 

Been going back and forth on this PR over the last week and wanted to share where I'm landing.

The original approach was the router sets audit headers (user identity, agent identity, tool params) on every request, and the operator wires up an EnvoyFilter access log to capture them.
After getting a more thorough understanding of the approach, particularly from a security point of view, there's a few things making me hesitate:

- Header laundering: audit headers set via ext_proc get forwarded to upstream MCP servers. Client-supplied baggage values (not necessarily trusted) get re-emitted as x-mcp-user-id (looks authoritative). Upstreams could trust
  them when they shouldn't.
- Parameter leakage: tool call arguments copied into headers expand the surface for sensitive data exposure (proxy logs, CDN caches, upstream app logs). I know i was proposing a flag here, but i'm backing away from adding this capability as a security stance.
- Identity trust model: without an AuthPolicy or header stripping, identity headers are client-controllable. We had default fallback headers (x-forwarded-email, x-auth-user) that would silently trust whatever the client sends. We have no 'one size fits all' as users will deploy the gateway in different auth setups.

Switched to dynamic metadata (consumed by access logs, not forwarded to upstreams), fixes some security issues. But the routing headers we already set (x-mcp-method, x-mcp-toolname, x-mcp-servername, mcp-session-id) are available to any access log today via %REQ(...)%.
The operator-managed access log saves writing one EnvoyFilter, so questionable value.

The real value, "who called this tool," only works reliably with an auth layer in place. And if you have AuthPolicy/Authorino, it already validates identity and can inject trusted headers. (you can also get info in authorino logs, but you have a correlation problem to solve then)
Istio's Telemetry API is the right tool for configuring the access log, not a custom EnvoyFilter from our operator.

My suggestion: 

close this PR and issue, and create a new issue for a guide showing how to get reliable MCP auditing using AuthPolicy (for trusted identity headers) + Istio Telemetry resource (for access log configuration).
No new CRD fields at this time.
We can revisit something more built-in when we get to the AccessPolicy CRD work (#804), where identity and audit have a more structured path.

david-martin:
> * `x-mcp-tool-params` is written as a request header for logging. If params are only meant for the audit projection, a non-forwarded surface would reduce where sensitive arguments can travel.

Agreed. Dynamic metadata would avoid this.
As per comment above, I'm tending away from parameter logging at this time.

> * Some immediate-response paths appear to bypass audit header population, for example invalid session or tool/prompt not found. Those failed attempts are audit-relevant too, even if the row only has method/session/identity plus transport status.

Good catch. These should be considered as well.

Patryk-Stefanski:
+1 for leaving this work off till the access policy and creating a doc for the interim

Rul1an:
That direction makes sense to me. AuthPolicy + Istio Telemetry feels like a cleaner interim boundary than adding a partial audit contract to `MCPGatewayExtension`, especially if parameter capture is not something the gateway wants to own yet.

For the guide, the useful line to keep crisp is probably:

- trusted identity comes from AuthPolicy / Authorino-managed headers
- baggage stays informational/correlation context unless something trusted validates it
- `response_code` queries are HTTP/transport-level audit, not full tool-outcome audit

Richer built-in audit semantics can fit much better once AccessPolicy gives the gateway a stronger policy/identity surface.

david-martin:
Closing in favour of #1023

PR #1005, opened 2026-05-19
fix: protect MCPManager.status with statusMu to eliminate data race

Aman-Cool:
While I was chasing the `singleflight` fix in #976 I kept `go test -race` running in the background and noticed it lighting up somewhere I hadn't looked yet; not in the session cache, not in the init path, but in the health check...

`MCPManager` runs an event loop in a background goroutine. Every 60 seconds (or whenever a `notifications/tools/list_changed` comes in) it calls `manage()`, which eventually calls `setStatus()`. That function writes eleven fields onto `man.status` one by one; including a `time.Time` and two slices; with no lock held.

Meanwhile, every Kubernetes readiness probe hits `/readyz`, which calls `IsReady()`, which calls `GetStatus()` on every registered manager. `GetStatus()` just does `return man.status` ; a bare struct copy, also with no lock. The comment even says "no locking is done here as it is expected to be called multiple times", as if frequency is the reason to skip synchronization rather than the reason to add it.

The struct copy and the field-by-field writes race. The `Ready` bool can flip the wrong way for one probe interval and cause Kubernetes to briefly drain the pod. The slice headers (`InvalidToolList`, `InvalidPromptList`) are 24 bytes each; a torn read gives a valid pointer with a stale length, and anything that iterates the result ; like the JSON encoder in `ValidateAllServers()` ; walks off the end.

The fix adds a dedicated `statusMu sync.RWMutex` (separate from `toolsLock` to keep the tools hot path unaffected), acquires it in `setStatus()` and `GetStatus()`, and covers `SetStatusForTesting()` too so race-detected test runs stay clean.

### DIFF
--- a/internal/broker/upstream/manager.go
+++ b/internal/broker/upstream/manager.go
@@ -129,6 +129,8 @@ type MCPManager struct {
 
 	// toolsLock protects tools, serverTools, prompts, serverPrompts
 	toolsLock sync.RWMutex
+	// statusMu protects status; kept separate from toolsLock to avoid contention
+	statusMu sync.RWMutex
 
 	logger *slog.Logger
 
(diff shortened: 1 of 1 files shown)


Aman-Cool:
Hey @maleck13 and @david-martin, Noticed this while the race detector was still running after #976 ; `setStatus()` writes eleven fields including slices and a `time.Time` from the event loop, while `/readyz` and `/status` read the whole struct on every probe with nothing between them. Added a dedicated `statusMu` so it stays out of `toolsLock`'s way. One mutex, three sites. Tests pass clean under `-race`.

david-martin:

SeeyaVhora:
This PR covers the same fix. See #1018 which also protects SetStatusForTesting — the third unguarded access path.

Aman-Cool:

PR #1009, opened 2026-05-19
fix: serialize concurrent manage() execution in MCPManager

Aryanburnwal05:
This fixes a race condition in MCPManager where manage() could run concurrently from multiple goroutines.

The issue was happening because manage() could be triggered simultaneously by :-

1.  the periodic ticker loop
2.  upstream notification callbacks

Both paths were accessing and updating shared manager state (serverTools / serverPrompts) at the same time.

Changes

1. added a dedicated manageMu mutex to serialize manage() execution
2. synchronized reads in shouldFetchTools() and shouldFetchPrompts() using read locks
3. added a concurrent regression test covering overlapping manage() calls

Why this approach?

manage() performs multiple state update steps internally, so serializing the full execution path keeps the manager state consistent and avoids overlapping refresh cycles.

The lock is scoped per MCPManager instance, so different upstream servers continue operating independently without blocking each other.

Testing

Verified with :-

1. go test ./...
2. added concurrent regression coverage for repeated parallel manage() execution

Unable to run go test -race locally due to Windows CGO/GCC environment limitations.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved stability by preventing race conditions during concurrent operations

* **Tests**
  * Added tests for concurrent operation scenarios

[Review Change Stack](https://app.coderabbit.ai/change-stack/Kuadrant/mcp-gateway/pull/1009?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/internal/broker/upstream/manager.go
+++ b/internal/broker/upstream/manager.go
@@ -130,6 +130,9 @@ type MCPManager struct {
 	// toolsLock protects tools, serverTools, prompts, serverPrompts
 	toolsLock sync.RWMutex
 
+	// manageMu protects against concurrent executions of manage()
+	manageMu sync.Mutex
+
 	logger *slog.Logger
 
 	// invalidToolPolicy controls behavior when upstream tools have invalid schemas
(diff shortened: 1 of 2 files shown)


maleck13 (commented):
The manager now uses a for select statement only one case can run at a time. So there is no need for this additional  mutex

Manage is an unexported method that cannot be called as the test is showing.

PR #1010, opened 2026-05-19
design: add protocol-neutral upstream config groundwork for A2A support

Aryanburnwal05:
This PR introduces a small set of protocol-neutral config abstractions to prepare the broker config model for future A2A upstream support without changing the current MCP behavior.

The existing config and runtime flow is heavily MCP-oriented today. Instead of forcing future A2A concepts into MCP-specific structures, this change introduces lightweight shared abstractions that can support multiple upstream protocols cleanly over time.

 What changed ?

1.  added generic upstream protocol and ID types
2.  introduced `UpstreamServer` and `UpstreamRegistry` interfaces
3.  extended existing `MCPServer` and `MCPServersConfig` implementations to satisfy the new interfaces
4.  added an initial `A2AServer` structure for future protocol support
5.  extended `BrokerConfig` with optional A2A upstream configuration support
6.  added tests covering interface compatibility and multi-protocol behavior
7.  added a design document describing the motivation, scope, and future integration direction

 Notes :-

This PR is intentionally limited in scope.

It does not implement :-

1.  A2A routing
2.  controller reconciliation
3.  runtime proxying
4.  protocol execution logic

The goal here is only to establish safer abstraction boundaries and reduce MCP-specific coupling before larger A2A work begins.

Testing

1.   go test ./...
2.   added coverage for protocol-neutral interface compatibility and registry behavior

PR #934, opened 2026-05-11
Add A2A e2e scenario docs

mugiwaraluffy56:
Adds a focused A2A e2e scenario document under the test suite so the expected project behavior is concrete before routing and broker implementation work starts.

The scenarios cover federated agent discovery, non-streaming and streaming task sends, task lookup and cancellation, agent card readiness, and a policy-denied request. The intent is to give issue #766 a shared test target that future implementation PRs can turn into executable Ginkgo coverage.

Checked the staged diff against main and ran make check-newlines. make spell could not run because cspell is not installed in this environment.

Refs #766

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive end-to-end test specifications for MCP Gateway A2A functionality, documenting expected behaviours for agent discovery, task routing with streaming support, task management and cancellation, agent status reporting, and policy enforcement handling.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Kuadrant/mcp-gateway/pull/934)

PR #1013, opened 2026-05-19
design: add initial multi-protocol config support for A2A upstreams

### LINKED ISSUE #847: Design a multi-protocol broker config model for A2A upstreams
Hi @david-martin, I wanted to open this as a focused design issue under #766 before any implementation work starts.

While going through the A2A project idea and the current mcp-gateway code, one of the first design points I noticed is that the broker config path is still very MCP-specific. The shared config secret currently maps into BrokerConfig with servers and virtualServers, and the runtime config uses MCPServersConfig, MCPServer, UpstreamMCPID, and GetServerConfigByName. The broker and router consume that shape directly.

For A2A support, it feels risky to force agents into the existing MCPServer model. A2A needs different concepts: agent card URL, task endpoint, supported protocol binding, agent or skill identity, task routing metadata, and possibly different status fields. If those get squeezed into MCP fields like toolPrefix or MCP path, the first prototype may work, but the API and config model could become hard to evolve.

I think it would be useful to agree on the config shape before adding broker, controller, or router code for A2A.

Some questions this issue could cover:

- should the config secret get a separate agents section for A2A upstreams
- should MCP and A2A share a generic upstream model with a protocol field
- what fields does the broker need for agent card federation vs request routing
- how should credentials and enabled state map across MCP and A2A
- how should the router look up A2A target information without depending on MCPServer
- what status data should be available later for agent card readiness and discovery

A first design pass could map the current BrokerConfig and MCPServersConfig flow to a proposed A2A-aware model, then call out the smallest change needed for the initial prototype.

This is not meant to replace #766. It is just a smaller design thread for one of the first pieces that seems to affect the rest of the A2A work.

Aryanburnwal05:
Resolves #847

This PR adds the initial configuration groundwork needed for future A2A upstream support while keeping the current MCP flow unchanged.

The changes are intentionally small and additive so we can start introducing protocol-aware upstream configuration without forcing a larger refactor too early.

Changes included

1. added a lightweight UpstreamServer interface for shared upstream access patterns
2. introduced a minimal A2AServer config model
3. updated broker config loading to recognize a2aServers
4. extended config mapping without changing the existing MCP server flow
5. added coverage to ensure current config loading behavior still works as expected

Notes
The existing MCPServersConfig structure and Servers field were intentionally kept as-is to avoid unnecessary churn and preserve backward compatibility while the broader A2A work is still evolving.

The new A2A-related types are also intentionally minimal for now so the config model can evolve naturally alongside the routing and broker implementation work.

Validation

1. ran go fmt ./...
2. ran go test ./...
3. verified existing MCP configs continue to load without changes

### DIFF
--- a/cmd/mcp-broker-router/main.go
+++ b/cmd/mcp-broker-router/main.go
@@ -422,6 +422,15 @@ func LoadConfig(path string) {
 	if err != nil {
 		log.Fatalf("Unable to decode server config into struct: %s", err)
 	}
+
+	var newA2AServers []*config.A2AServer
+	if viper.IsSet("a2aServers") {
+		err = viper.UnmarshalKey("a2aServers", &newA2AServers)
+		if err != nil {
+			log.Fatal("Failed to parse a2aServers configuration", "error", err)
+		}
+	}
+
 	var newVirtualServers []*config.VirtualServer
 	// Load virtualServers if present - this is optional
 	if viper.IsSet("virtualServers") {
@@ -432,7 +441,7 @@ func LoadConfig(path string) {
 	} else {
 		logger.Debug("No virtualServers section found in configuration")
 	}
-	mcpConfig.SetServers(newServers, newVirtualServers)
+	mcpConfig.SetServers(newServers, newA2AServers, newVirtualServers)
 
 	logger.Debug("config successfully loaded", "# servers", len(newServers))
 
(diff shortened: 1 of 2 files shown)


david-martin:

jasonmadigan:
thanks @Aryanburnwal05, closing -- A2A design work is being led by the core team via #766. we closed the other A2A-related PRs (#1010, #934) for the same reason.

Issue #960, opened 2026-05-13
Add e2e test tiers to reduce PR CI time


maleck13:
## Problem

E2E test execution time is growing as we add more tests. Every PR runs the full suite, slowing down the feedback loop for contributors.

## Proposal

Introduce two e2e test tiers:

**Tier 1 — PR gate (required, every PR)**
- Core routing and broker functionality
- Session management basics
- Critical auth paths
- Tool listing and tool/call happy paths

**Tier 2 — Nightly/opt-in (scheduled, or manually triggered)**
- Edge cases and error handling
- Performance/slow tests
- External service integration
- Full cross-cutting scenarios (e.g. multi-server federation, custom paths)

## Implementation ideas

- Use Ginkgo labels (e.g. `Label("tier1")`, `Label("tier2")`) to tag tests
- Tier 1 runs via `make test-e2e` (default, used in PR CI)
- Tier 2 runs via `make test-e2e-full` or a label filter override
- Add a nightly GitHub Actions workflow for the full suite
- Add a `/test-e2e-full` comment trigger for opt-in full runs on PRs

## Acceptance criteria

- [ ] Tests are labeled with tier annotations
- [ ] Tier 1 suite identified and kept under a target time (e.g. 5 min)
- [ ] PR CI only runs tier 1
- [ ] Nightly workflow runs both tiers
- [ ] Docs updated with guidance on which tier new tests belong in

manik3160:
Hey @maleck13 ! I have opened up a PR regarding this plz review :)

PR #1023, opened 2026-05-21
Add auditing guide for MCP tool call logging

### LINKED ISSUE #994: Auditing: user guide for MCP tool call audit trail
Parent: #710

## Summary

Write an auditing guide that explains how to configure structured JSON access logs for MCP tool calls with caller identity.

## Approach

Rather than adding new CRD fields or router code, the guide uses existing infrastructure:

- **MCP routing headers**: `x-mcp-method`, `x-mcp-toolname`, `x-mcp-servername`, `mcp-session-id` (already set by the router on every request)
- **AuthPolicy**: validates JWT and injects caller identity as `x-auth-identity` request header
- **Istio Telemetry API**: configures structured JSON access logging on the gateway, capturing MCP routing headers and the identity header

## Acceptance criteria

- [x] Guide published at `docs/guides/auditing.md`
- [x] Guide includes Istio MeshConfig extension provider and Telemetry resource examples
- [x] Guide includes AuthPolicy configuration for identity injection
- [x] Guide follows existing docs style (see `docs/CLAUDE.md`)

david-martin:
## Summary

Adds an auditing guide showing how to get structured JSON access logs with MCP context and caller identity using existing infrastructure (AuthPolicy + Istio Telemetry API). No new CRD fields or operator code.

This replaces the built-in audit trail approach from #995, which was closed due to security concerns (header laundering to upstream servers, tool call parameters leaking into proxy-visible headers, untrusted identity defaults from client-supplied baggage). The MCP routing headers (`x-mcp-method`, `x-mcp-toolname`, `x-mcp-servername`, `mcp-session-id`) are already set by the router on every request and available to any access log via `%REQ(...)%`. This guide shows how to combine them with trusted identity from AuthPolicy and structured logging from the Telemetry API.

A more built-in solution will be revisited with the AccessPolicy CRD work (#804).

## Changes

- **`docs/guides/auditing.md`** (new): How-to guide covering AuthPolicy identity injection (dual listener), MeshConfig extension provider, Telemetry resource, verification, and example queries
- **`docs/guides/README.md`**: Add auditing guide link
- **`config/istio/istio.yaml`**: Add `mcp-json-access-log` extension provider to MeshConfig
- **`docs/design/auditing/`** (removed): Design docs for the rejected built-in approach

## Closes

Closes #994

## Test plan

- [x] `make lint` passes
- [x] Guide verified end-to-end on a fresh Kind cluster (`make local-env-setup-olm && make auth-example-setup-no-vault`)
- [x] All four steps produce expected results: AuthPolicy enforced, identity header injected, Telemetry resource applied, JSON access log entries with all MCP fields populated
- [x] Review guide against `docs/CLAUDE.md` conventions


tests/e2e/test_cases.md
```diff
@@ -297,6 +297,10 @@
 
 - When the `category` or `hint` fields on a live MCPServerRegistration are updated, the controller should re-reconcile and the broker should reflect the new metadata in subsequent `discover_tools` calls. The old category should no longer match.
 
+### [Auditing] Test MCP audit trail in access logs
+
+When a Telemetry resource is configured for JSON access logging on the gateway and a client makes a tool call, the gateway access log should contain a JSON entry with the MCP method, tool name, server name, and session ID. These fields come from the MCP routing headers already set by the router on every request.
```
david-martin: This test is of questionable value in it's current state as it's really testing Istio logging behaviour.
It's not using an AuthPolicy either right now to expose the user id for logging.
Investigating options here...

david-martin: Related: #960 proposes e2e test tiers with Ginkgo labels so heavier tests (auth infra, external services) run nightly rather than on every PR.

This test could be a tier 2 candidate once that's in place.
Worth noting: this test doesn't exercise any mcp-gateway code. It validates the guide's Istio Telemetry + access log recipe (that MCP routing headers appear in structured logs).
@trepel thoughts on if there is sufficient value in having this test regardless, given it's really testing the guide's infrastructure setup rather than our code?

Rul1an:
This replacement direction reads much cleaner to me than the built-in audit CRD path. It keeps the boundary boring in the good way: trusted identity via AuthPolicy, MCP routing context via existing headers, structured output via Telemetry, and no parameter capture owned by the gateway.

Two wording/test details I'd keep tight while the guide settles:

- I'd describe `response_code` as the HTTP/access-log outcome, not the auth outcome. It can show denied requests and transport failures, but it does not carry the richer AuthPolicy decision semantics.
- The `x-auth-identity` field is the right bridge, as long as the guide is explicit that it is trusted because the auth layer overwrites or removes any client-supplied value before logging. If that behavior is guaranteed by Authorino, great; if not, it may be worth softening the wording or adding a small spoofing check.

That aside, this is exactly the cleaner interim route I was hoping #995 would collapse into.

Rul1an:
Small note on the e2e failure: this may be comparing two different session concepts.

For `tools/call`, the router appears to forward `mcp-session-id` as the remote backend session id. Since the access log reads `%REQ(MCP-SESSION-ID)%`, the logged value can differ from the gateway/client session returned by `mcpInitialize()`.

So the strict equality check may need to either go back to presence/non-empty, or compare against the backend session reported by the headers tool. That would also make the guide clearer about which session id the log field represents.

docs/guides/auditing.md
```diff
@@ -0,0 +1,316 @@
...
+```
+
+You should see `x-auth-identity` with the Keycloak username in the response.
+
+## Step 2: Add MeshConfig Extension Provider
+
+Add a custom access log provider to Istio's MeshConfig. This provider defines the JSON format for the audit log.
+
+```bash
+kubectl patch istio default -n istio-system --type='merge' \
```
trepel, line 150: This is ok for KIND/upstream but on OpenShift (if using Openshift Ingress) the Ingress operator reconciles the Istio CR back I think so this won't work. Not sure if that's an issue, but it's definitely something to bear in mind when this guide is converted to downstream

david-martin: Agreed. fyi @ShaunaDiaz 
(also, in case you missed it, the audit log feature has been slimmed back to just a new guide, no new features per-se. Taking advantage of what's already available)

docs/guides/auditing.md
```diff
@@ -0,0 +1,316 @@
...
+kubectl logs -n gateway-system -l gateway.networking.k8s.io/gateway-name=mcp-gateway --since=1h \
+  | grep '"mcp_method"' | jq 'select(.mcp_user_id == "mcp")'
+
+# All calls to a specific tool
+kubectl logs -n gateway-system -l gateway.networking.k8s.io/gateway-name=mcp-gateway --since=1h \
+  | grep '"mcp_method"' | jq 'select(.mcp_tool_name == "greet")'
+
+# All calls to a specific backend server
+kubectl logs -n gateway-system -l gateway.networking.k8s.io/gateway-name=mcp-gateway --since=1h \
+  | grep '"mcp_method"' | jq 'select(.mcp_server_name == "mcp-test/server1")'
```
Patryk-Stefanski:
```suggestion
  | grep '"mcp_method"' | jq 'select(.mcp_server_name == "mcp-test/test-server1")'
```

Patryk-Stefanski (approved):

docs/guides/auditing.md
ShaunaDiaz: So how should OpenShift users compete this step, @david-martin? Can we just use the Telemetry object for specific namespaces, or do we need to update the OSSM Operator in some way?

PR #1037, opened 2026-05-25
fix(session): set Redis TTL on session entries to match JWT expiry

Aman-Cool:
So I was looking at what happens to Redis when a broker pod goes down mid-session and noticed something a bit uncomfortable.

When `initializeMCPSeverSession` stores a backend session ID in Redis, it relies entirely on a `time.AfterFunc` to clean it up later. That timer lives in the pod's memory. If the pod crashes or rolls during a deployment, the timer dies with it ; but the Redis key stays around forever with no TTL set. Same story for the elicitation flag key.

So over time, especially across rolling updates (which happen on every release), Redis quietly accumulates stale session hashes that nobody ever cleans up. Worse, if a client's JWT is still valid and they hit a new pod, that pod finds the stale backend session ID in Redis and tries to use it, which the upstream MCP server rejects because that SSE connection died with the old pod.

The fix is straightforward: thread the JWT expiry time through as a TTL when writing to Redis. `AddSession` now runs an `HSet` + `Expire` in a pipeline, and `SetClientElicitation` passes the TTL to `SET` instead of `0`. The `time.AfterFunc` is still there ; it closes the upstream client handle on expiry, which it still needs to do, but Redis no longer depends on it for cleanup.

`SessionCache` interface got two new `ttl time.Duration` params, so all the in-memory test call sites needed updating (in-memory just ignores the value). Also added two small tests that lock in the new contract.

## Release Notes

* **Bug Fixes**
  * Corrected session cache encryption key derivation to use the appropriate signing mechanism.

* **New Features**
  * Added automatic session expiration support with configurable time-to-live (TTL) for cached sessions and client elicitation data, now tied to JWT token expiry.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Kuadrant/mcp-gateway/pull/1037?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/cmd/mcp-broker-router/main.go
+++ b/cmd/mcp-broker-router/main.go
@@ -195,7 +195,7 @@ func main() {
 	var sessionCacheOpts []func(*session.Cache)
 	sessionCacheOpts = append(sessionCacheOpts, session.WithRedisClient(redisClient))
 	if redisClient != nil {
-		encKey, encErr := session.DeriveEncryptionKey([]byte(jwtSigningKeyFlag))
+		encKey, encErr := session.DeriveEncryptionKey([]byte(gatewaySigningKeyFlag))
 		if encErr != nil {
 			panic("failed to derive encryption key: " + encErr.Error())
 		}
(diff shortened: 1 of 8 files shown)


Aman-Cool:
Hey @david-martin and @maleck13, quickest way to verify the Redis side is spinning up a local Redis and watching `TTL <gateway-session-id>` in redis-cli right after a tool call, you should see it tick down matching the JWT expiry instead of returning `-1`. that's the before/after in one command.

maleck13 (approved):
great. Thanks!

PR #846, opened 2026-05-01
Feat: wire OTel metrics, instrument tool call routing, fix status N+1

Aman-Cool:
Three things that were either missing or broken enough to bother me:

**1. Metrics were never wired up**

Traces and logs had full provider setup. Metrics had the OTel metric API as an indirect dep but no `MetricsProvider`, no `MeterProvider` init, nothing. Added `internal/otel/metrics.go` mirroring the existing logs/traces pattern, hooked it into `SetupOTelSDK`, and added `MetricsEndpoint()` / `MetricsEnabled()` to config. Controlled via `OTEL_EXPORTER_OTLP_METRICS_ENDPOINT` (or falls back to the base `OTEL_EXPORTER_OTLP_ENDPOINT`).

**2. Tool call routing had zero observability**

`HandleToolCall` is the hot path and we had no idea how long routing was taking or which servers were slow/erroring. Added a `mcp.router.tool_call.duration` histogram (seconds) with `mcp.server` and `status=ok|error` attributes. Used a deferred closure so we don't have to sprinkle `Record()` across every return path.

**3. `/status/{server}` was calling `ValidateAllServers` to look up one server**

That pings every upstream backend just to return status for one. Replaced it with a simple loop over `RegisteredMCPServers()` that breaks on first match; no backend calls.

---

To enable metrics, set `OTEL_EXPORTER_OTLP_METRICS_ENDPOINT=rpc://otel-collector:4317` (gRPC) or `http://otel-collector:4318` (HTTP). If you already have `OTEL_EXPORTER_OTLP_ENDPOINT` set, metrics will pick it up automatically.

## Summary by CodeRabbit

* **New Features**
  * Added OpenTelemetry metrics support with configurable metrics endpoint and runtime initialization for router performance measurement.

* **Improvements**
  * Router now records tool-call duration metrics for per-route telemetry.
  * Tracing instrumentation name consolidated for consistency.

* **Bug Fixes**
  * Session cleanup switched to a non-request-scoped context to avoid premature cancellation.

* **Tests**
  * Added unit tests for metrics config, provider creation, shutdown, and SDK setup with metrics enabled.

* **Chores**
  * Updated OpenTelemetry dependency versions.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Kuadrant/mcp-gateway/pull/846?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

PR #1044, opened 2026-05-26
feat: add OTel metrics for router, broker, session, and upstream manager

Aman-Cool:
I opened #846 a while back to wire up metrics, but it only got as far as a single histogram on `HandleToolCall` before I got pulled into other things and it stalled.

Came back to it this week and decided to actually read through the full data path before writing another line; router, broker, session cache, upstream manager, the whole thing. Glad I did: traces and logs had full provider setup, but metrics had the OTel metric API sitting there as an indirect dep, completely unused. No `MeterProvider`, no instruments. You could run this in production for months with zero signal on what the gateway was doing internally.

So I closed #846 and started over. The metrics here are **operational**; aimed at operators and SREs running the gateway, not per-tool-call analytics (that's still future work, tracked in the observability design doc).

**Router** : active gRPC streams (`mcp.router.stream.active`), session cache lookups with hit/miss labels (`mcp.router.session.lookups`), backend session inits (`mcp.router.session.inits`).
- *As an operator, I want the session-cache hit/miss rate so I can decide whether scaling the session store (Redis) is worth it.*
- *As an SRE, I want active `ext_proc` stream counts so I can spot stream leaks or saturation between Envoy and the router.*

**Session store** : Redis op latency histogram (`mcp.session.op.duration`) with explicit buckets, Redis path only.
- *As an SRE, I want session-store operation latency so I can alert when Redis is degrading before it shows up as request failures.*

**Upstream manager** : connection tracking (`mcp.broker.upstream.connections`) and per-server tool-fetch duration (`mcp.broker.upstream.tool_fetch.duration`).
- *As an SRE, I want upstream tool-fetch latency per server so I can set alerts on degraded backends.*
- *As an operator, I want active upstream connection counts so I can confirm every registered MCP server is actually connected.*

**Broker** : config reload counter (`mcp.broker.config.reloads`).
- *As an operator, I want a config-reload count so I can correlate sudden behavior changes with a config push.*

The hit/miss attribute options are precomputed on `ExtProcServer` to avoid per-tool-call allocations on the hot path.

The `MetricsProvider` mirrors the existing logs/traces pattern exactly.., same URL scheme switching (`rpc://` for gRPC, `http://`/`https://` for HTTP), same 30-second periodic reader, same `Shutdown()` wired into `SetupOTelSDK`. Controlled via `OTEL_EXPORTER_OTLP_METRICS_ENDPOINT` with fallback to the base `OTEL_EXPORTER_OTLP_ENDPOINT`. User-facing reference added to `docs/guides/opentelemetry.md`.

### DIFF
--- a/.github/workflows/conformance.yaml
+++ b/.github/workflows/conformance.yaml
@@ -94,19 +94,19 @@ jobs:
 
       - name: Wait for conformance server to get ready
         run: |
-          echo "Waiting for Gateway to wire up the MCP routes and broker to fetch tools..."
-          for i in {1..30}; do
+          echo "Waiting for Envoy gateway to be programmed..."
+          kubectl wait --for=condition=Programmed gateway/mcp-gateway -n gateway-system --timeout=120s
+          echo "Waiting for MCP route to become active..."
+          for i in {1..60}; do
             HTTP_STATUS=$(curl --max-time 2 -s -o /dev/null -w "%{http_code}" ${{ env.MCP_URL }})
             if [[ "$HTTP_STATUS" != "000" && "$HTTP_STATUS" != "0" && "$HTTP_STATUS" != "502" && "$HTTP_STATUS" != "503" ]]; then
               echo "Gateway route is active! HTTP Status: $HTTP_STATUS"
-              # Extra wait for broker to connect and fetch tools from the conformance server
-              sleep 10
               break
             fi
-            echo "Attempt $i/30: Gateway route not ready yet (Status: $HTTP_STATUS). Retrying in 2 seconds..."
+            echo "Attempt $i/60: Gateway route not ready yet (Status: $HTTP_STATUS). Retrying in 2 seconds..."
             sleep 2
-            if [ $i -eq 30 ]; then
-              echo "ERROR: Gateway routes failed to propagate after 60 seconds."
+            if [ $i -eq 60 ]; then
+              echo "ERROR: Gateway routes failed to propagate after 120 seconds."
               exit 1
             fi
           done
--- a/.github/workflows/conformance.yaml
+++ b/.github/workflows/conformance.yaml
@@ -96,7 +96,7 @@ jobs:
         run: |
           echo "Waiting for Gateway to wire up the MCP routes and broker to fetch tools..."
           for i in {1..30}; do
-            HTTP_STATUS=$(curl --max-time 2 -s -o /dev/null -w "%{http_code}" ${{ env.MCP_URL }} || echo "000")
+            HTTP_STATUS=$(curl --max-time 2 -s -o /dev/null -w "%{http_code}" ${{ env.MCP_URL }})
             if [[ "$HTTP_STATUS" != "000" && "$HTTP_STATUS" != "0" && "$HTTP_STATUS" != "502" && "$HTTP_STATUS" != "503" ]]; then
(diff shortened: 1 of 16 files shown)


Aman-Cool:
@maleck13 and @david-martin, First attempt at this (#846) only scratched the surface ; one histogram and called it done. Spent more time reading through the full data path this time and realised how much was actually missing. Happy to iterate on naming or bucket boundaries if anything looks off.

Aman-Cool:
The E2E failure (`should disable and re-enable an MCPServerRegistration`) is a pre-existing flake ; same test, same line timed out on `feat/gateway-ca-cert-bundle` two days ago (run 26686322940), nothing to do with this PR. It hits whenever the suite runs close to the 30-minute limit.

david-martin (changes requested):
**Separate the CI fix**
The conformance wait-loop fixes (double 000, kubectl wait, etc.) are good and unrelated to metrics. Moving them to their own PR lets them land sooner without being blocked on the metrics review.

**Add documentation**
There is no user-facing reference for the new metrics: what they are called, what they measure, what units they use, or how to enable export. A section in the docs (user guide or reference) would make these actually discoverable. Without it, users have to read the source to know what is available.

**Tie metrics to user stories**
The PR description explains the motivation well, but the metric choices themselves are not grounded in specific user stories or use cases. Even 1-2 per area would help reviewers (and future contributors) understand why these metrics and not others. For example: "As an operator, I want to see cache hit rate so I can decide whether to scale the session store" or "As an SRE, I want upstream tool-fetch latency so I can set alerts on degraded backends."

**Use centralized constants in session cache**
`internal/session/cache.go` hardcodes `"mcp-session"` and `"mcp.session.op.duration"` instead of using `SessionMeterName` and `MetricSessionOpDuration` from `internal/otel/otel.go`. The other components all use the centralized constants.

**Update observability design doc**
`docs/design/observability.md` describes metrics conceptually (tool call counts, error rates, etc.) but this PR implements a different set of metrics (stream counts, cache hit/miss, config reloads, connection tracking, fetch latency). The design doc should be updated to reflect what is actually implemented now, and what remains planned.

Aman-Cool:
Thanks for the thorough review @david-martin. First thing I hit when I came back to this was that the whole branch had gone red.., turned out a stale merge from main had left a router variable half-renamed, and that one undefined symbol was knocking over the build and everything downstream. The history was a tangle too, so I rebased onto current main, dropped the old merge commits, and that brought CI back.

Then your list: the session cache was the odd one out still carrying its own const and hardcoded string, so it now pulls from `internal/otel` like everything else. Added a Metrics section to the OpenTelemetry guide (table of all seven instruments + how to enable them) since there was no way to discover these without reading the source. And updated the design doc with a "Current Implementation Status" split; what's actually built vs. the tool-call analytics that are still just a plan. PR description's reworked too, with the operator/SRE story behind each metric.

The one I left for now is splitting out the conformance CI fixes.., agreed they don't belong here. They're three commits touching only `conformance.yaml`, so once this is close I'll pull them onto their own PR and drop them from here.
lmk what you think...

Aman-Cool:

aniketpandey05:
@Aman-Cool are you looking to open this one

Aman-Cool:
@aniketpandey05, yes🙂

PR #1008, opened 2026-05-19
docs: add design spec for custom TLS configuration (#659)

Patryk-Stefanski:
# Verification Guide: Custom TLS Configuration (#659)

This guide walks a reviewer through verifying the custom CA certificate feature on a local Kind cluster.

## Prerequisites

- Kind, kubectl, Helm installed
- Docker running
- This branch checked out

## Part 1: Automated Tests

These tests cover all code paths including actual TLS connections with self-signed CAs.

```bash
# Unit tests — includes TLS connection tests with real CA/server certs
make test-unit

# Controller integration tests (envtest) — 30 specs including:
#   - caCertSecretRef happy path
#   - missing label validation
#   - default key behavior
#   - secret not found
#   - missing key in secret
#   - both credentialRef + caCertSecretRef
#   - invalid PEM data (NEW: controller-side validation)
#   - CA cert size exceeds maximum
make test-controller-integration

# Lint
make lint
```

### What the TLS unit tests prove

The tests in `internal/broker/upstream/mcp_test.go` spin up a real TLS server using `httptest.NewUnstartedServer` with a self-signed CA:

| Test | What it proves |
|------|----------------|
| `TestBuildHTTPClient_NoCACert` | No CACert → nil client (default behavior preserved) |
| `TestBuildHTTPClient_WithValidCACert` | Valid PEM → custom client created |
| `TestBuildHTTPClient_WithInvalidPEM` | Bad PEM → error returned |
| `TestBuildHTTPClient_TLSConnection` | Custom client with correct CA → TLS handshake succeeds, HTTP 200 |
| `TestBuildHTTPClient_TLSConnectionFailsWithoutCA` | No CA configured → default client rejects self-signed cert |
| `TestBuildHTTPClient_WrongCACertFailsTLS` | Wrong CA → TLS handshake fails |
| `TestBuildHTTPClient_MultiCertBundle` | CA bundle with multiple certs → correct CA trusted |

## Part 2: CRD and Controller Verification on Kind

This verifies the Kubernetes API, CRD defaults, controller reconciliation, and status conditions.

### Step 1: Set up the cluster

```bash
# Create Kind cluster, install Istio, Gateway API, build images, deploy controller + gateway
make local-env-setup

# Deploy test servers and example MCPServerRegistrations
make deploy-test-servers
make deploy-example

# Verify test server and HTTPRoute are ready
kubectl wait --for=condition=Available deployment/mcp-test-server1 -n mcp-test --timeout=60s
[...]

PR #1046, opened 2026-05-27
docs: add design spec for gateway-level CA certificate bundle

vibhor-5:
## Summary

Design proposal for adding `caCertBundleRef` to `MCPGatewayExtension` — a gateway-level CA certificate bundle that acts as a shared trust pool for all upstream MCP server connections.

## Problem

With custom TLS support (#659, #1008), each `MCPServerRegistration` can reference its own CA via `caCertSecretRef`. When many upstream servers share the same CA (OpenShift service-serving CA, shared cert-manager issuer, corporate root CA), this causes:

- **Duplication** — same CA PEM repeated N times across N registrations
- **Config secret bloat** — each CA cert (up to 64 KiB) is embedded per-server in the config YAML, putting pressure on the 1 MiB Kubernetes Secret limit
- **Operational overhead** — CA rotation requires updating N Secrets instead of one

## Proposed Solution

```yaml
apiVersion: mcp.kuadrant.io/v1alpha1
kind: MCPGatewayExtension
spec:
  targetRef:
    name: mcp-gateway
    sectionName: mcp
  caCertBundleRef:
    name: shared-ca-bundle
```

The broker loads this bundle at startup as a base trust pool. Per-server `caCertSecretRef` appends additional CAs for backends with unique CAs. This mirrors how Kubernetes works — cluster-wide CA bundle plus per-resource overrides.

## Design Document Structure

```
docs/design/gateway-ca-cert-bundle/
├── gateway-ca-cert-bundle-design.md    # Main design document
└── tasks/
    ├── tasks.md                        # 7 ordered implementation tasks
    ├── e2e_test_cases.md               # 9 E2E test cases
    └── documentation.md                # Documentation plan
```

## Key Design Decisions

- **Volume mount, not config embedding** — CA bundle is mounted as a volume on the broker pod, not embedded per-server in the config YAML. Avoids config secret bloat.
- **Additive trust pools** — per-server `caCertSecretRef` appends to the gateway bundle, never replaces. System roots → gateway bundle → per-server CA.
- **File watch for hot reload** — broker watches the mounted file for changes, rebuilds cert pool without pod restart.
- **256 KiB size limit** — larger than the per-server 64 KiB limit since bundles typically contain multiple CAs.

## Verification

This is a docs-only PR. No code changes.

Relates to #659

## Summary by CodeRabbit

* **New Features**
  * Gateway-level shared CA certificate bundle support, enabling a gateway-wide trust pool that composes additively with per-server CAs.

* **Documentation**
  * Design, implementation plan, operational guidance for rotation/propagation, validation/labeling and size constraints, and user-guide/API documentation updates.

* **Tests**
  * E2E test scenarios for trust composition, handshake failures, validation/error reporting, rotation propagation, and no-conflict cases.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Kuadrant/mcp-gateway/pull/1046?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/docs/design/gateway-ca-cert-bundle/gateway-ca-cert-bundle-design.md
+++ b/docs/design/gateway-ca-cert-bundle/gateway-ca-cert-bundle-design.md
@@ -219,6 +219,10 @@ The current design uses a Secret for the CA bundle. A future enhancement could s
 
 The broker could parse the CA certificates and emit metrics or log warnings when certificates are near expiry. This would help operators detect upcoming CA rotations before they cause outages.
 
+### Per-Server CA Deduplication
+
+When a server's `caCertSecretRef` references the same CA that is already in the gateway `caCertBundleRef`, the CA PEM is stored twice in the config — once under `gatewayCACertPEM` and once inline in the server's `caCert` field. This is functionally correct (additive trust pools) but wastes space. A future optimization could skip embedding per-server CAs that match the gateway bundle. However, this introduces complexity: Secrets may be in different namespaces, and if one Secret is deleted, the CA disappears from config even though another Secret references the same CA. Given the low likelihood of users having many duplicate CAs without coordinating, this optimization is deferred.
+
 ## Execution
 
 See:
--- a/docs/design/gateway-ca-cert-bundle/gateway-ca-cert-bundle-design.md
+++ b/docs/design/gateway-ca-cert-bundle/gateway-ca-cert-bundle-design.md
@@ -17,7 +17,6 @@ Add a `caCertBundleRef` field to `MCPGatewayExtension` that references a shared 
 - Eliminate CA PEM duplication in the config secret when multiple servers share the same CA
 - Reduce operational burden for CA rotation to a single Secret update
 - Maintain backward compatibility — existing `caCertSecretRef` on `MCPServerRegistration` continues to work identically
-- Keep the CA PEM data out of the per-server config entries in the config secret
 
 ## Non-Goals
 
@@ -64,7 +63,7 @@ spec:
     sectionName: mcp
   caCertBundleRef:
     name: shared-ca-bundle
-    key: ca-bundle.crt    # optional, defaults to "ca-bundle.crt"
+    key: ca.crt    # optional, defaults to "ca.crt"
 ```
 
 ```go
@@ -88,65 +87,64 @@ type CACertBundleReference struct {
     Name string `json:"name,omitempty"`
 
     // key is the key within the Secret that contains the CA bundle PEM data.
-    // If not specified, defaults to "ca-bundle.crt".
+    // If not specified, defaults to "ca.crt".
     // +optional
-    // +default="ca-bundle.crt"
+    // +default="ca.crt"
     Key string `json:"key,omitempty"`
 }
 ```
 
-The default key is `ca-bundle.crt` (not `ca.crt`) to distinguish from per-server CA secrets and align with common Kubernetes conventions (e.g. OpenShift's `ca-bundle.crt` in ConfigMaps).
+The default key is `ca.crt` — the same default used by per-server `caCertSecretRef`. This means a user can promote an existing per-server CA Secret to gateway-level without creating a new Secret or changing keys.
 
 #### Config Type Changes
 
-`MCPServersConfig` in `internal/config/types.go` gains a field for the gateway-level CA bundle:
+`BrokerConfig` in `internal/config/types.go` gains a field for the gateway-level CA bundle:
 
 ```go
-type MCPServersConfig struct {
-    // ... existing fields ...
-
-    // GatewayCACertBundle is the PEM-encoded CA bundle from MCPGatewayExtension.
-    // Loaded once by the broker, not embedded per-server.
-    GatewayCACertBundle string
+type BrokerConfig struct {
+    Servers          []MCPServer           `json:"servers"          yaml:"servers"`
+    VirtualServers   []VirtualServerConfig `json:"virtualServers,omitempty" yaml:"virtualServers,omitempty"`
+    GatewayCACertPEM string                `json:"gatewayCACertPEM,omitempty" yaml:"gatewayCACertPEM,omitempty"`
 }
 ```
 
-This field is not serialized into the per-server config YAML — it is passed to the broker via a separate volume mount of the CA Secret, avoiding config secret bloat.
+The gateway CA bundle PEM is written once into the existing `mcp-gateway-config` Secret alongside the server list. The broker reads it from the same config source — no separate Secrets or volume mounts. Servers that share the gateway CA no longer need individual `CACert` entries, reducing the per-server duplication from N copies to 1.
 
 ### Architecture
 
 #### How the CA Bundle Reaches the Broker
 
+The gateway CA bundle uses the same config path as everything else — the `mcp-gateway-config` Secret:
+
 ```
-MCPGatewayExtension                Broker Pod
-┌──────────────────┐         ┌────────────────────────┐
-│ caCertBundleRef: │         │                        │
-│   name: shared-  │────────▶│  Volume Mount:         │
-│         ca-bundle│         │  /etc/mcp-gateway/ca/  │
-│   key: ca-bundle │         │    ca-bundle.crt       │
-│         .crt     │         │                        │
-└──────────────────┘         │  Broker reads at       │
-                             │  startup + watches     │
-                             │  for changes           │
-                             └────────────────────────┘
+MCPGatewayExtension              Controller                     Broker
+┌──────────────────┐     ┌──────────────────────┐     ┌────────────────────┐
+│ caCertBundleRef: │     │ Reads CA Secret,      │     │                    │
+│   name: shared-  │────▶│ validates PEM,        │────▶│ Reads config.yaml  │
+│         ca-bundle│     │ writes PEM into       │     │ extracts           │
+│   key: ca.crt    │     │ config.yaml under     │     │ gatewayCACertPEM,  │
+└──────────────────┘     │ gatewayCACertPEM key  │     │ builds base cert   │
+                         └──────────────────────┘     │ pool               │
+                                                      └────────────────────┘
 ```
 
 The MCPGatewayExtension controller:
 1. Validates the referenced CA Secret exists, has the required label, contains valid PEM, and is within size limits
-2. Adds the Secret as a volume mount on the broker-router Deployment
-3. Passes the mount path via a command-line flag `--ca-cert-bundle-path=/etc/mcp-gateway/ca/ca-bundle.crt`
+2. Writes the CA PEM into the `mcp-gateway-config` Secret's `config.yaml` under the `gatewayCACertPEM` key
+3. Existing config watcher detects the Secret update and calls `mcpConfig.Notify()` to trigger observers
 
 The broker:
-1. Reads the CA bundle PEM from the file at startup
-2. Builds a base `*x509.CertPool` from system roots + gateway CA bundle
+1. Receives config via `OnConfigChange()` (the existing observer pattern)
+2. Reads `GatewayCACertPEM` from the config, builds a base `*x509.CertPool` from system roots + gateway CA bundle
 3. For each upstream server, clones this base pool and appends any per-server CA from the config (if `CACert` is set on the `MCPServer`)
-4. Watches the file for changes (Kubernetes volume mount propagation) and rebuilds the pool on update
+
+Using the existing config Secret ensures **atomic updates** — the server list and CA bundle are always in sync. There is no timing window where the broker has a new server config but not yet the CA needed to connect.
 
 #### Trust Pool Construction
 
 ```
 System Root CAs (Go default)
-         ├── + Gateway CA Bundle (from caCertBundleRef)
+         ├── + Gateway CA Bundle (from gatewayCACertPEM in config)
          │        = Base Trust Pool (shared by all servers)
          │
          ├── Server A: uses base pool only (no caCertSecretRef)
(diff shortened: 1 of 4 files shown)


maleck13:
@Patryk-Stefanski can you give this a look over also

docs/design/gateway-ca-cert-bundle/gateway-ca-cert-bundle-design.md
```diff
@@ -0,0 +1,234 @@
...
+
+    // key is the key within the Secret that contains the CA bundle PEM data.
+    // If not specified, defaults to "ca-bundle.crt".
+    // +optional
+    // +default="ca-bundle.crt"
+    Key string `json:"key,omitempty"`
+}
+```
+
+The default key is `ca-bundle.crt` (not `ca.crt`) to distinguish from per-server CA secrets and align with common Kubernetes conventions (e.g. OpenShift's `ca-bundle.crt` in ConfigMaps).
```
maleck13: while I understand the reason for this, I think it could be confusing from a user perspective to have 2 different keys depending on where the secret is added. 

Example:

I add one MCP with ca.crt. Later I need to add two more so I decide to use the GatewayExtension mechanism. I can't just copy the secret over I need to create a new one with a new key. I think this adds friction.

docs/design/gateway-ca-cert-bundle/gateway-ca-cert-bundle-design.md
```diff
@@ -0,0 +1,234 @@
...
+The MCPGatewayExtension controller:
+1. Validates the referenced CA Secret exists, has the required label, contains valid PEM, and is within size limits
+2. Adds the Secret as a volume mount on the broker-router Deployment
+3. Passes the mount path via a command-line flag `--ca-cert-bundle-path=/etc/mcp-gateway/ca/ca-bundle.crt`
+
+The broker:
+1. Reads the CA bundle PEM from the file at startup
+2. Builds a base `*x509.CertPool` from system roots + gateway CA bundle
+3. For each upstream server, clones this base pool and appends any per-server CA from the config (if `CACert` is set on the `MCPServer`)
+4. Watches the file for changes (Kubernetes volume mount propagation) and rebuilds the pool on update
```
maleck13: We should ensure that updates to this file flow through the `mcpConfig.Notify()` method to ensure synchronization on the config

docs/design/gateway-ca-cert-bundle/gateway-ca-cert-bundle-design.md
```diff
@@ -0,0 +1,234 @@
...
+
+The size limit is 256 KiB (vs 64 KiB for per-server CAs) because a shared bundle typically contains multiple CA certificates (root + intermediates, or multiple issuer CAs).
+
+### Volume Mount Strategy
+
+The CA bundle Secret is mounted as a read-only volume, not embedded in the config secret. This:
+
+1. **Avoids config bloat** — the CA PEM is not duplicated per-server in the config YAML
+2. **Enables hot reload** — Kubernetes propagates Secret updates to volume mounts (60-120s kubelet sync), and the broker watches the file for changes
+3. **Follows the existing pattern** — the aggregated credentials secret is already mounted as a volume on the broker-router Deployment
```
maleck13: `aggregated credentials secret` this is stale. There is only one secret for the broker and controller it is the full config there is no separate credential secret. 

Having a separate secret introduces a timing issue as they are likely to sync at different times:

- config updates -> manager tries to connect -> no Gateway CA -> fail
- (manager keeps trying continues to fail)
- CA updates -> manager recreatd tries to connect -> succeeds

Adding a gateway level secret should reduce the need to add a CA per server. So I am less concerned on the bloat with this approach.  Yes we are adding an additional CA to the secret but it will cover N servers.
The complication of having two config sources here is not worth it at this point in my opinion. I suggest we move to updating the gateway secret with the gateway level CA bundle and document the limitation. 
If this limitation becomes an issue for someone (they have many MCPs each with there own CA) they can always set up more than one Gateway to cover this requirement. But it would be odd IMO to have this type of setup

maleck13 (changes requested):
Good start. But I think we should not use a separate secret and instead use the main gateway mcp-config secret for the gateway CA bundle

docs/design/gateway-ca-cert-bundle/gateway-ca-cert-bundle-design.md
vibhor-5: Yeah makes sense , changed it to default ca.crt

docs/design/gateway-ca-cert-bundle/gateway-ca-cert-bundle-design.md
vibhor-5: > Having a separate secret introduces a timing issue as they are likely to sync at different times:
> 
> * config updates -> manager tries to connect -> no Gateway CA -> fail
> * (manager keeps trying continues to fail)
> * CA updates -> manager recreatd tries to connect -> succeeds

> Adding a gateway level secret should reduce the need to add a CA per server. So I am less concerned on the bloat with this approach. Yes we are adding an additional CA to the secret but it will cover N servers.
> The complication of having two config sources here is not worth it at this point in my opinion. I suggest we move to updating the gateway secret with the gateway level CA bundle and document the limitation.
> If this limitation becomes an issue for someone (they have many MCPs each with there own CA) they can always set up more than one Gateway to cover this requirement. But it would be odd IMO to have this type of setup

Yeah sounds reasonable , even with one ca for the ca bundle it'll be a net reduction ultimately .
Modified the approach to address this and update the gateway secret with the gateway level CA bundle in design as well as tasks

vibhor-5:
@maleck13 addresed the comments pls check

docs/design/gateway-ca-cert-bundle/tasks/e2e_test_cases.md
```diff
@@ -35,8 +35,8 @@ tags: Happy,CACertBundle
...
 
-### [CACertBundle] Config secret does not contain duplicate CA PEM
+### [CACertBundle] Gateway CA stored once in config, not per-server
 
-- When an MCPGatewayExtension has `caCertBundleRef` set and multiple MCPServerRegistrations exist without `caCertSecretRef`, the config secret (`mcp-gateway-config`) should NOT contain inline CA PEM data for those servers. This verifies that the gateway-level CA bundle avoids config bloat — the CA is delivered via volume mount, not embedded in the config YAML.
+- When an MCPGatewayExtension has `caCertBundleRef` set and multiple MCPServerRegistrations exist without `caCertSecretRef`, the config secret (`mcp-gateway-config`) should contain the CA PEM exactly once under the `gatewayCACertPEM` key, and individual server entries should NOT have inline `caCert` values. This verifies that the gateway-level CA bundle eliminates per-server CA duplication in the config.
```
maleck13: I think this should be an integration test in the controller. 

The e2e test we need here is to add a custom CA to the gateway . We already have a test that does this for the MCPServerReg. So  it should just be adding an additional test to only update the cert ref on the gateway extension and testing that tools also show up at the gateway

docs/design/gateway-ca-cert-bundle/tasks/tasks.md
```diff
@@ -93,63 +92,40 @@ Depends on: Task 2.
...
-
-### Task 5: MCPServerRegistration controller — stop embedding shared CAs in config
-
-Depends on: Task 2. Independent of Tasks 3-4 (controller-side only).
+Depends on: Task 2.
 
 **Files:**
 - `internal/controller/mcpserverregistration_controller.go` — modify config building:
-  1. If the server's `caCertSecretRef` points to the same Secret as the gateway's `caCertBundleRef`, skip embedding `CACert` in the config (it's already mounted as a volume)
+  1. If the server's `caCertSecretRef` points to the same Secret as the gateway's `caCertBundleRef`, skip embedding `CACert` in the per-server config (it's already in `gatewayCACertPEM`)
```
maleck13: Not sure what the plan is to achieve that? These could be in different namespaces. Validating in the controller that the CA doesn't already exist before adding it would make sense. Problem with that though is it adds complexity. If you have two secrets with the same CA and one gets deleted, the CA in the config will be removed and wont be added again unless we trigger a reconcile of the second secret. Perhaps here if we find a duplicate CA we mark the MCPServerReg as not ready, reason duplicate CA. This would force the removal of the secret with the duplicate CA by the user and keep things simple

maleck13: Lets add that as a follow on tbh. I think the likely hood of users having multiple of the same CA and not coordinating is small enough that we can just add it if it happens and add a future enhancement to consider how we want to handle that case.

maleck13 (changes requested):
almost there some pretty minor comments now. Main one about dealing with duplicate CAs . I think we should push that out to a follow on enhancement as it adds complexity when there is really only small value in handling that case

docs/design/gateway-ca-cert-bundle/tasks/e2e_test_cases.md
vibhor-5: Moved the config structure verification to a controller integration test in Task 2's acceptance criteria. The e2e coverage you described (add CA to gateway extension → tools show up) is already covered by the first test case:  `[Happy,CACertBundle] Gateway CA bundle enables TLS connection to upstream server` , which sets  caCertBundleRef  on MCPGatewayExtension with no per-server  caCertSecretRef  and verifies tools are discovered and callable through the gateway.

docs/design/gateway-ca-cert-bundle/tasks/tasks.md
vibhor-5: Done, removed it from current scope of design and added to future considerations.

vibhor-5:
> almost there some pretty minor comments now. Main one about dealing with duplicate CAs . I think we should push that out to a follow on enhancement as it adds complexity when there is really only small value in handling that case

Yuppp changed that, kindly check

docs/design/gateway-ca-cert-bundle/gateway-ca-cert-bundle-design.md
```diff
@@ -0,0 +1,231 @@
...
+When a platform engineer needs to rotate the CA certificate used by all upstream servers (e.g. cert-manager issuer renewal), they want to update a single Secret and have all servers pick up the new CA within one reconciliation cycle, instead of updating N Secrets and waiting for N independent reconciliations.
+
+### When I have servers with unique CAs alongside a shared CA
+
+When a platform engineer has most servers behind a shared corporate CA but one or two servers using their own private CA, they want to set the shared CA at the gateway level and the unique CAs per-server, so that both types work without duplication.
+
+### When I want to avoid hitting the Kubernetes Secret size limit
+
+When a platform engineer registers many servers that each embed CA PEM data in the config secret, they want the shared CA to be loaded once by the broker (not embedded N times in the config), so that the config secret stays well under the 1 MiB Kubernetes limit.
+
```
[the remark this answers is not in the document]
maleck13: this feature targets only that persona

docs/design/gateway-ca-cert-bundle/gateway-ca-cert-bundle-design.md
```diff
@@ -0,0 +1,231 @@
...
+### CA Bundle Rotation
+
+When the CA bundle Secret is updated:
+
+1. The MCPGatewayExtension controller detects the Secret update (via watch on labeled Secrets)
+2. The controller re-validates the PEM and writes the updated `gatewayCACertPEM` into the config Secret
+3. The config watcher detects the config Secret change and calls `mcpConfig.Notify()`
+4. The broker's `OnConfigChange()` observer rebuilds the base trust pool and re-registers servers
+
+End-to-end propagation: ~15-30 seconds (controller re-reconciliation + config watcher cycle). This is faster than the previous volume-mount approach since it does not depend on kubelet volume sync (60-120s).
```
maleck13: I missed this previously , this is not accurate. The secret config uses the volume mount and sync approach. So it is the same it depends on the kube sync

docs/design/gateway-ca-cert-bundle/gateway-ca-cert-bundle-design.md
```diff
@@ -0,0 +1,231 @@
...
+- **No new trust** — the gateway CA bundle extends the system root pool, it does not replace it. Servers that work today with publicly-trusted CAs continue to work without changes
+- **Additive-only** — per-server CAs always append to the gateway bundle, never override. An operator cannot accidentally remove trust for a server by setting the gateway bundle
+- **Size limit** — 256 KiB prevents accidental inclusion of large files. This is generous enough for typical CA chains (~20-30 CAs at ~2 KiB each)
+- **Config Secret size** — the gateway CA bundle contributes to the 1 MiB Kubernetes Secret limit, but since it replaces N per-server copies with 1, the net size is always smaller when servers share a CA. For extreme cases, operators can partition servers across multiple gateways
+
+## Future Considerations
+
+### ConfigMap Support
+
+The current design uses a Secret for the CA bundle. A future enhancement could support ConfigMaps, which are more natural for non-sensitive CA certificates and align with how OpenShift distributes the service-serving CA (`openshift-service-ca.crt` ConfigMap). This would require a union-type field or separate `caCertBundleConfigMapRef`.
```
maleck13: This can be removed.  We are using a secret for the config as there can be credentials written to that config so we wouldn't move to a configmap

docs/design/gateway-ca-cert-bundle/gateway-ca-cert-bundle-design.md
```diff
@@ -0,0 +1,231 @@
...
+
+### Certificate Expiry Monitoring
+
+The broker could parse the CA certificates and emit metrics or log warnings when certificates are near expiry. This would help operators detect upcoming CA rotations before they cause outages.
```
maleck13: I would say this is out of scope of the operator. These things are often managed by external services like cert-manager

docs/design/gateway-ca-cert-bundle/tasks/e2e_test_cases.md
```diff
@@ -0,0 +1,42 @@
...
+
+### [Happy,CACertBundle] Per-server CA appends to gateway bundle
+
+- When an MCPGatewayExtension has `caCertBundleRef` set with CA-A, and an MCPServerRegistration has `caCertSecretRef` set with CA-B (a different CA), the broker should trust both CA-A and CA-B for that server. A server whose certificate is signed by CA-B should connect successfully. This verifies additive behavior.
+
+### [CACertBundle] Gateway bundle alone insufficient for server with unique CA
+
+- When an MCPGatewayExtension has `caCertBundleRef` set with CA-A, and an upstream server's certificate is signed by CA-B (not in the gateway bundle), and the MCPServerRegistration does NOT have `caCertSecretRef`, the broker should fail the TLS handshake with a certificate verification error. The MCPServerRegistration should not become Ready.
+
+### [CACertBundle] No gateway bundle — per-server CA still works
```
maleck13: This will already be covered by existing tests as they wont create the gateway CA ref

maleck13 (approved):
ok I think this is good enough to proceed with now

PR #1051, opened 2026-05-27
refactor(router): introduce MCPClientHandle interface for InitForClient

Aman-Cool:
I've been making fixes inside `initializeMCPSeverSession` for a while now and every time I went to write a test for the success path I hit the same wall: `InitForClient` returns a `*client.Client`, and you can't construct one of those in a test without a live MCP server on the other end. So I'd either pre-populate the cache to skip `InitForClient` entirely, or make the mock error out early and call it a day.

At some point I stopped and actually looked at what that means. Everything after the `InitForClient` call ; `GetSessionId()`, the TTL guard, `AddSession`, arming the cleanup timer ; none of it has ever been touched by a unit test. This became particularly uncomfortable after #1037, which added the `AddSession` rollback path and the TTL guard specifically to handle failure cases. We landed that logic with zero unit-test coverage on the branches it was protecting.

The router only ever calls two methods on the returned client: `GetSessionId()` and `Close()`. So this is really just a missing interface. This PR adds one:

```go
type MCPClientHandle interface {
    GetSessionId() string
    Close() error
}
```

`InitForClient` now returns `MCPClientHandle` instead of `*client.Client`. `clients.Initialize` still returns a concrete `*client.Client` under the hood ; it satisfies the interface implicitly ; so nothing changes at runtime. As a side effect, the `mcp-go` client import disappears from `server.go` entirely, which means future `mcp-go` bumps only need touching `internal/clients`.

With the interface in place, three tests that were previously impossible now exist: the happy path where a session gets written to cache and the correct ID comes back, the `InitForClient` error path, and the `AddSession` failure rollback that #1037 added but couldn't test.

## Summary by CodeRabbit

* **Refactor**
  * Updated MCP router/client initialization to use a lightweight client handle, improving decoupling.
  * The initialization now accepts an explicit token and automatically sets the routing and server host headers required for correct downstream requests.

* **Tests**
  * Added/updated unit tests for MCP session initialization, covering the happy path, upstream initialization failures, and session-cache add failures (including proper cleanup/close behavior).


Aman-Cool:
Hey @david-martin, @maleck13 and @jasonmadigan; One thing I didn't mention ; removing the `mcp-go` import from `server.go` was a nice side effect but not the goal, just wanted to make sure that doesn't look like the point of the PR. The interface is purely to unlock the test coverage.
Lmk if anything looks off, happy to rework accordingly!

Aman-Cool:
@coderabbitai, review...

Aman-Cool:

PR #1052, opened 2026-05-28
fix: serialize manager event handling to avoid concurrent manage races

### LINKED ISSUE #916: Bug:  concurrent manage() calls cause unsynchronised access to serverTools

Aryanburnwal05:
This fixes the race condition reported in #916 where manage() could run concurrently from both the ticker loop and notification callbacks leading to unsynchronised access to serverTools.

Instead of protecting the entire manager with a broad mutex, this change routes notification triggered reconciliations through the existing event loop so all manage() executions happen sequentially in a single goroutine.

What changed

1. Added serialized event handling for tool and prompt notifications
2. Updated notification callbacks to enqueue events instead of calling manage() directly
3. Added buffered non-blocking event channels to avoid callback stalls
4. Kept targeted locking around shared state reads/writes
5. Added a regression test covering concurrent notification delivery

Notes

1.  Notifications are intentionally coalesced when an event is already pending
2.  Channels are not closed during shutdown, avoiding callback panics after stop
3.  Existing reconciliation behaviour is preserved

Fixes #916

## Summary by CodeRabbit

* **Tests**
  * Added a concurrency/race test that stress-tests handling of many simultaneous notifications, verifies the system converges to a single managed tool with no duplicates, and confirms invalid tool lists remain empty—improving reliability and preventing race-condition regressions.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Kuadrant/mcp-gateway/pull/1052?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/internal/broker/upstream/manager.go
+++ b/internal/broker/upstream/manager.go
@@ -148,7 +148,9 @@ type MCPManager struct {
 	toolEvents   chan struct{}
 	promptEvents chan struct{}
 	done         chan struct{} // closed when the event loop exits
-	status       ServerValidationStatus
+
+	statusLock sync.RWMutex
+	status     ServerValidationStatus
 }
 
 // DefaultTickerInterval is the default interval for backend health checks
@@ -486,12 +488,17 @@ func (man *MCPManager) shouldFetchPrompts(event eventType) bool {
 }
 
 // GetStatus returns the current status of the MCP Server
-// no locking is done here as it is expected to be called multiple times
+// This is called concurrently by external controllers and tests, so it uses RLock.
 func (man *MCPManager) GetStatus() ServerValidationStatus {
+	man.statusLock.RLock()
+	defer man.statusLock.RUnlock()
 	return man.status
 }
 
 func (man *MCPManager) setStatus(err error, toolCount int, promptCount int, invalidTools []InvalidToolInfo, invalidPrompts []InvalidPromptInfo) {
+	man.statusLock.Lock()
+	defer man.statusLock.Unlock()
+
 	man.status.ID = string(man.mcp.ID())
 	man.status.LastValidated = time.Now()
 	man.status.Name = man.MCPName()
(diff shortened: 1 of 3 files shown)


jasonmadigan:
thanks @Aryanburnwal05, closing -- #943 refactored the manager to a channel-based event loop that serialises all manage() calls through a single select loop, fixing this race by design.

PR #1058, opened 2026-05-28
feat(elicitation): support configurable headers for URL elicitation tokens

### LINKED ISSUE #970: Allow configuring header name and format for URL elicitation tokens
Allow configuring the header name and value format for tokens injected via URL elicitation.

Currently the router injects the elicited token as `Authorization: Bearer <token>`. Some upstream MCP servers expect different header names or formats (e.g. `X-API-Key`, or a bare token without the `Bearer` prefix).

**Proposed API change in `MCPServerRegistration`:**
```yaml
tokenURLElicitation:
  headerName: Authorization          # default
  headerValueFormat: "Bearer {token}" # default
```

`{token}` is replaced with the user-submitted value at injection time.

**Examples:**
```yaml
# GitHub PAT (current default behavior)
tokenURLElicitation:
  headerName: Authorization
  headerValueFormat: "Bearer {token}"

# API key header
tokenURLElicitation:
  headerName: X-API-Key
  headerValueFormat: "{token}"
```

Aryanburnwal05:
This updates URL elicitation token injection to support configurable header names and value formats instead of always injecting :-

```yaml
Authorization: Bearer <token>
```

The default behavior remains unchanged, but MCP server registrations can now customize both the header name and token formatting when forwarding elicited credentials upstream.

Examples :-

```yaml
tokenURLElicitation:
  headerName: Authorization
  headerValueFormat: "Bearer {token}"
```

```yaml
tokenURLElicitation:
  headerName: X-API-Key
  headerValueFormat: "{token}"
```

What changed

1. Added headerName and headerValueFormat to token URL elicitation configuration
2. Propagated the new fields through controller/internal config handling
3. Replaced the hardcoded auth injection path with configurable header injection
4. Centralized default resolution logic for token headers/formats
5. Added safe handling for missing {token} placeholders
6. Updated router and hairpin passthrough flows to use the same logic

Backward compatibility

Existing behavior is preserved when the new fields are omitted :-

```yaml
Authorization: Bearer <token>
```

remains the default behavior for both standard injection and pass through/hairpin flows.

Tests

Extended the existing router test coverage with cases for :-

1.  default Authorization bearer behavior
2.  custom header names
3.  custom token formats
4.  bare token formats
5.  multiple {token} placeholders
6.  missing placeholder handling
7.  nil/empty config handling
8.  passthrough compatibility

Notes

I was not able to regenerate CRD/deepcopy artifacts locally due to a controller-gen / `golang.org/x/tools` issue on my Windows environment (`invalid array length -delta * delta`).

The implementation and tests are complete and passing locally, but generated artifacts may still need to be refreshed from a Linux/WSL environment if CI requires it.

Fixes #970

## Summary by CodeRabbit

* **New Features**
  * Configurable header injection for URL-elicitation tokens with customizable header name and value format (defaults to Authorization / Bearer {token}).
  * Router respects configured header settings and formats injected values accordingly, avoiding overwriting an existing non-empty Authorization header.
* **Tests**
  * Added tests covering default and custom header names, value formatting (including multiple/placeholders), and missing-placeholder behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Kuadrant/mcp-gateway/pull/1058?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/api/v1alpha1/types.go
+++ b/api/v1alpha1/types.go
@@ -150,6 +150,19 @@ type TokenURLElicitationConfig struct {
 	// +optional
 	// +kubebuilder:validation:Pattern=`^https?://`
 	URL string `json:"url,omitempty"`
+
+	// headerName specifies the name of the header to inject the token into.
+	// Defaults to "Authorization".
+	// +optional
+	// +default="Authorization"
+	HeaderName string `json:"headerName,omitempty"`
+
+	// headerValueFormat specifies the format string for the header value.
+	// The token is substituted into the {token} placeholder.
+	// Defaults to "Bearer {token}".
+	// +optional
+	// +default="Bearer {token}"
+	HeaderValueFormat string `json:"headerValueFormat,omitempty"`
 }
 
 // TargetReference identifies an HTTPRoute that points to MCP servers.
--- a/charts/mcp-gateway/crds/mcp.kuadrant.io_mcpserverregistrations.yaml
+++ b/charts/mcp-gateway/crds/mcp.kuadrant.io_mcpserverregistrations.yaml
@@ -208,6 +208,19 @@ spec:
                   When set, the router uses the MCP spec's URLElicitationRequiredError (-32042) flow
                   to collect tokens from capable clients at tool-call time.
                 properties:
+                  headerName:
+                    default: Authorization
+                    description: |-
+                      headerName specifies the name of the header to inject the token into.
+                      Defaults to "Authorization".
+                    type: string
+                  headerValueFormat:
+                    default: Bearer {token}
+                    description: |-
+                      headerValueFormat specifies the format string for the header value.
+                      The token is substituted into the {token} placeholder.
+                      Defaults to "Bearer {token}".
+                    type: string
                   url:
                     description: |-
                       url overrides the default broker token page URL.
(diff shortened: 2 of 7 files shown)


jasonmadigan (changes requested):
changes requested. ci is red (lint, crd sync, e2e).

**blockers**

- `buildRouterConfigKey` is dead code, never called. remove it (this is why lint fails)
- default token format changed from raw to `Bearer {token}`. `WithAuth` injected the raw value before. any deployment using elicitation with raw PATs (e.g. github) will break silently. either default to `{token}` for backwards compat or document the migration
- crd manifests out of sync -- run `make generate-all`

**should fix**

- no validation on `headerName` in the crd. a cluster admin could set it to `:authority` or `mcp-session-id` and overwrite routing-critical headers. add a kubebuilder enum/pattern or runtime blocklist
- the "already has authorization" guard in both code paths only checks for `Authorization` specifically. if `headerName` is `X-API-Key` and the client already sends one, gateway overwrites it silently. generalise the guard to check the configured header name

**nits**

- guard block between `initializeMCPSeverSession` and `resolveUpstreamToken` is repeated. shared helpers help but the guard itself could be extracted
- hairpin path lowercases header name, tool/call path preserves case. http/2 normalises anyway but the inconsistency is odd

david-martin:
Thanks for the contributions! One thing to flag: our [contributing guidelines](https://github.com/Kuadrant/mcp-gateway/blob/main/CONTRIBUTING.md) ask contributors to limit themselves to one open PR at a time. The idea is that getting one change reviewed and merged is more valuable than having several open in parallel.

You currently have three open PRs. Could you pick whichever one you think is most ready for review, and we'll focus on getting that one through first?

Patryk-Stefanski:
Closing due to inactivity

PR #1060, opened 2026-05-29
chore: clean up of main and removal of dead comments

maleck13:
No functionality changed. Just improving consistency, and readability of the main file in particular

## Summary by CodeRabbit

* **Chores**
  * Refactored startup/lifecycle for more reliable startup/shutdown, timeouts, and config hot-reload
  * Expanded HTTP surface: consolidated health/readiness/status endpoints, OPTIONS handling, OAuth-protected resource endpoints, SSE streaming endpoint with session handling and trace-context propagation
  * Added gRPC routing support for Envoy integration

* **Documentation**
  * Clarified session lifecycle: gateway does not explicitly terminate upstream sessions

[Review Change Stack](https://app.coderabbit.ai/change-stack/Kuadrant/mcp-gateway/pull/1060?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/cmd/mcp-broker-router/broker.go
+++ b/cmd/mcp-broker-router/broker.go
@@ -1,0 +1,111 @@
+package main
+
+import (
+	"fmt"
+	"net/http"
+	"os"
+	"time"
+
+	mcpv1alpha1 "github.com/Kuadrant/mcp-gateway/api/v1alpha1"
+	"github.com/Kuadrant/mcp-gateway/internal/broker"
+	"github.com/mark3labs/mcp-go/server"
+	"go.opentelemetry.io/otel"
+	"go.opentelemetry.io/otel/propagation"
+)
+
+func (a *app) createBroker() {
+	invalidToolPolicy := mcpv1alpha1.InvalidToolPolicy(a.brokerCfg.invalidToolPolicy)
+	if invalidToolPolicy != mcpv1alpha1.InvalidToolPolicyFilterOut && invalidToolPolicy != mcpv1alpha1.InvalidToolPolicyRejectServer {
+		panic("--invalid-tool-policy must be FilterOut or RejectServer")
+	}
+
+	managerTickerInterval := time.Duration(a.brokerCfg.managerTickerIntervalSecs) * time.Second
+	if managerTickerInterval <= 0 {
+		panic("flag mcp-check-interval cannot be 0 or less seconds")
+	}
+
+	a.mcpBroker = broker.NewBroker(a.logger.With("component", "broker"),
+		broker.WithEnforceCapabilityFilter(a.brokerCfg.enforceCapabilityFiltering),
+		broker.WithTrustedHeadersPublicKey(os.Getenv("TRUSTED_HEADER_PUBLIC_KEY")),
+		broker.WithManagerTickerInterval(managerTickerInterval),
+		broker.WithInvalidToolPolicy(invalidToolPolicy),
+		broker.WithElicitationEnabled(a.brokerCfg.enableURLElicitation),
+		broker.WithDiscoveryToolsEnabled(a.brokerCfg.discoveryToolsEnabled),
+		broker.WithDiscoveryToolThreshold(a.brokerCfg.discoveryToolThreshold),
+		broker.WithSessionCache(a.sessionCache),
+	)
+	a.tokenHandler = broker.NewTokenHandler(a.sessionCache, a.tokenElicitMap, *a.logger)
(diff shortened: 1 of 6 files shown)


Patryk-Stefanski (commented):
Nice cleanup — the global state elimination and config reload resilience are both solid improvements. Left a few minor suggestions inline.

cmd/mcp-broker-router/main.go
```diff
@@ -44,419 +35,330 @@ var (
...
-	logger.Info("shutting down MCP Broker and MCP Router")
 
+	a.logger.Info("shutting down MCP Broker and MCP Router")
 	shutdownCtx, shutdownRelease := context.WithTimeout(context.Background(), 10*time.Second)
 	defer shutdownRelease()
 
-	if err := otelShutdown(shutdownCtx); err != nil {
-		logger.Error("OpenTelemetry shutdown error", "error", err)
+	if err := a.otelShutdown(shutdownCtx); err != nil {
+		a.logger.Error("OpenTelemetry shutdown error", "error", err)
```
Patryk-Stefanski: `otelShutdown` could be nil if `SetupOTelSDK` returns a nil function on error. In `run()` you call `a.otelShutdown(shutdownCtx)` unconditionally — that would panic. Worth a nil guard:

```go
if a.otelShutdown != nil {
    if err := a.otelShutdown(shutdownCtx); err != nil {
        a.logger.Error("OpenTelemetry shutdown error", "error", err)
    }
}
```

cmd/mcp-broker-router/main.go
```diff
@@ -44,419 +35,330 @@ var (
...
 	}()
 
 	go func() {
-		logger.Info("[grpc] starting MCP Router", "listening", grpcAddr)
-		log.Fatal(routerGRPCServer.Serve(lis))
+		a.logger.Info("[grpc] starting MCP Router", "listening", a.routerCfg.addr)
+		if err := a.grpcServer.Serve(lis); err != nil {
+			a.logger.Error("grpc server error", "error", err)
+			os.Exit(1)
+		}
```
Patryk-Stefanski, line 303: `os.Exit(1)` in these goroutines skips all deferred functions and the graceful shutdown sequence. The old code had the same issue (`log.Fatal`), but since this is a cleanup — consider sending to the `stop` channel instead so the main goroutine handles shutdown properly.

cmd/mcp-broker-router/broker.go
```diff
@@ -0,0 +1,111 @@
...
+		ReadTimeout:  5 * time.Second,
+		WriteTimeout: writeTimeout,
+	}
+
+	streamableHTTPOpts := []server.StreamableHTTPOption{
+		server.WithStreamableHTTPServer(httpSrv),
+	}
+	if a.jwtMgr != nil {
+		a.logger.Debug("jwt session manager configured")
+		streamableHTTPOpts = append(streamableHTTPOpts, server.WithSessionIdManager(a.jwtMgr))
```
Patryk-Stefanski, line 84: nit: this was `logger.Info` before — intentional downgrade to Debug? JWT manager configuration feels like a lifecycle event that warrants Info level per the project convention (`logger.Info` for lifecycle only, `logger.Debug` for per-request).

cmd/mcp-broker-router/main.go
```diff
@@ -44,419 +35,330 @@ var (
...
 	flag.Parse()
 
-	loggerOpts := &slog.HandlerOptions{}
+	if bc.publicHost == "" {
+		panic("--mcp-gateway-public-host cannot be empty. The mcp gateway needs to be informed of what public host to expect requests from so it can ensure routing and session mgmt happens. Set --mcp-gateway-public-host")
+	}
+
+	// copy common config so router has its own snapshot
+	rc.commonConfig = bc.commonConfig
+
```
Patryk-Stefanski, line 161: The struct copy here works since these are read-only after parse, but the two copies will silently diverge if anyone later mutates one. A brief comment noting they are immutable after init would prevent a future subtle bug.

Patryk-Stefanski (commented):
Nice cleanup — the global state elimination and config reload resilience are both solid improvements. Left a few minor suggestions inline.

cmd/mcp-broker-router/main.go
maleck13: looks to me like SetupOTelSDK never returns a nil shutdown? But can be extra defensive I guess in case it changes

Patryk-Stefanski (approved):

PR #1064, opened 2026-05-29
improve the token url elicitatoin demo by enabling authpolicies to be…

maleck13:
This PR updates the demo users can do to use token url elicitation. It allows the host to be specified and generates the resources with that value. 

## Summary by CodeRabbit

* **Configuration & Improvements**
  * Demo now supports configurable OIDC issuer, external URL, and client ID; policies are rendered from templates and applied at deploy time.
  * Demo script accepts flags for environment values and derives gateway URLs from the configured external URL.

* **Behavior & Security**
  * Redirect handling tightened to mitigate open-redirects.
  * Cookie domain and Secure attribute are now configurable/conditional.

* **Chores**
  * Added a utility to generate environment-specific AuthPolicy manifests.


david-martin:
👀

david-martin (approved):
verified the script changes locally.

Issue #949, opened 2026-05-12
Investigate support for Sampling


david-martin:
Split from #257.

Investigate support for MCP [Sampling](https://modelcontextprotocol.io/specification/2025-06-18/client/sampling) capability.

There are currently zero matches for `sampling` or `createMessage` across the codebase.

> [!IMPORTANT]
> **Investigation first — no implementation until this is complete.**
>
> Before any code is written, we need a clear understanding of:
> 1. Whether Sampling fits the mcp-gateway use cases and architecture
> 2. What specific use cases we would be supporting and why
> 3. The scope of changes required and how they affect the gateway overall complexity
> 4. What test coverage is needed, including any test server changes to exercise Sampling end-to-end
>
> We are being intentional about the growing scope of the gateway. New capabilities need to be justified by clear use cases and land with adequate test coverage from the start.

## Subtasks

- [ ] Investigate whether Sampling fits the mcp-gateway use cases
- [ ] Define the use cases to be supported
- [ ] Define the testing approach (test coverage, test server changes)
- [ ] Implement Sampling support (if applicable)

## Notes from community (#257)

> The elicitation implementation is a good model since both sampling and elicitation are server-initiated flows back to the client. For sampling, extending `sseRewriter` to detect `sampling/createMessage`, rewriting the ID via idmap, forwarding to the downstream client, and routing the response back to the correct upstream server would work well. Also declare the `sampling` capability in `clients.go` when the client supports it.

xenonnn4w:
hey @david-martin investigated fit against current codebase.

Sampling lines up well with the existing elicitation flow; both are server-initiated requests over the SSE stream with the client response coming back as a separate HTTP request. Most of the plumbing can be reused:

  - `sseRewriter` in `internal/mcp-router/elicitation.go`, extend to also match `sampling/createMessage`.
  - `HandleElicitationResponse` in `internal/mcp-router/request_handlers.go`, generalize into a shared handler;
  sampling responses are identified by `result.role` + `result.content` instead of `result.action`.
  - Capability detection on `initialize`, session cache flag, and upstream capability advertisement, mirror what's
  done for elicitation.
  - `idmap` , reuse as-is.

 also imo sampling consumes the client's LLM budget, so we probably want a per-`MCPServerRegistration` opt-in flag rather than enabling it for every backend.

 No blockers. Happy to pick this up.

xenonnn4w:
since it pretty big in one go; i'll do it in as phases 
Phase 1
[...]

Issue #788, opened 2026-04-20
Add support for MCP Resources federation


david-martin:
Split from #208.

Add support for federating [MCP Resources](https://modelcontextprotocol.io/specification/2025-03-26/server/resources) through the gateway, similar to how tools and prompts are currently federated.

**Tracking:** [CONNLINK-1232](https://redhat.atlassian.net/browse/CONNLINK-1232)

## Scope

Investigation is complete. This issue is scoped to partial resources support for the **stateful MCP protocol** (2024-11-05 / 2025-03-26 / 2025-06-18), targeting **MCP Apps (SEP-1865) UI rendering** as the motivating use case.

### In scope

1. **`resources/list` + `resources/read` federation** — broker aggregates resource lists from all connected upstreams at request time, rewrites `ui://` URIs with a per-server prefix to avoid collisions across upstreams, and merges results. The router dispatches `resources/read` to the correct upstream by parsing the prefix from the URI.

2. **`_meta.ui.resourceUri` rewriting in `tools/call` responses** — when a tool result contains a `_meta.ui.resourceUri` field, the router rewrites the URI to its prefixed form before it reaches the client. Without this, the URI in the tool response and the URI in `resources/list` diverge and the client lookup fails silently.

3. **URI scheme: `ui://` only** — prefix is injected into the authority segment of the URI (e.g. `ui://template.html` → `ui://insights_template.html`). General schemes (`file://`, `git://`, `https://`) are a follow-up tracked in #1238.

### Out of scope

- Full resource federation: subscriptions, URI templates (`resources/templates/list`), non-`ui://` schemes — tracked in #1238
- Stateless (Streamable HTTP) protocol resources support
- `cacheScope` / `ttlMs` cache-aware proxying (upcoming spec SEP-2549)
- VirtualServer resource filtering

## Subtasks

- [x] Investigate whether Resources federation fits the mcp-gateway use cases
- [x] Define the use cases to be supported
- [x] Determine scope
- [ ] Write a design doc and get maintainer approval before implementing
- [ ] Implement Resources federation
- [ ] Add test coverage (unit + e2e)

wolf-06:
Hi @david-martin 
is it open for contribution ?
I would like to work on this.

david-martin:
Hi @wolf-06, as mentioned on Slack, this is a significant feature piece, so not ideal as a starting point.

I'd recommend picking up something labelled "good first issue" first to get familiar with the codebase (I've recently labelled more issues).
[...]

PR #1068, opened 2026-06-02
broker: federate MCP resources alongside tools and prompts

Aman-Cool:
Resources are the third MCP primitive and we were federating two of the three ; this fills the gap.

I used prompts as the template throughout. The pattern is identical: `SupportsResources` / `SupportsResourcesListChanged` on `MCPServer`, a `ResourcesAdderDeleter` interface, resource fields on `MCPManager` with the same diff-on-tick logic, a `FilterResources` hook wired into `AddAfterListResources`, and `Resources []string` on `VirtualServer` for scoping. The one real difference from prompts is that URIs are globally unique so there's no prefix rewriting anywhere ; `GetServerInfoByResource` is a straight map lookup, no `strings.CutPrefix`.

For `resources/read` routing I followed the same `ext_proc` path as `prompts/get` ; added `HandleResourceRead` in `request_handlers.go` so traffic still flows through Envoy rather than short-circuiting through the broker directly. Felt important to keep that consistent with the "all MCP traffic flows through Envoy" principle in the design doc.

The constructor signature for `NewUpstreamMCPManager` gains one parameter (`resourcesServer ResourcesAdderDeleter`), which touched a lot of call sites ; all mechanical `nil` additions in tests. Everything else is additive.

Tests follow the existing patterns exactly: capability cases in `mcp_test.go`, diff/notification/nil-server cases in `manager_test.go`, JWT and virtual-server filtering in `filtered_resources_handler_test.go`, and lookup cases in `broker_test.go`. All green, builds clean.

## Summary by CodeRabbit

* **New Features**
  * Gateway now aggregates upstream resources, supports resources/list and resources/read, routes reads to the correct backend, and lets VirtualServer configs declare resources; resources can be JWT-filtered by capability claims and scoped by VirtualServer allowlists.

* **Tests**
  * Added unit and e2e coverage for listing, JWT filtering, VirtualServer filtering, routing, read error handling, and resource sync behavior.

* **Documentation**
  * Added e2e test cases for resources federation.

### DIFF
--- a/internal/broker/upstream/manager.go
+++ b/internal/broker/upstream/manager.go
@@ -76,6 +76,7 @@ type ServerValidationStatus struct {
 	Ready             bool                `json:"ready"`
 	TotalTools        int                 `json:"totalTools"`
 	TotalPrompts      int                 `json:"totalPrompts"`
+	TotalResources    int                 `json:"totalResources"`
 	InvalidTools      int                 `json:"invalidTools"`
 	InvalidToolList   []InvalidToolInfo   `json:"invalidToolList,omitempty"`
 	InvalidPrompts    int                 `json:"invalidPrompts"`
@@ -349,12 +350,13 @@ func (man *MCPManager) manage(ctx context.Context, event eventType) {
 		man.removeAllTools()
 		man.removeAllPrompts()
 		_ = man.mcp.Disconnect()
-		man.setStatus(fmt.Errorf("server is disabled"), 0, 0, nil, nil)
+		man.setStatus(fmt.Errorf("server is disabled"), 0, 0, 0, nil, nil)
 		return
 	}
 
 	numberOfTools := len(man.tools)
 	numberOfPrompts := len(man.prompts)
+	numberOfResources := len(man.resources)
 	// during connect the client will validate the protocol. So we don't have a separate validate requirement currently. If a client already exists it will be re-used.
 	man.logger.DebugContext(ctx, "attempting to connect", "upstream mcp server", man.mcp.ID())
 	if err := man.mcp.Connect(ctx, man.registerCallbacks()); err != nil {
@@ -366,7 +368,7 @@ func (man *MCPManager) manage(ctx context.Context, event eventType) {
 		man.removeAllPrompts()
 		// we call disconnect here as we may have connected but failed to initialize
 		_ = man.mcp.Disconnect()
-		man.setStatus(err, numberOfTools, numberOfPrompts, nil, nil)
+		man.setStatus(err, numberOfTools, numberOfPrompts, numberOfResources, nil, nil)
 		return
 	}
 	// there may be an active client so we also ping
@@ -379,7 +381,7 @@ func (man *MCPManager) manage(ctx context.Context, event eventType) {
 		man.removeAllTools()
 		man.removeAllPrompts()
 		_ = man.mcp.Disconnect()
-		man.setStatus(err, numberOfTools, numberOfPrompts, nil, nil)
+		man.setStatus(err, numberOfTools, numberOfPrompts, numberOfResources, nil, nil)
 		return
 	}
 
@@ -527,6 +529,7 @@ func (man *MCPManager) manage(ctx context.Context, event eventType) {
 			} else {
 				man.toolsLock.Lock()
 				man.resources = fetchedResources
+				numberOfResources = len(fetchedResources)
 				man.resourcesMap = make(map[string]*mcp.Resource, len(fetchedResources))
 				man.servedResourcesMap = make(map[string]*mcp.Resource, len(fetchedResources))
 				for i := range fetchedResources {
@@ -551,7 +554,7 @@ func (man *MCPManager) manage(ctx context.Context, event eventType) {
 	}
 
 	jointErr := errors.Join(toolErr, promptErr, resourceErr)
-	man.setStatus(jointErr, numberOfTools, numberOfPrompts, invalidTools, invalidPrompts)
+	man.setStatus(jointErr, numberOfTools, numberOfPrompts, numberOfResources, invalidTools, invalidPrompts)
 	if jointErr != nil {
 		man.applyBackoff()
 	} else {
@@ -596,7 +599,7 @@ func (man *MCPManager) GetStatus() ServerValidationStatus {
 	return man.status
 }
 
-func (man *MCPManager) setStatus(err error, toolCount int, promptCount int, invalidTools []InvalidToolInfo, invalidPrompts []InvalidPromptInfo) {
+func (man *MCPManager) setStatus(err error, toolCount int, promptCount int, resourceCount int, invalidTools []InvalidToolInfo, invalidPrompts []InvalidPromptInfo) {
 	man.status.ID = string(man.mcp.ID())
 	man.status.LastValidated = time.Now()
 	man.status.Name = man.MCPName()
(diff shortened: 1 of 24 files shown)


Aman-Cool:
Hey @david-martin @jasonmadigan and @maleck13,, wanted to drop a quick note on why I picked this up.

While going through the codebase I noticed tools and prompts are both federated but resources, the third MCP primitive ; weren't wired up at all. Felt like an obvious gap, especially since the Everything Server already serves resources and there was clearly intent to support them (the SDK already has `AddResources`, `DeleteResources`, `ListResources` on the server side). The groundwork was basically there, just needed someone to connect the dots.

I used the prompts implementation as a strict template ; same diff-on-tick pattern in the manager, same two-pass filtering (JWT capabilities + `VirtualServer` scoping), same `ext_proc` routing path for `resources/read` so traffic still goes through Envoy. The only meaningful difference from prompts is that resource URIs are globally unique, so there's no prefix rewriting needed anywhere in the routing path.

Happy to adjust anything.., especially around the `NewUpstreamMCPManager` signature change since that touched a lot of call sites. If you'd prefer a different approach there (like an options struct or a separate setter) I can rework it.

Aman-Cool:
@david-martin @jasonmadigan and @maleck13, took a look at this.., the failing test is [`Happy] should aggregate prompts from multiple servers with different prefixes`, which this PR never touches. The actual cause is the `mcp-gateway` pod crashing mid-run (`0/1 Error` at ~7 minutes in), which drops the SSE connection and causes everything waiting on a response to time out. Not a regression from this change. A re-run should clear it.

Bhuvanesh66:
Hi @Aman-Cool,
Great contribution — filling the resources gap was the right call and 
using prompts as a strict template makes the diff easy to follow. The 
diff-on-tick pattern, two-pass filtering, and keeping resources/read 
on the ext_proc path are all consistent with the existing design. A few 
things worth addressing before merge:

**Needs fixing:**

- The stub handler in `resourceToServerResource()` returns empty content 
  with no error. Works today because ext_proc routing bypasses it, but 
  any direct call silently returns nothing with HTTP 200. Should return 
  an explicit error instead. (see inline comment on manager.go)

- `setStatus()` never receives resource counts so the 
  MCPServerRegistration CR status will always show 0 resources. Tools 
  and prompts both pass their counts in — resources should follow the 
  same pattern. (see inline comment on manager.go)

**Minor:**

- `"kuadrant/id"` used as a raw string literal in 
  `filtered_resources_handler.go` instead of the `gatewayServerID` 
  constant defined in `manager.go`.

- `createResourceTestManager` and `createResourceFilterTestManager` 
  in two separate test files are nearly identical and will drift over 
  time — worth consolidating into a shared helper.

- The E2E test only asserts HTTP 200 and no JSON-RPC error code. Since 
  the Everything Server serves known content, the test should also 
  assert that actual resource data is returned, not just that the 
  request didn't crash.

The failing E2E (`[Happy] should aggregate prompts from multiple 
servers`) looks like an infra flake as you described — a clean re-run 
before merge would help reviewers confirm that with confidence.

Overall this is solid work, the two items under "needs fixing" are 
the only things blocking from my side.

internal/broker/upstream/manager.go
```diff
@@ -831,3 +960,100 @@ func (man *MCPManager) SetPromptsForTesting(prompts []mcp.Prompt) {
 		man.servedPromptsMap[prefixedName(man.mcp.GetPrefix(), prompts[i].Name)] = &prompts[i]
 	}
 }
+
+func (man *MCPManager) resourceToServerResource(newResource mcp.Resource) server.ServerResource {
```
Bhuvanesh66, line 990: The stub handler registered for every resource silently returns empty 
content with no error:

    Handler: func(_ context.Context, _ mcp.ReadResourceRequest) ([]mcp.ResourceContents, error) {
        return []mcp.ResourceContents{}, nil
    }

This works today only because resources/read is routed through ext_proc 
and never hits this handler directly. But if any code path calls it 
directly — a test, a future refactor, a non-ext_proc fallback — it will 
return empty content with HTTP 200, silent data loss with no indication 
anything went wrong.

Suggested fix, make the stub fail loudly:

    Handler: func(_ context.Context, _ mcp.ReadResourceRequest) ([]mcp.ResourceContents, error) {
        return nil, fmt.Errorf("resources/read must be routed via ext_proc, not called directly")
    }

internal/broker/upstream/manager.go
```diff
@@ -466,7 +511,46 @@ func (man *MCPManager) manage(ctx context.Context, event eventType) {
...
+				}
+				if len(toAddResources) > 0 {
+					man.resourcesServer.AddResources(toAddResources...)
+				}
+			}
+		}
+	}
+
+	jointErr := errors.Join(toolErr, promptErr, resourceErr)
 	man.setStatus(jointErr, numberOfTools, numberOfPrompts, invalidTools, invalidPrompts)
```
Bhuvanesh66: Resource counts are never passed into setStatus():

    man.setStatus(jointErr, numberOfTools, numberOfPrompts, invalidTools, invalidPrompts)

The MCPServerRegistration CR status will always show 0 resources 
regardless of how many are actually being federated. This makes it 
impossible to observe whether resource federation is working or whether 
resources are being silently filtered out.

Tools and prompts both pass their counts in — resources should follow 
the same pattern.

Bhuvanesh66:
I've traced through the code and this is not a flake — it's a 
deterministic failure introduced by this PR, and it reproduces 
every run.

The failing test registers two MCPServerRegistrations both pointing 
to sharedMCPTestServer1 with different prefixes (s1a_ and s1b_). 
Because they point to the same backend, they expose identical 
resource URIs.

When the second manager runs findResourceConflicts, it calls 
checkConflicts against the gateway's resource registry. The first 
manager already registered those URIs under its own UpstreamMCPID. 
The second manager has a different UpstreamMCPID (different prefix 
+ hostname in the ID), so the ownership check fails:

    if ownerID, ok := existing[key]; ok && ownerID != string(man.mcp.ID()) {
        conflicting = append(conflicting, key)
    }

This sets resourceErr, which flows into jointErr, which causes 
setStatus to set Ready=false. The MCPServerRegistration never 
becomes ready and the test times out after 300s.

The same test with the same two registrations passed before this 
PR because resources were never federated at all — there was no 
conflict check to fail.

The root cause is that resource URI uniqueness is global in the 
MCP spec, but two different gateway registrations pointing at the 
same backend legitimately serve the same URIs. The tools path 
avoids this because prefixes make tool names unique across 
registrations — resources have no prefix rewriting, so two 
registrations of the same server will always conflict on resources.

The fix needs to decide: should two registrations of the same 
backend be allowed to serve the same resource URIs (ownership 
check by backend URL rather than full UpstreamMCPID), or should 
the resource list be deduplicated rather than treated as a 
conflict?

Aman-Cool:
Hey @Bhuvanesh66, just wanted to say.., your review on this was genuinely impressive!
The depth of analysis, especially tracing the same-backend URI conflict through the UpstreamMCPID format, prefix rewriting, and all the way to the test timeout; that's not something you catch with a quick skim. 
The fact that you've only got one merged PR makes it even more mind blowing to me, it shows you clearly read and understand this codebase at a level that most people with far more contributions don't. 
Really appreciate the thorough review, it made the PR significantly better.🙂

Aman-Cool:
@maleck13, would love a review from your side when you get a chance.

david-martin:
Hey @Aman-Cool, thanks for the work on this and #1076. Really appreciate the initiative, especially using the prompts pattern as a template.

We're going to close both PRs for now though. As we've done with other capabilities (see #949 for Sampling as an example), we want to complete an investigation into how Resources federation fits the gateway before landing implementation. There's more to work through than just the code: use cases, scope, impact on gateway complexity, and test coverage all need to be well understood first.

We've updated #788 to reflect this. Once there's a clear direction, contributions would be very welcome. Feel free to chime in on #788 with any thoughts on the design side too!

PR #1072, opened 2026-06-02
update to security arch doc

maleck13:
add security properties for each section to allow better analysis via agents

## Summary by CodeRabbit

* **Documentation**
  * Enhanced security architecture docs: clearer component boundaries and responsibilities, explicit security properties and non-goals, clarified authentication/authorization delegation and token handling expectations, strengthened session isolation guarantees (signing, session ID rewriting/expiry behavior), and documented limits on request/response processing to mitigate prompt-injection/context-pollution.

### DIFF
--- a/docs/design/security-architecture.md
+++ b/docs/design/security-architecture.md
@@ -41,11 +41,13 @@ MCP Client
 └──────────┬──────────────────────┬────────────┘
            │                      │
            ▼                      ▼
-   ┌──────────────┐     ┌──────────────────┐
-   │  MCP Broker  │     │  Upstream MCP    │
-   │  (internal/  │     │  Servers         │
-   │  broker/)    │     │  (tools/call)    │
-   └──────────────┘     └──────────────────┘
+   ┌──────────────────┐     ┌──────────────────┐
+   │  MCP Broker      │     │  Upstream MCP    │
+   │  (internal/      │     │  Servers         │
+   │  broker/)        │     │  (tools/call)    │
+   │  tools/list,     │     └──────────────────┘
+   │  initialize      │
+   └──────────────────┘
 ```
 
 ### Data crossing each boundary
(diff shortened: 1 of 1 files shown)


docs/design/security-architecture.md
```diff
@@ -42,8 +43,8 @@ MCP Client
            ▼                      ▼
    ┌──────────────┐     ┌──────────────────┐
    │  MCP Broker  │     │  Upstream MCP    │
-   │  (tools/list,│     │  Servers         │
-   │  initialize) │     │  (tools/call)    │
+   │  (internal/  │     │  Servers         │
```
Patryk-Stefanski: Nit: The old text (tools/list, initialize) described what the broker does. The new text gives the source path but drops the functional context, this is a biot confusing when looking and trying to compare what mcp broken and upstream mcp do

docs/design/security-architecture.md
```diff
@@ -83,6 +84,14 @@ The gateway is NOT responsible for:
 - **Content filtering**: responses from upstream servers are streamed back as-is
 - **Upstream server security**: backend MCP servers must implement their own input validation or leverage prompt guard tooling integrated independently at the gateway
 
+### Security properties
+
+- The router (`internal/mcp-router/`) never accesses `credentialRef` values; it only reads routing headers and the JSON-RPC method/tool name from the request body
+- The broker (`internal/broker/`) never writes `credentialRef` values to client-facing responses, logs, or headers forwarded through Envoy
+- The controller (`internal/controller/`) scopes generated Secrets to the operator namespace and sets owner references for garbage collection. The controller's ClusterRole grants cluster-wide secrets access, but the informer cache is filtered by label (`mcp.kuadrant.io/aggregated=true`) to limit the working set
```
Patryk-Stefanski: 1. The informer cache IS filtered by label — but it's `mcp.kuadrant.io/secret=true` ([cmd/main.go#L91-L92](https://github.com/Kuadrant/mcp-gateway/blob/d6754b16/cmd/main.go#L91-L92)), which is `ManagedSecretLabel`/`ManagedSecretValue` defined at [mcpserverregistration_controller.go#L43-L45](https://github.com/Kuadrant/mcp-gateway/blob/d6754b16/internal/controller/mcpserverregistration_controller.go#L43-L45). This label is for credential secrets and the aggregated config secret.

2. `mcp.kuadrant.io/aggregated=true` is a different label — it's applied to the aggregated config secret at [config_writer.go#L114](https://github.com/Kuadrant/mcp-gateway/blob/d6754b16/internal/config/config_writer.go#L114) and used in [config_writer.go#L200](https://github.com/Kuadrant/mcp-gateway/blob/d6754b16/internal/config/config_writer.go#L200) for cleanup queries (`client.MatchingLabels`). It's not what filters the informer cache.

3. Both labels are on the config secret ([config_writer.go#L113-L116](https://github.com/Kuadrant/mcp-gateway/blob/d6754b16/internal/config/config_writer.go#L113-L116) show both `mcp.kuadrant.io/aggregated: "true"` and `mcp.kuadrant.io/secret: "true"`), which is why the config secret passes the informer filter — but the filtering label is `mcp.kuadrant.io/secret`, not `mcp.kuadrant.io/aggregated`.

docs/design/security-architecture.md
maleck13: yeah fair enough will update

maleck13:
fixed those two 👍

Patryk-Stefanski (approved):

PR #1076, opened 2026-06-03
controller: wire resources field through MCPVirtualServer CRD

Aman-Cool:
While working on resource federation I noticed that `VirtualServer.Resources` was already sitting in the config types (we added it as part of the federation PR), but nothing was actually populating it. The `MCPVirtualServerSpec` had no `resources` field, so even if someone wrote a `VirtualServer` manifest with a resources list, the controller would silently drop it and the filtering code we wired up would never see anything.

Related PR: #1068 

This fills that gap. Added `resources` to the CRD spec alongside `tools` and `prompts`, updated the deepcopy, threaded it through `generateVirtualServerConfig` in the controller (one line, mirrors exactly what `tools` and `prompts` do), updated all three CRD YAML copies, added `WithResources` to the e2e builder, and documented it in the API reference.

Also added the first unit test for `generateVirtualServerConfig` since there were none; uses a fake client, no integration tag needed, covers the full `tools+prompts+resources` mapping plus the omitted-means-nil case.

## Summary by CodeRabbit

* **New Features**
  * Added a new optional "resources" field to MCPVirtualServer to explicitly list resource URIs to expose (omitting it exposes all).

* **Documentation**
  * Updated reference docs to describe the new "resources" field and its behavior.

* **Tests**
  * Added unit and e2e test coverage and builder support for the new "resources" field.


Aman-Cool:
Hey @david-martin and @maleck13, this is a follow-up to #1068. While working on resource federation I realized `VirtualServer.Resources `was already in the config types but never reachable from the CRD; the spec field was missing so the controller just dropped it silently. This wires it up end to end.

Issue #1020, opened 2026-05-20
Split operator and operand into separate repositories


david-martin:
## Summary

Split the mcp-gateway monorepo into separate repositories for the operator and the operand (broker, router, and related components) before GA.

Jira: https://redhat.atlassian.net/browse/CONNLINK-1025

## Why

A single repo couples the operator and operand release cycles. Separate repos enable independent versioning, build pipelines, and release cadences — needed to unblock release automation.

## What needs to happen

- Define the repo boundary (what stays, what moves)
- Migrate operator code, CI pipelines, and OLM packaging to a new repo
- Update cross-repo references (image builds, e2e tests, docs)

Aditya7880900936:
Hi @david-martin ,
I started exploring the current monorepo structure to better understand the operator ↔ operand coupling before the split.

From the initial exploration, it seems the challenging part may not just be moving code, but preserving the existing integration flow across:

* image build and publishing pipelines
* deployment manifests / OLM packaging
* e2e orchestration
* shared CRDs and APIs
* docs/examples
* release and version compatibility between operator and operand components

I’m also trying to trace how tightly the reconciliation flow currently depends on operand-specific implementation details versus stable interfaces/contracts, since that will likely influence how cleanly the repositories can evolve independently after the split.

One thing I’m curious about exploring further is whether the separation boundary should primarily follow:

1. control-plane vs data-plane ownership, or
2. release/build independence boundaries.

I’d also be interested in helping with some of the preparatory work for the split while exploring the current structure further — especially around identifying shared packages/interfaces, CI coupling points, dependency boundaries, and potential incremental refactors that could make the eventual migration smoother.

vibhor-5:
>  Separate repos enable independent versioning, build pipelines, and release cadences — needed to unblock release automation.
  
Don't you think it can create issues between the two parts of the tool to maintain a sync between them ?

Bhuvanesh66:
I've been tracing the monorepo to understand what a split would actually involve for #1020 — sharing my initial read in case it helps the discussion.

The runtime boundary already exists (control-plane vs data-plane: `cmd/main.go` / `mcp-controller` vs `cmd/mcp-broker-router/main.go` / `mcp-gateway`, with `MCPGatewayExtension` deploying the operand). The harder part looks like extracting stable contracts and decoupling CI, not just moving directories.

A few coupling points stood out:
[...]

PR #1078, opened 2026-06-04
docs: add design doc for operator/operand repo split

Patryk-Stefanski:
## Summary

- Design doc for splitting `mcp-gateway` into two repos: `mcp-gateway` (operand) and `mcp-gateway-operator` (control-plane)
- Covers repository boundary, runtime contracts (config secret, status endpoint, deployment), CI/CD split, OLM/release pipeline, and phased migration plan
- Full coupling analysis: 10 compile-time coupling points (5 controller→operand, 5 operand→CRD) to resolve in Phase 1

Relates-to: #1020

## Test plan

- [ ] Review design doc for technical accuracy
- [ ] Verify coupling analysis matches current codebase (`grep -rn 'api/v1alpha1' internal/broker/ cmd/mcp-broker-router/`)
- [ ] Verify config secret schema matches `internal/config/types.go`

## Summary by CodeRabbit

* **Documentation**
  * Added a comprehensive design proposing splitting the monorepo into separate operand and operator repositories. Defines clear repo boundaries, runtime and config contracts between operator and operand, deployment and health-check expectations, CI/CD and release-flow guidance, packaging/artifact ownership, cross-repo compatibility/testing workflow, and a three-phase migration plan with compatibility and migration guidance.

### DIFF
--- a/docs/design/repo-split/repo-split-design.md
+++ b/docs/design/repo-split/repo-split-design.md
@@ -69,12 +69,15 @@ mcp-gateway-operator/
 │   └── config/          # config secret writer + serialization types
 ├── charts/mcp-gateway/  # Helm chart (CRDs, RBAC, controller, MCPGatewayExtension)
 ├── bundle/              # OLM manifests
+├── catalog/             # FBC catalog
 ├── config/
 │   ├── crd/             # generated CRD YAMLs
 │   ├── mcp-system/      # K8s deployment manifests
-│   ├── test-servers/    # K8s manifests for test servers
+│   ├── rbac/            # RBAC manifests
+│   ├── test-servers/    # K8s manifest for a minimal test server
 │   └── samples/         # example manifests
-├── tests/e2e/           # full e2e tests (pulls released operand image)
+├── tests/
+│   └── e2e/             # controller-focused e2e (envtest + minimal server)
 ├── Dockerfile
 └── go.mod               # github.com/Kuadrant/mcp-gateway-operator
 ```
@@ -121,7 +124,7 @@ Both repos define their own Go structs for this schema. The operator serializes 
 - The operator must not remove a field without a deprecation period spanning at least one operand release, preventing older operands from receiving a config missing a field they expect.
 - New required fields must have sensible defaults so older operators that don't set them still produce valid config.
 - Breaking changes require a coordinated release with both repos tagged.
-- The config includes a top-level `configVersion: 1` field. The operand checks this on load — known versions are processed normally; unknown versions log a warning with the expected and actual values. This enables detection of operator/operand version mismatches without hard failures.
+- The config includes a top-level `configVersion: 1` field. This is a schema version, not tied to the operator or operand release version — it is bumped only when the config secret format has a breaking structural change (field renames, type changes, removed fields). Additive changes (new optional fields) do not bump the version. The operand checks this on load — known versions are processed normally; unknown versions log a warning with the expected and actual values.
 
 #### Status Endpoint (`GET /status`)
 
@@ -155,12 +158,6 @@ The operator controls the operand's deployment specification:
 | Ports | Fixed | 8080 (HTTP), 50051 (gRPC), 8181 (config) |
 | Health | HTTP | `GET /readyz` on port 8080 |
 | Config mount | Volume | `/config/config.yaml` |
-
-### MCP Spec Considerations
-
-A future MCP specification revision will introduce a stateless protocol model (no `initialize`/`Mcp-Session-Id`). The operand will need to support both old and new spec versions simultaneously during the transition period. This dual-spec support is entirely operand-internal — the operator-operand contracts above are spec-version-agnostic.
-
-The operator does not need to know which MCP spec version clients or backends use. If a `protocolVersion` CRD field is needed in the future, it would be a post-split change managed in the operator repo.
 
 ### CI/CD Pipeline Split
 
@@ -202,8 +199,8 @@ Most feature work lands in the operand repo (broker/router), not the operator. T
 
 A workflow in the operator repo that deploys both components together and runs full integration tests:
 
-- **Nightly:** Latest released operator + latest released operand. Catches drift.
-- **Manual:** Accepts image overrides for both components (e.g., operand PR image + operator PR image). Used for coordinated changes that touch both repos.
+- **Nightly:** Builds `:nightly` tagged images from main of both repos and tests them together. Catches drift between the two codebases.
+- **Manual:** Accepts image overrides for both components. Used for custom combinations — e.g., running `rc1` of the operand with `v1.5` of the operator when there were no operator changes.
 - **Pre-release gate:** Must pass before tagging either component.
 
 #### Release Flow
@@ -284,7 +281,7 @@ Add a CI check that builds each binary (`cmd/main.go` and `cmd/mcp-broker-router
 | `charts/`, `bundle/`, `catalog/` | mcp-gateway-operator |
 | `bundle.Dockerfile`, `build/olm.mk`, `utils/generate-catalog.sh` | mcp-gateway-operator |
 | `config/crd/`, `config/mcp-system/` | mcp-gateway-operator |
-| `config/test-servers/` | mcp-gateway-operator |
+| `config/test-servers/` | mcp-gateway-operator (minimal — one server for MCPServerRegistration testing) |
 | `tests/e2e/` | mcp-gateway-operator |
 | `tests/servers/` (source code) | stays in mcp-gateway |
 | `Dockerfile` (operand) | stays in mcp-gateway |
@@ -302,7 +299,7 @@ Add a CI check that builds each binary (`cmd/main.go` and `cmd/mcp-broker-router
 
 The current `Makefile` has targets for both components (`build`, `docker-build`, `docker-build-controller`, `generate`, `manifests`, etc.). Post-split:
 
-- **Operator repo** gets a new `Makefile` with controller-specific targets: `generate`, `manifests`, `docker-build`, `test-unit`, `test-controller-integration`, `lint`, `helm-*`, and CRD generation (`controller-gen`).
+- **Operator repo** gets a new `Makefile` with controller-specific targets: `generate`, `manifests`, `docker-build`, `test-unit`, `test-controller-integration`, `lint`, `helm-*`, and CRD generation (`controller-gen`). Includes `make deploy-operator OPERAND_IMAGE=<image>:<tag>` to deploy the operator with a specific operand version — used for local development and cross-repo testing.
 - **Operand repo** retains the existing `Makefile` but removes controller-specific targets (`docker-build-controller`, `generate`, `manifests`, CRD generation). Keeps `build`, `docker-build`, `test-unit`, `lint`, and performance test targets.
 
 ### CLAUDE.md
@@ -318,7 +315,7 @@ User-facing guides are published at docs.kuadrant.io from the `Kuadrant/docs.kua
 
 ### Test Server Images
 
-Test server source code (`tests/servers/`) stays in the operand repo and produces test server container images. The operator repo's e2e tests reference these images. The operator repo contains the Kubernetes manifests (`config/test-servers/`) that deploy them but does not build them — it pulls pre-built images from the operand repo's CI.
+Test server source code (`tests/servers/`) stays in the operand repo and produces test server container images. The operator repo only needs a single minimal test server and its K8s manifest to test MCPServerRegistration reconciliation — it does not need the full suite of test servers. The operand repo's e2e tests use the full set of test servers directly.
 
 ## Current Coupling Analysis
 
--- a/docs/design/repo-split/repo-split-design.md
+++ b/docs/design/repo-split/repo-split-design.md
@@ -212,9 +212,9 @@ Independent release cadences. The operator pins to a known-good operand version.
 
 ### OLM and Release Pipeline
 
-CI is GitHub Actions-based — no Tekton pipelines to split. The OLM packaging artifacts (`bundle/`, `catalog/`, `bundle.Dockerfile`, `build/olm.mk`, `utils/generate-catalog.sh`) and AppStudio ReleasePlan resources all move to the operator repo. The operand repo has no OLM footprint.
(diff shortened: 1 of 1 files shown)


docs/design/repo-split/repo-split-design.md
```diff
@@ -0,0 +1,349 @@
...
+A future MCP specification revision will introduce a stateless protocol model (no `initialize`/`Mcp-Session-Id`). The operand will need to support both old and new spec versions simultaneously during the transition period. This dual-spec support is entirely operand-internal — the operator-operand contracts above are spec-version-agnostic.
+
+The operator does not need to know which MCP spec version clients or backends use. If a `protocolVersion` CRD field is needed in the future, it would be a post-split change managed in the operator repo.
+
+### CI/CD Pipeline Split
+
+#### `mcp-gateway` (operand) CI
+
+| Workflow | Trigger | Purpose |
+|---|---|---|
```
maleck13, line 177: I think we want e2e on this repo as it is the core repo that offers the user facing APIs?  The e2e suite needs some thought with this split. Need to think about that one a little more. But we def want some form of e2e suite as most features will land here in PRs not in the operator

docs/design/repo-split/repo-split-design.md
```diff
@@ -0,0 +1,349 @@
...
+| `images.yaml` | Push to main, tags | Build + push `ghcr.io/kuadrant/mcp-gateway` |
+| `test-images.yaml` | Push to main | Build + push test server images |
+| `conformance.yaml` | Push to main | MCP protocol conformance tests |
+| `spelling.yaml` | PR | Spell check |
+
+#### `mcp-gateway-operator` CI
+
+| Workflow | Trigger | Purpose |
+|---|---|---|
+| `tests.yaml` | PR, push to main | Unit tests for controllers |
```
maleck13, line 190: we would also execute the controller integration tests here.

maleck13: NVM I see it below

docs/design/repo-split/repo-split-design.md
```diff
@@ -0,0 +1,349 @@
...
+| `e2e-auth.yaml` | Push to main, manual | Auth-specific e2e |
+| `code-style.yaml` | PR | Lint, vet, formatting |
+| `images.yaml` | Push to main, tags | Build + push `ghcr.io/kuadrant/mcp-controller` |
+| `helm-release.yaml` | Tags | Publish Helm chart |
+| `helm-install-test.yaml` | PR | Helm install smoke test |
+| `verify-crd-sync.yaml` | PR | Ensure generated CRDs match chart CRDs |
+
+#### Cross-Repo E2E
+
+The operator repo's e2e tests pull the operand image by tag:
```
maleck13: The issue is we want a bunch of e2e tests to execute on changes to the gateway not just on changes to the operator

docs/design/repo-split/repo-split-design.md
```diff
@@ -0,0 +1,349 @@
...
+Operator release:
+  1. Tag mcp-gateway-operator → CI builds image + publishes Helm chart
+  2. Helm chart references the operand image tag from the bump above
+```
+
+Independent release cadences. The operator pins to a known-good operand version. Compatibility is documented in a version matrix in the operator repo README.
+
+### OLM and Release Pipeline
+
+CI is GitHub Actions-based — no Tekton pipelines to split. The OLM packaging artifacts (`bundle/`, `catalog/`, `bundle.Dockerfile`, `build/olm.mk`, `utils/generate-catalog.sh`) and AppStudio ReleasePlan resources all move to the operator repo. The operand repo has no OLM footprint.
```
maleck13: AppStudio?

Patryk-Stefanski: Old name for konflux, will update.

docs/design/repo-split/repo-split-design.md
Patryk-Stefanski: What if the operand repo's CI builds a temporary image from the PR, deploys the latest released operator (via Helm), and runs the operator repo's e2e suite against that image. 

The e2e test  definitions stay in the operator repo as the single source of truth — the operand CI pulls them at test time (git checkout or similar). The operator repo continues running the same suite on its own PRs against a released operand image.

This gives full e2e coverage on operand PRs without duplicating the test suite or adding cross-repo dispatch latency.

maleck13: Well yes that works in many cases. But it doesn't work when the operator and the operand change together. 
This is not that un-common, by nature they are tightly coupled components, this is why the single monorepo works so well. 

Without the mono repo you end up In a chicken and egg situation where a new flag is  added or env var you end up with a chicken and egg situation. (operator needs new image of the operand to properly test , operand needs a new image of the operator to properly test). 

To do this properly we probably need the following:

- Helm based deployment in the operand repo that deploys the gateway (without operator)
- Gateway focused e2e test suite in the operand repo that executes against functionality of the gateway (creates the config secret with expected config, sets the needed env vars / flags to enable the features etc under test , uses the test servers). Main functionality it tested here.

- Helm based deployment in the operator repo that deploys the operator
- Integration (controller envtest  suite )in the operator repo that tests it created the deployment and secret etc correctly based on MCPServerReg , MCPGatewayExt etc...
- Manually triggered job that deploys version x of the operand with version y of the operator. (could execute nightly also, and must be executed on release).

So to summerise:

  Operand repo — tests the gateway as a gateway. Direct config secret, direct flags, no operator in the loop. This is where most feature work lands and where most e2e coverage lives.

  Operator repo — tests the controller as a controller. envtest proves it produces correct Deployments/Secrets/RBAC from CRs. No real gateway needed.

  Cross-repo job — the integration point. Manually triggered or nightly, catches version skew. Required gate before release.

Or we just keep it as a mono repo :)

Patryk-Stefanski: I added the e2e suggestion to the design doc, keeping it a mono repo I think is a wider discussion 😆

docs/design/repo-split/repo-split-design.md
```diff
@@ -0,0 +1,359 @@
...
+│   ├── controller/      # all reconcilers + server validator
+│   └── config/          # config secret writer + serialization types
+├── charts/mcp-gateway/  # Helm chart (CRDs, RBAC, controller, MCPGatewayExtension)
+├── bundle/              # OLM manifests
+├── config/
+│   ├── crd/             # generated CRD YAMLs
+│   ├── mcp-system/      # K8s deployment manifests
+│   ├── test-servers/    # K8s manifests for test servers
+│   └── samples/         # example manifests
+├── tests/e2e/           # full e2e tests (pulls released operand image)
```
maleck13: this is out of date now

docs/design/repo-split/repo-split-design.md
```diff
@@ -0,0 +1,359 @@
...
+```
+
+Both repos define their own Go structs for this schema. The operator serializes it; the operand deserializes it. Unknown fields are ignored by the operand, enabling forward compatibility.
+
+**Schema evolution rules:**
+- The operand must ignore unknown fields (forward compatibility).
+- The operator must not remove a field without a deprecation period spanning at least one operand release, preventing older operands from receiving a config missing a field they expect.
+- New required fields must have sensible defaults so older operators that don't set them still produce valid config.
+- Breaking changes require a coordinated release with both repos tagged.
+- The config includes a top-level `configVersion: 1` field. The operand checks this on load — known versions are processed normally; unknown versions log a warning with the expected and actual values. This enables detection of operator/operand version mismatches without hard failures.
```
maleck13: what is deciding the configVersion? Is that based on something like the operator version?

docs/design/repo-split/repo-split-design.md
```diff
@@ -0,0 +1,359 @@
...
+| Contract | Type | Details |
+|---|---|---|
+| Image | `RELATED_IMAGE_ROUTER_BROKER` | Env var on operator, defaults to `ghcr.io/kuadrant/mcp-gateway:latest` |
+| CLI flags | Stable set | `--mcp-broker-public-address`, `--mcp-gateway-private-host`, `--mcp-gateway-config`, `--mcp-check-interval`, `--mcp-gateway-public-host`, `--enable-url-elicitation` |
+| Env vars | Stable set | `JWT_SESSION_SIGNING_KEY`, `TRUSTED_HEADER_PUBLIC_KEY`, `CACHE_CONNECTION_STRING` |
+| Ports | Fixed | 8080 (HTTP), 50051 (gRPC), 8181 (config) |
+| Health | HTTP | `GET /readyz` on port 8080 |
+| Config mount | Volume | `/config/config.yaml` |
+
+### MCP Spec Considerations
```
maleck13: I think we cab remove this section. It has no bearing on the the actual design imo

docs/design/repo-split/repo-split-design.md
```diff
@@ -0,0 +1,359 @@
...
+| `images.yaml` | Push to main, tags | Build + push `ghcr.io/kuadrant/mcp-controller` |
+| `helm-release.yaml` | Tags | Publish Helm chart |
+| `helm-install-test.yaml` | PR | Helm install smoke test |
+| `verify-crd-sync.yaml` | PR | Ensure generated CRDs match chart CRDs |
+
+#### Cross-Repo Compatibility
+
+A workflow in the operator repo that deploys both components together and runs full integration tests:
+
+- **Nightly:** Latest released operator + latest released operand. Catches drift.
```
maleck13: I think this should run main against main so build a "nigthly" or something like that tag and execute the :nightly operator with the :nightly operand

docs/design/repo-split/repo-split-design.md
```diff
@@ -0,0 +1,359 @@
...
+- **Manual:** Accepts image overrides for both components (e.g., operand PR image + operator PR image). Used for coordinated changes that touch both repos.
```
maleck13: This manual one allows us to do something custom. Like run rc1 of the operand with v1.5 of the operator cause there were no changes to the operator code

docs/design/repo-split/repo-split-design.md
```diff
@@ -0,0 +1,359 @@
...
+
+## Additional Artifacts
+
+### Makefile
+
+The current `Makefile` has targets for both components (`build`, `docker-build`, `docker-build-controller`, `generate`, `manifests`, etc.). Post-split:
+
+- **Operator repo** gets a new `Makefile` with controller-specific targets: `generate`, `manifests`, `docker-build`, `test-unit`, `test-controller-integration`, `lint`, `helm-*`, and CRD generation (`controller-gen`).
+- **Operand repo** retains the existing `Makefile` but removes controller-specific targets (`docker-build-controller`, `generate`, `manifests`, CRD generation). Keeps `build`, `docker-build`, `test-unit`, `lint`, and performance test targets.
+
```
maleck13, line 304: We need to have an operator make command that will deploy the operator with a sepecific version of the operand. So from the operand we could run `make load-image` and then `make deploy-operator OPERAND_IMAGE=xxxx:y

docs/design/repo-split/repo-split-design.md
```diff
@@ -0,0 +1,359 @@
...
+- **Operator repo** `CLAUDE.md`: CRD types, controller reconciliation, config secret writing, Helm chart, OLM bundle, operator deployment.
+- **Operand repo** `CLAUDE.md`: broker, router, session management, upstream MCP connections, ext_proc protocol, performance guidelines.
+
+### docs.kuadrant.io
+
+User-facing guides are published at docs.kuadrant.io from the `Kuadrant/docs.kuadrant.io` repo. The nav config in `mkdocs.yml` references files from the MCP Gateway section. Post-split, these references need updating to point at the operator repo since guides and CRD reference docs move there. This update should be part of Phase 2 step 4 (CI setup). No redirects are needed — the docs repo pulls files by path from a configured repo, so changing the source repo in `mkdocs.yml` is sufficient.
+
+### Test Server Images
+
+Test server source code (`tests/servers/`) stays in the operand repo and produces test server container images. The operator repo's e2e tests reference these images. The operator repo contains the Kubernetes manifests (`config/test-servers/`) that deploy them but does not build them — it pulls pre-built images from the operand repo's CI.
```
maleck13: Do we still need tests/servers in the operator repo? Perhaps we just need one so we can test MCPServerRegistration?

maleck13 (commented):
few more minor things but getting there

docs/design/repo-split/repo-split-design.md
Patryk-Stefanski: It's a schema version, not tied to the operator release version. Renamed to `schemaVersion` to make that clearer.

It only bumps when the config secret format has a breaking structural change (field renames, type changes, removed fields). Additive changes like new optional fields don't bump it. This way the operand just checks "do I understand this config shape?" — a simple integer comparison — without needing to know anything about operator release versions. The operator can ship many releases without the schema version changing at all.

david-martin:
The doc covers which code moves where, but the user-facing docs boundary could be stated more explicitly. The "What Moves" table lists `docs/guides/` and `docs/reference/`, and the docs.kuadrant.io section mentions updating mkdocs.yml, but a reader (or agent/claude) could miss that all user-facing documentation lives in the operator repo post-split.

A few places where this could be clearer:

- **Repository Boundary table**: no mention of docs ownership. A row for user-facing docs (or a note that the operand repo has none) would make the boundary obvious upfront.
- **Repository Structure trees**: neither tree shows `docs/guides/` or `docs/reference/`. The operator tree should reflect that these live there.
- **docs.kuadrant.io section**: says to update mkdocs.yml references but doesn't explicitly state the principle: all user-facing docs live in the operator repo.
- **Phase 3 cleanup**: lists code to remove from the operand but doesn't mention removing `docs/guides/` and `docs/reference/`.

david-martin:
Let's hold this, pending alignment with https://github.com/Kuadrant/architecture/pull/179
@maleck13 I know you had some thoughts on how including mcp gateway in the umbrella operator work could substitute for this work

maleck13:
Yes it would be worth thinking about. Rather than splitting out the operator. That said Kuadrant is not currently a multi-tenant deployment but MCP-Gateway is. If we could put the effort we were going to put into this work into the umbrella operator approach that would likely be a better use of time

maleck13:
cc @mikenairn

Patryk-Stefanski:
/closing for now , superseded by https://github.com/Kuadrant/mcp-gateway/pull/1198

PR #1102, opened 2026-06-08
condition for jwt hs256 extending till 32 bytes

### LINKED ISSUE #1086: No minimum signing key length for JWT HS256
## Description

`NewJWTManager` (`session/jwt.go:66-67`) only checks `signingKey == ""`. A 1-character key is accepted. HS256 with a short key is brute-forceable. The encryption path (`DeriveEncryptionKey` in `crypto.go:19`) enforces 16 bytes, but the JWT signing path does not.

## Impact

An operator who sets `GATEWAY_SIGNING_KEY=x` gets a running system with trivially forgeable session and backend-init tokens.

## Proposed Fix

Add minimum key length check in `NewJWTManager`:

```go
if len(signingKey) < 32 {
    return nil, fmt.Errorf("signing key must be at least 32 bytes for HS256")
}
```

## Review Guidance

- `internal/session/jwt.go:64-80`
- Check if any tests use short keys that would need updating

## Manual Verification

1. Try starting with `GATEWAY_SIGNING_KEY=x`
2. Before fix: starts. After fix: exits with error

malladinagarjuna2:
Fixes #1086

Just have doubt with the  imutable whether it is correct or not ? or is it  immutable ?? i have to keep immutable for the passing of lint checks 
 
kept a condition that signing key length should'nt be more than 32. 
changed the unit test description wherever it was necessary

## Summary by CodeRabbit

* **Bug Fixes**
  * Enforced a minimum 32-byte `GATEWAY_SIGNING_KEY` length for JWT signing and session encryption key derivation.
  * Updated error handling to reject keys shorter than the required length.

* **Tests**
  * Updated JWT tests to use 32+ byte signing keys and added coverage for invalid (31-byte) keys.
  * Updated encryption/decryption tests to use longer key material for key derivation.

* **Documentation**
  * Added a breaking-change release note describing the 32-byte minimum key requirement and migration steps.

### DIFF
--- a/internal/mcp-router/request_handlers_test.go
+++ b/internal/mcp-router/request_handlers_test.go
@@ -24,6 +24,8 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+const testSigningKey = "test-signing-key-must-be-at-least-32-bytes"
+
 func TestMCPRequestValid(t *testing.T) {
 
 	testCases := []struct {
@@ -169,7 +171,7 @@ func TestHandleRequestBody(t *testing.T) {
 	require.NoError(t, err)
 
 	// Create JWT manager for test
-	jwtManager, err := session.NewJWTManager("test-signing-key-must-be-at-least-32-bytes", 0, logger, cache)
+	jwtManager, err := session.NewJWTManager(testSigningKey, 0, logger, cache)
 	require.NoError(t, err)
 
 	// Generate a valid JWT token
@@ -471,7 +473,7 @@ func TestValidateSession(t *testing.T) {
 	cache, err := session.NewCache()
 	require.NoError(t, err)
 
-	jwtManager, err := session.NewJWTManager("test-signing-key-must-be-at-least-32-bytes", 0, logger, cache)
+	jwtManager, err := session.NewJWTManager(testSigningKey, 0, logger, cache)
 	require.NoError(t, err)
 
 	validToken := jwtManager.Generate()
@@ -761,7 +763,7 @@ func TestHandleElicitationResponse(t *testing.T) {
 		cache, err := session.NewCache()
 		require.NoError(t, err)
 
-		jwtManager, err := session.NewJWTManager("test-signing-key-must-be-at-least-32-bytes", 0, logger, cache)
+		jwtManager, err := session.NewJWTManager(testSigningKey, 0, logger, cache)
 		require.NoError(t, err)
 
 		validToken := jwtManager.Generate()
@@ -845,7 +847,7 @@ func TestHandleElicitationResponse(t *testing.T) {
 		cache, err := session.NewCache()
 		require.NoError(t, err)
 
-		jwtManager, err := session.NewJWTManager("test-signing-key-must-be-at-least-32-bytes", 0, logger, cache)
+		jwtManager, err := session.NewJWTManager(testSigningKey, 0, logger, cache)
 		require.NoError(t, err)
 
 		validToken := jwtManager.Generate()
@@ -880,7 +882,7 @@ func TestHandleElicitationResponse(t *testing.T) {
 		cache, err := session.NewCache()
 		require.NoError(t, err)
 
-		jwtManager, err := session.NewJWTManager("test-signing-key-must-be-at-least-32-bytes", 0, logger, cache)
+		jwtManager, err := session.NewJWTManager(testSigningKey, 0, logger, cache)
 		require.NoError(t, err)
 
 		validToken := jwtManager.Generate()
@@ -921,7 +923,7 @@ func TestHandleElicitationResponse(t *testing.T) {
 		cache, err := session.NewCache()
 		require.NoError(t, err)
 
-		jwtManager, err := session.NewJWTManager("test-signing-key-must-be-at-least-32-bytes", 0, logger, cache)
+		jwtManager, err := session.NewJWTManager(testSigningKey, 0, logger, cache)
 		require.NoError(t, err)
 
 		validToken := jwtManager.Generate()
@@ -978,7 +980,7 @@ func TestHandleElicitationResponse_ViaRouteMCPRequest(t *testing.T) {
 	cache, err := session.NewCache()
 	require.NoError(t, err)
 
-	jwtManager, err := session.NewJWTManager("test-signing-key-must-be-at-least-32-bytes", 0, logger, cache)
+	jwtManager, err := session.NewJWTManager(testSigningKey, 0, logger, cache)
 	require.NoError(t, err)
 
 	validToken := jwtManager.Generate()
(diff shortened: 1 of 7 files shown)


malladinagarjuna2:
@jasonmadigan  kindly review this pr!

malladinagarjuna2:
just realised that for local testing i was using make-lint, which includes the spell check, thus this error was found!, whereas in the github actions make lint-go is used which expels the use of spell, it only runs check-gofmt, check-goimports, check-newlines, etc. Thus this spelling mistake was ignored. @jasonmadigan sir  am i right ?? i think there is  no word like imutable, the only word is immutable i guess

malladinagarjuna2:
image
realised that these are not required so i am removing them

malladinagarjuna2:
hey @jasonmadigan  removed all the unimportant lines kindly review the pr

jasonmadigan:
tests are failing

malladinagarjuna2:
@jasonmadigan sir, i have made the changes kindly review

malladinagarjuna2:
@jasonmadigan  kindly review my pr !!

malladinagarjuna2:
hey @jasonmadigan  @david-martin @Patryk-Stefanski  can you please review this pr

internal/session/jwt.go
```diff
@@ -63,8 +63,8 @@ type JWTManager struct {
 
 // NewJWTManager creates a new JWT manager with the provided signing key
 func NewJWTManager(signingKey string, sessionLength int64, logger *slog.Logger, sessionHandler Deleter) (*JWTManager, error) {
-	if signingKey == "" {
-		return nil, fmt.Errorf("no signing key provided")
+	if len(signingKey) < 32 {
```
jasonmadigan, line 66: **blocker:** this is a breaking change for deployments with `GATEWAY_SIGNING_KEY` between 16 and 31 bytes. per project rules (`.claude/rules/breaking-changes.md`), env var changes need a release note in `docs/release-notes/` with migration steps.

include: new minimum requirement, example key generation command (`openssl rand -base64 32`), and note that this is enforced at startup.

also: `DeriveEncryptionKey` in `crypto.go:19` still accepts keys >= 16 bytes. since both consume the same `GATEWAY_SIGNING_KEY`, consider aligning that floor to 32 as well -- the JWT check catches it first, but the inconsistency is confusing.

internal/session/jwt_test.go
```diff
@@ -31,7 +31,7 @@ func testLogger() *slog.Logger {
 
 func TestNewJWTManager(t *testing.T) {
 	t.Run("with custom key", func(t *testing.T) {
-		key := "test-signing-key"
+		key := "test-signing-key-must-be-at-least-32-bytes"
```
jasonmadigan: **blocker:** missing boundary test for the new validation. the PR's core change is `len < 32` but no test exercises a key of length 1-31. the empty-string test at line 64 passes incidentally because `len("") < 32`, but it doesn't validate the new behaviour.

add at minimum:
- 31-byte key -> expect error
- 32-byte key -> expect success

this is the primary acceptance criterion for #1086.

internal/session/jwt_test.go
```diff
@@ -74,7 +74,7 @@ func TestNewJWTManager(t *testing.T) {
 }
 
 func TestGenerate(t *testing.T) {
-	manager, _ := NewJWTManager("test-key", 0, testLogger(), nil)
+	manager, _ := NewJWTManager("test-key-must-be-at-least-32-bytes", 0, testLogger(), nil)
```
jasonmadigan: repeated string `"test-key-must-be-at-least-32-bytes"` appears ~14 times in this file (and ~12 more in `request_handlers_test.go`). extract to a `const testSigningKey` at the top of each file -- reduces noise and makes future length changes a one-line edit.

internal/mcp-router/request_handlers_test.go
```diff
@@ -169,7 +169,7 @@ func TestHandleRequestBody(t *testing.T) {
 	require.NoError(t, err)
 
 	// Create JWT manager for test
-	jwtManager, err := session.NewJWTManager("test-signing-key", 0, logger, cache)
+	jwtManager, err := session.NewJWTManager("test-signing-key-must-be-at-least-32-bytes", 0, logger, cache)
```
jasonmadigan: same note -- `"test-signing-key-must-be-at-least-32-bytes"` is repeated 12 times in this file. a file-level `const testSigningKey` would clean this up.

jasonmadigan (changes requested):
approach is correct -- enforcing 32-byte minimum for HS256 per RFC 7518 Section 3.2. validation placement in the constructor is right (fail-fast at startup).

two blockers before merge, plus some cleanup items. details inline.

internal/mcp-router/request_handlers_test.go
malladinagarjuna2: hey i did this because the unit test werent passing

malladinagarjuna2: @jasonmadigan  at first i removed it, but again the unit test were failing

internal/session/jwt_test.go
malladinagarjuna2: ok sure understood it

internal/mcp-router/request_handlers_test.go
jasonmadigan: > hey i did this because the unit test werent passing

Not sure I understand why this would fix tests

malladinagarjuna2: no  sir  actually i understood your idea when i  read this comment 


> same note -- `"test-signing-key-must-be-at-least-32-bytes"` is repeated 12 times in this file. a file-level `const testSigningKey` would clean this up.

will implement it.  #979  is ready for eview kindly review it

internal/session/jwt.go
malladinagarjuna2: hey really sorry for this one! 
> also: DeriveEncryptionKey in crypto.go:19 still accepts keys >= 16 bytes. since both consume the same GATEWAY_SIGNING_KEY, consider aligning that floor to 32 as well -- the JWT check catches it first, but the inconsistency is confusing.

i don't know how did i miss this one 

> this is a breaking change for deployments with GATEWAY_SIGNING_KEY between 16 and 31 bytes. per project rules (.claude/rules/breaking-changes.md), env var changes need a release note in docs/release-notes/ with migration steps.

wasn't aware of the above one

malladinagarjuna2:
hey @jasonmadigan i have implemented all the changes kindly review my pr!!

jasonmadigan (approved):
round 1 blockers resolved: boundary tests added (31/32 byte), release note added, DeriveEncryptionKey aligned to 32. tests pass.

remaining nits (not blocking):
- jwt_test.go:67: test name "with empty key returns error" is stale -- tests len<32 now
- response_handlers_test.go:382: raw string instead of testSigningKey const
- request_handlers_test.go:1171: raw literal for other-key

PR #1105, opened 2026-06-09
Keep virtual server filtering scoped

### LINKED ISSUE #1082: Internal header re-injection allows virtual server scope bypass
## Description

Two related issues combine to allow a client to bypass virtual server scoping.

### M1: Internal header re-injection on broker passthrough path

`HandleNoneToolCall` (`request_handlers.go:860-866`) re-injects `x-mcp-authorized` and `x-mcp-virtualserver` from the original client-supplied headers after they were stripped in the headers phase (line 233). The stripping tells Envoy to remove them, but `localRequestHeaders` (line 104) retains the original client headers, which are copied to `mcpRequest.Headers` (line 242). The re-injection reads from this stale copy.

Data flow:
1. Client sends `x-mcp-virtualserver: attacker-value`
2. Headers phase strips it (Envoy removes it)
3. Body phase: `HandleNoneToolCall` reads it from `mcpReq.Headers` (original headers still in memory) and re-injects it
4. Broker receives the client-controlled value

### L4: Virtual server filter fail-open

When `GetVirtualSeverByHeader` fails (unknown virtual server name), `applyVirtualServerFilter` (`filtered_tools_handler.go:222-226`) returns the full unfiltered tool list instead of an empty list.

### Combined impact

A client injects `x-mcp-virtualserver: nonexistent`, the router re-injects it, the broker's lookup fails, and all tools are returned unfiltered -- bypassing virtual server scoping entirely.

## Proposed Fix

1. **M1:** Compute `x-mcp-authorized` and `x-mcp-virtualserver` server-side from validated session state and routing config, not from client-supplied headers
2. **L4:** Return empty tools on virtual server lookup failure (fail-closed)

## Review Guidance

- `internal/mcp-router/request_handlers.go:860-866` -- the re-injection path
- `internal/broker/filtered_tools_handler.go:212-226` -- the fail-open filter
- Check all other uses of `internalOnlyHeaders` for similar patterns

## Manual Verification

1. Deploy with virtual server configured
2. Send a `tools/list` request with header `x-mcp-virtualserver: nonexistent`
3. Before fix: all tools returned. After fix: empty or error

manik3160:
Fixes https://github.com/Kuadrant/mcp-gateway/issues/1082

This stops the router from re-adding client-supplied internal filter headers on the broker passthrough path after Envoy strips them.

It also makes tool filtering fail closed when `x-mcp-virtualserver` points to a missing virtual server, so the broker returns no tools instead of the full unfiltered list.

Tests run:
- go test ./internal/mcp-router
- go test ./internal/broker

## Summary by CodeRabbit

* **Bug Fixes**
  * Validation for virtual-server headers now rejects unknown values with a 400 and causes filtered results to be empty instead of returning unfiltered items.
  * Router no longer reinjects internal-only headers and will forward authorization headers when present.

* **Tests**
  * Tests updated and expanded to cover stricter filtering, header handling, and 400 response behavior.

* **Chores**
  * CI lint installs now retry on failure; tooling version bumped.

### DIFF
--- a/internal/mcp-router/headers.go
+++ b/internal/mcp-router/headers.go
@@ -36,7 +36,7 @@ var internalOnlyHeaders = []string{mcpAuthorizedHeader, mcpVirtualServerHeader, 
 
 // clientStrippedHeaders are headers that the router unconditionally strips from
 // incoming client requests during the headers phase.
-var clientStrippedHeaders = []string{mcpVirtualServerHeader, mcpVerifiedSubHeader}
+var clientStrippedHeaders = []string{mcpAuthorizedHeader, mcpVirtualServerHeader, mcpVerifiedSubHeader}
 
 func getSingleValueHeader(headers *basepb.HeaderMap, name string) string {
 	if headers == nil {
--- a/internal/mcp-router/headers.go
+++ b/internal/mcp-router/headers.go
@@ -36,7 +36,7 @@ var internalOnlyHeaders = []string{mcpAuthorizedHeader, mcpVirtualServerHeader, 
 
 // clientStrippedHeaders are headers that the router unconditionally strips from
 // incoming client requests during the headers phase.
-var clientStrippedHeaders = []string{mcpAuthorizedHeader, mcpVirtualServerHeader, mcpVerifiedSubHeader}
+var clientStrippedHeaders = []string{mcpVirtualServerHeader, mcpVerifiedSubHeader}
 
 func getSingleValueHeader(headers *basepb.HeaderMap, name string) string {
 	if headers == nil {
--- a/internal/mcp-router/headers.go
+++ b/internal/mcp-router/headers.go
@@ -34,6 +34,10 @@ const (
 // and routing that must be stripped before forwarding to upstream MCP servers.
 var internalOnlyHeaders = []string{mcpAuthorizedHeader, mcpVirtualServerHeader, mcpVerifiedSubHeader}
 
+// clientStrippedHeaders are headers that the router unconditionally strips from
+// incoming client requests during the headers phase.
+var clientStrippedHeaders = []string{mcpVirtualServerHeader, mcpVerifiedSubHeader}
+
 func getSingleValueHeader(headers *basepb.HeaderMap, name string) string {
 	if headers == nil {
 		return ""
--- a/internal/mcp-router/request_handlers.go
+++ b/internal/mcp-router/request_handlers.go
@@ -872,10 +872,6 @@ func (s *ExtProcServer) HandleNoneToolCall(ctx context.Context, mcpReq *MCPReque
 
 	}
 	headers.WithMCPServerName("mcpBroker")
-	// re-inject x-mcp-authorized so the broker can apply auth-based tool filtering
-	if v := mcpReq.GetSingleHeaderValue(mcpAuthorizedHeader); v != "" {
-		headers.WithCustomHeader(mcpAuthorizedHeader, v)
-	}
 	virtualServerID, ok, err := s.validVirtualServerHeader(mcpReq)
 	if err != nil {
 		s.Logger.WarnContext(ctx, "rejecting broker request: invalid virtual server header", "error", err)
--- a/internal/mcp-router/request_handlers.go
+++ b/internal/mcp-router/request_handlers.go
@@ -236,12 +236,12 @@ func (s *ExtProcServer) HandleRequestHeaders(ctx context.Context, headers *eppb.
 	// the token is already trusted. We surface the verified sub as the internal
 	// x-mcp-verified-sub header so the broker can bind token submissions to an
 	// identity without re-parsing the raw token. The header is in
-	// internalOnlyHeaders, so any client-supplied value is stripped first.
+	// clientStrippedHeaders, so any client-supplied value is stripped first.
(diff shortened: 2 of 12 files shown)


internal/mcp-router/request_handlers.go
```diff
@@ -858,16 +858,38 @@ func (s *ExtProcServer) HandleNoneToolCall(ctx context.Context, mcpReq *MCPReque
...
-		}
+	// re-inject x-mcp-authorized so the broker can apply auth-based tool filtering
+	if v := mcpReq.GetSingleHeaderValue(mcpAuthorizedHeader); v != "" {
+		headers.WithCustomHeader(mcpAuthorizedHeader, v)
+	}
+	virtualServerID, ok, err := s.validVirtualServerHeader(mcpReq)
+	if err != nil {
+		s.Logger.WarnContext(ctx, "rejecting broker request: invalid virtual server header", "error", err)
+		return response.WithImmediateResponse(400, "bad request").Build()
+	}
```
jasonmadigan, line 885: this still re-injects x-mcp-authorized from client-supplied headers. #1082 recommended computing this server-side from validated session state. as-is, a client can inject an arbitrary value here.

jasonmadigan (changes requested):
virtual server validation and fail-closed behaviour look correct. one issue: x-mcp-authorized is still re-injected from client-supplied headers (line 868). #1082 flagged this as a bypass risk -- the value should come from validated session state, not from what the client sent. the VS half is fixed properly, but the authorized header needs the same treatment before this fully closes #1082.

internal/mcp-router/request_handlers.go
manik3160: Good catch !

manik3160:
Thanks @jasonmadigan ! I've removed the re-injection entirely since Authorino sets it on the request anyway, so it still reaches the broker without needing to be copied from client headers.

manik3160:
Hey @jasonmadigan ! just following up on this PR. I've addressed the review feedback and all checks are passing. Would appreciate another look when you have time. Thanks!

Patryk-Stefanski:
@manik3160 please resolve conflicts

manik3160:
Sure @Patryk-Stefanski !

Patryk-Stefanski (changes requested):
The core fixes are correct — failing closed on missing virtual server and stopping the re-injection of client-supplied headers on the broker path are both the right calls.

One blocking issue before this can merge, plus two minor nits.

---

**Blocking — `x-mcp-authorized` accidentally removed from client stripping**

`clientStrippedHeaders` no longer includes `mcpAuthorizedHeader`:

```go
var clientStrippedHeaders = []string{mcpVirtualServerHeader, mcpVerifiedSubHeader}
```

The PR's intent was to stop the router *re-injecting* `x-mcp-authorized` — correct. But it also stopped *stripping* the client-supplied value in the headers phase, which is a separate concern and should stay.

The consequence: if a client sends any value in `x-mcp-authorized`, Authorino appends its wristband JWT rather than overwriting, so the broker receives 2 values. `parseAuthorizedCapabilitiesJWT` requires exactly 1 value (`internal/broker/filtered_tools_handler.go:130`) and returns an error when it gets 2. The broker then returns an empty tool list — even when `enforceCapabilityFilter` is false. Any client that sends this header gets silently empty tools.

Fix — add `mcpAuthorizedHeader` back to `clientStrippedHeaders`:

```go
var clientStrippedHeaders = []string{mcpAuthorizedHeader, mcpVirtualServerHeader, mcpVerifiedSubHeader}
```

Stripping the client value in the headers phase and not re-injecting it in the body phase are both independently correct.

---

**Nit — `validVirtualServerHeader` returns 400 when `RoutingConfig` is nil**

A nil `RoutingConfig` is a server error, not a bad client request. The 400 status code is misleading. Since `HandleRequestHeaders` already dereferences `RoutingConfig` unconditionally and would panic first, this path is unreachable in practice — but if the guard is there for safety, a 500 would be more accurate.

---

**Nit — nil vs empty slice inconsistency in `applyVirtualServerFilter`**

On error the function returns `[]mcp.Tool{}` (non-nil empty slice), but when a VS exists and no tools match, `filtered` is `nil`. Both mean "no tools" and callers using `len()` are unaffected, but it's worth making consistent — use `var filtered []mcp.Tool` and return that on error too.

manik3160:
Thanks @Patryk-Stefanski ! fixed all the points

internal/mcp-router/headers.go
Patryk-Stefanski, line 1: `mcpAuthorizedHeader` is still missing from `clientStrippedHeaders` — flagged in https://github.com/Kuadrant/mcp-gateway/pull/1105#pullrequestreview-4579215615 as a blocking issue.

```go
var clientStrippedHeaders = []string{mcpVirtualServerHeader, mcpVerifiedSubHeader}
```

This means a client-supplied `x-mcp-authorized` value is no longer stripped in the headers phase. The fix is one line:

```go
var clientStrippedHeaders = []string{mcpAuthorizedHeader, mcpVirtualServerHeader, mcpVerifiedSubHeader}
```

Stopping re-injection in the body phase and stripping in the headers phase are two independent concerns — both need to be in place.

david-martin:
Closing due to inactivity

PR #1101, opened 2026-06-08
removed the mcp lock from validateservertools

malladinagarjuna2:
closes #1088

 fixes a race condition where OnConfigChange could modify upstream servers between tool validation and scope application during a select_tools request.

- Held mcpLock.RLock() across the entire handleSelectTools transaction.
- Refactored validateToolSelection to validateToolSelectionLocked.
- Fixed a minor spelling typo (imutable -> immutable).
- Tested with go test -race to guarantee thread safety.

## Summary by CodeRabbit

* **Chores**
  * Fixed documentation typo in configuration comments.

* **Refactor**
  * Optimized internal tool validation performance by reducing lock acquisition overhead.

PR #1113, opened 2026-06-09
ensured that inmemorymap stores the encrypted token and it must be de…

### LINKED ISSUE #1087: In-memory cache stores user tokens in plaintext
## Description

When Redis is unavailable, the session cache falls back to in-memory storage. `SetUserToken` (`cache.go:149-164`) stores plaintext tokens directly in the in-memory branch. The encryption block at lines 167-173 is only reached in the Redis code path.

## Impact

If an attacker achieves memory read (core dump, /proc/mem, debug endpoint like pprof), user upstream tokens are recoverable in plaintext. With Redis, they would be AES-256-GCM encrypted.

## Proposed Fix

Either:
1. Apply `encrypt`/`decrypt` in the in-memory path too
2. Or document that in-memory mode is dev-only and must not be used in production with real user tokens

## Review Guidance

- `internal/session/cache.go:148-175` (SetUserToken) and `179-221` (GetUserToken)
- Compare the in-memory and Redis code paths

malladinagarjuna2:
fixes #1087
 - just moved the encryption block to the top of setusertoken section 
 - and added decryption block to the getusertoken section

## Summary by CodeRabbit

* **Bug Fixes**
  * Ensured tokens are encrypted consistently across all storage backends for enhanced security
  * Improved session configuration validation to catch configuration errors earlier during initialization

### DIFF
--- a/cmd/mcp-broker-router/main.go
+++ b/cmd/mcp-broker-router/main.go
@@ -206,26 +206,26 @@ func (a *app) setupSessionInfra(ctx context.Context) {
 		}
 	}
 
-	var sessionCacheOpts []func(*session.Cache)
-	sessionCacheOpts = append(sessionCacheOpts, session.WithRedisClient(a.redisClient))
-	if a.redisClient != nil {
-		encKey, err := session.DeriveEncryptionKey([]byte(a.brokerCfg.gatewaySigningKey))
-		if err != nil {
-			panic("failed to derive encryption key: " + err.Error())
-		}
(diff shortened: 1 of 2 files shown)


jasonmadigan:
thanks @malladinagarjuna2, closing -- we've closed #1087 as the in-memory plaintext threat model doesn't warrant the fix (an attacker with pod memory access has bigger problems). your other PRs (#1101, #1102) are solid though.

Issue #166, opened 2025-09-11
Support External MCP Servers


david-martin:
Need to support both external/remote MCP servers and those internally hosted.

At this time, the MCPServer CRD targets a HTTPRoute.
The HTTPRoute examples are all for Service backends in the same cluster.

It may be possible to use an Istio DestinationRule and ServiceEntry, but it needs to be tested.
There's a guide in the kuadrant project that uses this kind of approach to route to an external host https://github.com/Kuadrant/kuadrant-operator/blob/db8e2d1747cfcd4be69769d2a627bdfd5256814c/doc/user-guides/misc/external-api.md

jasonmadigan:
Will take a look at testing this, should work. Will try with something like the GitHub public MCP server.

PR #1114, opened 2026-06-10
docs: add A2A protocol support design document

Aman-Cool:
Design proposal for A2A protocol support in the MCP Gateway, refs #766 ; this document describes the full direction; agent registration and discovery, path-per-agent routing through ext_proc, and per-principal task ownership.., and, following review, an implementation-phases section that lands it incrementally behind an experimental --enable-a2a router flag, default off, with any new APIs alpha.

Phase 1 is the slice we've agreed to build first (#1333): router-only passthrough for auditing, auth and observability. The router strips inbound x-a2a-* headers, parses /a2a POST bodies envelope-only, and sets x-a2a-method and x-a2a-agent (the agent identity being the first path segment after /a2a/, a documented convention since the A2A body names no agent) ; unparseable POSTs fail closed, GETs pass through, users hand-author their agent HTTPRoutes, and the headers feed Istio Telemetry and AuthPolicy with no further gateway code. Registration/discovery and task ownership follow as phases 2 and 3; kept as the direction we're cutting from rather than a commitment, gated on where A2A support lands long-term and on the broker settling after the MCP spec-version work.

A few design positions worth calling out for review context: signed v1.0 AgentCards are served verbatim (a URL rewrite breaks the JWS over the JCS-canonicalized card), with the catalog advertising gateway paths alongside rather than inside the card ; task IDs are agent-assigned and pass through unchanged; routing is by path, so the gateway never mints or rewrites them, and ownership is an insert-only (agent, taskID) -> principal record that fails closed on any miss or mismatch ; and A2A authenticates per-request via OAuth bearer with ownership scoped to the principal, so it is untouched by MCP's stateless changes. The doc's decisions have been exercised end-to-end in a working prototype; discovery, routing and streaming passthrough against a live gateway; so the phases start from validated ground rather than paper.

All review feedback to date is incorporated and the open markers are resolved ; the discovery convention (RFC 9727 catalog vs the emerging AI Catalog) is deliberately held loose until phase 2 ships a catalog at all. Happy to adjust anything that would make this easier to build against as phase 1 lands.


Aman-Cool:
Hey @david-martin, @jasonmadigan and @maleck13, 

Did some digging on Q2 (which spec version to target) since this blocks the test server work; turns out the spec has moved more than I expected, so wanted to lay out what I found.

There are effectively three versions in the wild right now:

- **v0.2.x** : what this design doc currently assumes. Methods are `message/send`, `tasks/get`, `tasks/cancel`, agent card at `/.well-known/agent.json`.
- **v0.3.0** : same method names, but the agent card moved to `/.well-known/agent-card.json`. That path change was breaking enough that there are open issues on `google/adk-python` about clients failing after the upgrade, and the general guidance that emerged is to serve both paths.
- **v1.0 (Linux Foundation, early this year)** : bigger shift. Methods got renamed entirely (`message/send` → `SendMessage`, `tasks/get` -> `GetTask`, `tasks/resubscribe` -> `SubscribeToTask`, plus a new `ListTasks`), agent cards can be signed, and there's an `A2A-Version` header for negotiation.

My suggestion: target v0.3.0 for this term. Every method name in the design doc and the routing logic stays valid; the only real change is serving the card at `agent-card.json` (with `agent.json` kept as a legacy alias, which costs us one extra mux line). v1.0 would mean renaming every method constant, dealing with card signing, and version negotiation; feels like a lot of surface for a prototype, and the v0.3 -> v1.0 migration would be a well-scoped follow-up once the architecture is proven.

One wrinkle worth flagging: the official Go SDK (`a2aproject/a2a-go/v2`) only speaks v1.0 now, so for the test server I'd hand-roll the JSON-RPC handling rather than pull in the SDK; which honestly suits us better anyway, since the e2e tests need the server to produce controlled edge cases (SSE chunks split mid-JSON, keepalives, etc.) that an SDK would abstract away.

If this sounds reasonable I'll update the doc to reflect v0.3.0 + the dual card path and get moving on the test server.

Aman-Cool:
**Design conformance issue found during Week 2 prep: `message/send` has no `skill` parameter**

_(Cross-posting from the Slack group so the decision is tracked here.)_

While going through the A2A v0.3.0 spec line by line to build the test server, I found that the protocol has no `skill` parameter in `message/send`. [`[MessageSendParams](https://a2a-protocol.org/v0.3.0/specification/#711-messagesendparams-object)`](https://a2a-protocol.org/v0.3.0/specification/#711-messagesendparams-object) is `{message, configuration, metadata}` ; skills exist only in the Agent Card for discovery. The protocol never routes by skill; a client sends message parts and the receiving agent decides what to do with them.

This design doc currently assumes skill-based dispatch in three places:
- The `message/send` flow diagrams (`GetAgentInfo("weather_forecast")` extracted from the request)
- The "agent sends a long-running task" job story ("using a skill prefix to identify the target agent")
- The e2e test cases ("message/send with a skill that matches a registered agent's prefix")

Standard A2A clients will never send that field. It also explains why [`[agentgateway](https://agentgateway.dev/docs/standalone/main/tutorials/a2a/)`](https://agentgateway.dev/docs/standalone/main/tutorials/a2a/) (Linux Foundation) does one-route-per-agent rather than skill dispatch; the protocol doesn't support it.

**Options**

1. **Gateway convention via `params.metadata.skill`** : `metadata` is an open `{[key: string]: any}` map designed for extensions, so this is legal. Keeps the single `/a2a` endpoint and the federated card as designed. Downside: only gateway-aware clients can route; off-the-shelf A2A clients can't use the gateway for first messages.

2. **Path-per-agent: `/a2a/{prefix}`** : each registered agent gets a path segment; the router routes on path alone at the headers phase. Any unmodified A2A client works. The federated card becomes a discovery directory and each agent keeps its own card at its own path. This simplifies the router significantly, and the gateway-owned task ID mapping stays for the ID-isolation security goal.

3. **Hybrid** : multi-turn messages already carry `message.taskId`, so continuations route via the task store; only first messages need agent selection (via path or metadata).

**Recommendation**

Option 2. It matches how the rest of the ecosystem solved this, works with unmodified clients, and removes complexity from the router rather than adding it. The skill-prefix federation idea survives in the discovery layer; just not in invocation routing.

**Impact on current work**

Not blocked: the test server's agent card, `tasks/get`, `tasks/cancel`, and SSE streaming are unaffected, so I'm building those now. The `message/send` routing handler and the corresponding design doc sections wait for this decision. Once decided, I'll update the flow diagrams, job stories, and e2e cases in this PR to match.

Aman-Cool:
Pushed an update to the design doc based on some spec research I did this week; wanted to share what I found and get your thoughts before moving forward.

The main thing: while going through the A2A v0.3.0 spec more carefully, I noticed that `message/send` has no `skill` parameter. `MessageSendParams` is just `{message, configuration, metadata}`; skills only appear in the `AgentCard` for discovery. So the original routing approach in the draft (extracting a skill from the request body) wouldn't actually work with a spec-compliant client.

Given that, I looked at a few options and updated the doc toward path-per-agent (`/a2a/{prefix}`) since it seemed like the cleanest fit; it's also what `agentgateway` does. For discovery I went with `/.well-known/api-catalog` (RFC 9264) instead of a merged `agent.json`. But I'm not certain either of these is the right call and **would really value your input before this gets locked in.**

Also took a first pass at Q1 (CRD group -> `mcp.kuadrant.io`) and Q2 (v0.3.0) based on what felt consistent with the rest of the codebase, **but happy to revisit both.**

Q4 (session handling) I've left open : planning to bring that to the Week 8 sync as originally discussed.

@david-martin, @jasonmadigan and @maleck13, What do you think? Anything I should reconsider or approach differently?

maleck13:
Thanks @Aman-Cool . This is on my list to review this week. Sorry at the moment just a bit swamped. Thank you for the effort put in so far

Aman-Cool:
@maleck13, genuinely no rush on the review. I know it's a swamped stretch, and I'd much rather you get to this with real headspace than rush a skim. I'm posting this less to add to your queue and more so that when you (and @jasonmadigan) get a window, the reasoning behind the routing change is in one place instead of scattered across the earlier comments.

One important caveat up front: this is where I've landed after going deep on the spec. @david-martin had an early look and the direction seemed reasonable to him, but it isn't a settled group decision.., you and @jasonmadigan haven't had a chance to weigh in yet, and I really don't want to lock it until you guys also have. So please read everything below as a proposal I'm fairly confident in, with the door wide open to push back; especially on the one piece I've flagged as still genuinely unresolved.

## How we got here

The original draft routed A2A requests by reading a skill out of the `message/send` body. Going through the v0.3.0 spec to build the test server, I hit the problem I flagged earlier: `message/send` has no skill to read. `MessageSendParams` is just `{message, configuration, metadata}`; skills only appear in the `AgentCard`, for discovery. So skill-dispatch wasn't something we could keep or drop by preference; it was leaning on a field the protocol doesn't have.

That leaves two honest ways to tell the gateway which agent a request is for:

1. **A path per agent** : `/a2a/{prefix}`, known at the request-headers phase from `:path`.
2. **A header** : something like `x-a2a-agent`, set by the caller.

I'm proposing the path. Reasoning below, and if either of you sees a third option I've missed, I'd genuinely like to hear it.

## Why I'm leaning path-per-agent

The header approach has the problem that a stock A2A client has no concept of `x-a2a-agent`, so only a gateway-aware client could ever set it, the same limitation as stuffing the routing key in metadata. It only works for clients we've specifically taught about the gateway.

The path doesn't need the client to know anything. A2A clients already fetch an agent card and POST to whatever `url` the card advertises, that's just the normal protocol flow. So if the gateway serves each agent's card with its `url` pointed at `…/a2a/{prefix}`, a completely unmodified client just follows it to the right place. The routing key lives in the card we serve, not in something the caller has to be told.

## The part that almost caught me out

I want to call this out specifically, because it's subtle and it changed how I wrote the discovery flow. The original draft had the broker simply proxying the upstream Agent Card through unchanged. That looks correct right up until you trace what a real client does with it: it reads the `url` field and sends `message/send` straight there. If we pass the upstream `url` through untouched, the client takes the card *from the gateway* and then talks *directly to the upstream agent*; bypassing the gateway entirely. No `AuthPolicy`, no `RateLimitPolicy`, no logging. The whole reason for the gateway quietly evaporates, and nothing errors; it just silently routes around us.

So rewriting the card's `url` to the gateway path isn't a cosmetic detail.., it's the load-bearing piece that makes path-per-agent actually work with unmodified clients. It took me a beat to see that clearly. It's now written up explicitly in the design doc, with its own acceptance criteria on the broker + test-server tasks and an assertion in the discovery e2e test, specifically so it can't silently regress later.

## Prior art

Worth noting this isn't us out on a limb. `agentgateway` (Solo.io, now under the Linux Foundation, same home as A2A) landed in the same place: route-per-agent rather than skill multiplexing, and they rewrite agent card URLs to point back at the gateway... the same mechanism. Reassuring to have a real implementation pulling the same way, though obviously I don't want "someone else did it" to stand in for your own read.

The skill-prefix federation idea from the original draft doesn't die, by the way. It just stops being the routing mechanism and becomes the discovery/index layer instead.

## The one piece still genuinely open; and where I'd most value your input

Discovery. A2A assumes one card per agent at the well-known path, so "how does a client find all the agents behind a single gateway" is a convention we have to define ourselves. The doc currently uses an RFC 9264 API Catalog at `/.well-known/api-catalog`, which is the shape a client would already expect for "list everything under one base URL." We're not alone in needing this, there's active upstream movement on exactly the same problem:

- `a2aproject/A2A#641` : multi-agent open discovery under a single base URL via an RFC 9264 catalog
- `a2aproject/A2A#883` : best practices for exposing multi-skill agents as unified external services
- `a2aproject/A2A` discussion #166 : hosting multiple agents under a single base URL

This is the part I'd treat as "track upstream and stay aligned" rather than lock down hard, and it's the spot where a second and third opinion would help most, whether we commit to the RFC 9264 catalog now or keep it deliberately loose while the upstream discussions settle.

One framing I'd put on the table, since I think it's a useful north star: this is a fast-moving spec and any specific shape of it could shift under us. The durable value the gateway adds is the auth / rate-limit / observability layer, which holds no matter how A2A itself settles; so I'm comfortable holding the discovery convention loosely as long as the routing and policy story underneath is solid.

## What's already in the PR

I've pushed the design doc and task plan updates so they reflect the path-per-agent approach and the `url` rewriting rather than the original skill-dispatch draft. The e2e cases are updated to match too. None of it is load-bearing on a decision being final, it's just easier to review the reasoning against an up-to-date doc than against the old one.

So: no urgency at all, but whenever you both have a moment, I'd really value your read on (1) path-per-agent as the routing mechanism and (2) how firmly we want to commit to the RFC 9264 catalog for discovery. I'll hold off treating either as settled until you've had the chance.

david-martin:
Hey @Aman-Cool, apologies for not flagging this sooner. We recently updated our [contributing guidelines](https://github.com/Kuadrant/mcp-gateway/blob/main/CONTRIBUTING.md) to ask contributors to limit themselves to one open PR at a time, so you may not have seen the change. The idea is that getting one change reviewed and merged is more valuable than having several open in parallel.

You currently have four open PRs. Which one do you think is best to focus on first? We'll prioritize getting that one through.

Aman-Cool:
Thanks @david-martin, and no worries at all; the guideline makes complete sense, and it's a good forcing function. I'd rather get one thing landed well than spread thin across several.

I'd like to focus on #1114 (the A2A design doc). It's the one the rest of my A2A work depends on.., until the routing approach and discovery convention are agreed, the CRD, controller, and router PRs would all be building on a design that isn't settled yet. CI's green and it's the furthest along, so it should be the cleanest to get over the line.

To keep things to a single active PR, I'll move my other open PRs to draft and bring them back one at a time once #1114 is merged. The two small bug fixes: the `MCPManager` status data race (#1005) and the `singleflight` session-init fix (#930), are independent of A2A and self-contained, so those are the natural ones to pick up next.

Really appreciate you prioritizing it.

Aman-Cool:
Small update.., spent the last few days poking the design with a bunch of "…ok but what actually happens when X?" questions, and each one turned out worth writing down. Six commits:

- **Policy enforcement** : how A2A traffic picks up `AuthPolicy`/ratelimit; with the honest caveat that skill-filtering is visibility, not access control (the agent is the real boundary, since `message/send` carries no skill).
- **SSE multimodal** : the passthrough parses only the JSON-RPC envelope to swap the task id; big base64 artifact bytes ride through as raw bytes, never decoded.
- **Card sync** : the spec has no card-changed push, so the broker polls cheaply (conditional GET + version/hash, act only on change) instead of thrashing the secret.
- **Upstream auth** : `credentialRef` stays discovery-only (like MCP); `message/send` uses the client's identity / token-exchange, not a static gateway cred; dodges the confused-deputy trap.
- **Tracing** : async tasks span separate traces, so every span gets `a2a.task.id` to stitch the whole lifecycle together in Tempo/Jaeger.
- **Push notifications** : still a non-goal, but sketched the webhook-relay shape so an agent can't call the client directly behind the gateway's back.

Funny thing: nearly all of them circle back to auth / ratelimit / observability.., which feels like where the gateway actually earns its keep. Happy to break any out into its own thread if that's easier to review 🙂

Aman-Cool:
Added a short "Card integrity (v1.0 signed cards)" note under Upstream Authentication.., nothing for the v0.3.0 target, but one for the Q2 call. v1.0 signs the `AgentCard` and the signature covers `url`, so a signed card can't be url-rewritten without invalidating it. The doc walks the three options and leans on the RFC 9727 catalog (no rewrite), which path-per-agent already gives us... so we're positioned for it either way. Just wanted it on record before the spec-version decision.

david-martin:
@Aman-Cool Some thoughts while reviewing the proposal. Let me know if something doesn't make sense.
I'd question the value of doing something just for the v0.3 api as it immediately is behind spec. There would have to be something important being investigated & derisked against that version to warrant not also taking 1.0 into scope, for there to be enough value in my opinion.

 1. Happy for the API Catalog endpoint to be included. I understand there's a manual config step for each
  agent card endpoint as agents don't standardize on the catalog endpoint as a way of configuring yet.
  2. I understand the request mode change mid-request (buffered vs streamed), and that Envoy should support it, but haven't seen it done before. Good to derisk this early.
  3. Status reporting has changed on main. Discovered tools no longer included, and "Ready" means it's
  configured not necessarily fully ready to take tool calls. Same pattern should be followed here.
  4. Cross-namespace targetRef is fixed (or at least PR'd?), so cross-namespace should be allowed.
  5. Oldest-wins for prefix collision is not great. Non-deterministic with reconcile and GitOps. Timing
  shouldn't matter in eventually consistent systems.
  6. Skill-level filtering should just say "not applicable to A2A" or similar. message/send doesn't name the skill, so filtering doesn't make sense.
  7. Dig deeper into v1.0 multi-tenancy model, particularly how signed cards work with multi-agent hosting
  and URL rewriting.
  8. Don't dismiss the tenant field for policy enforcement. ext_proc can extract it from the body and expose
  it as a header, same pattern as x-mcp-toolname, for authorino and limitador to use in policy
  9. Consider whether a fork of mcp-gateway is a better starting point for A2A exploration. The overlap is
  mainly ext_proc and CRD/controller patterns. The broker concept is much thinner for A2A (just a card
  cache, no aggregation or sessions), and the api catalog is new. A fork would let you move fast without MCP regression concerns.

Aman-Cool:
Thanks @david-martin, this is genuinely useful... a lot of it lands, and a couple I'd already started pulling on, so let me go through them.

On the v0.3-vs-1.0 framing.., that's actually the direction I'd already been moving, so good to get it on the table. I'd been into the proto at v1.0.1 and landed on the same read: v0.3 is already behind (v1.0.1 is current, the `a2a-go` SDK is v1.0-only), and the gap is structural rather than just renamed methods; proto is canonical now, no top-level card `url`, `supported_interfaces`, the `tenant` field, blocking-by-default `SendMessage`. So the plan I'd been forming is to shift the target to v1.0.., the hard parts we actually need to derisk; the mid-request mode change, SSE task-ID rewriting, policy integration, the catalog; are version-agnostic, so we derisk those while targeting v1.0's surface, with the version-specific bits (method names, well-known path, card shape) isolated behind one mapping so the version isn't load-bearing. I'll write that up as the explicit Q2 resolution.

On the quick ones, all fair:

- **(3) status** : agreed, I'll follow the new pattern: `Ready` = configured, not necessarily able to take calls, and I'll keep the card-discovered condition from over-promising.
- **(4) cross-namespace** : yep, that one's actually my issue (#1199), and #1139 is the fix (still draft). I'll design A2A to allow cross-namespace assuming it lands, rather than working around it.
- **(5) oldest-wins** : agreed, I don't love it either; the non-determinism is exactly the argument for namespace-qualifying the path (`/a2a/{namespace}/{prefix}`), which makes collisions structurally impossible instead of timing-dependent. Dropping oldest-wins for that.
- **(6) skill filtering** : right, `message/send` names no skill (§7.1.1), so I'll just say "not applicable to A2A" instead of dressing it up.
- **(2) mode change** : agree on derisking it early; good news is we're not starting from zero, the elicitation path already flips to STREAMED `ModeOverride` (`response_handlers.go`), so A2A is extending a proven mechanism.., the genuinely new bit is choosing `BUFFERED` vs `STREAMED` by method, which is what I'd spike first.

On (8) : you're right, and I'll fix that line. I conflated policy attachment with enforcement.., `ext_proc` can absolutely lift `tenant` into a header the same way the router sets `x-mcp-toolname`, and let Authorino/Limitador key off it. So `tenant` carries enforcement fine, path-per-agent's only real edge is per-agent attachment (distinct policies per agent) and being Gateway-API-native, not "tenant can't do policy."

Which connects straight to (7), and this is the part I got a bit excited about; I'd already gone into the v1.0 multi-tenancy + signing model. The signatures are JWS (RFC 7515) over the canonicalized (JCS) card, and v1.0 dropped the top-level `url` for `supported_interfaces[].url`; so any url rewrite invalidates the signature, which is the real tension with how we'd serve federated cards. But the way out is the same `tenant` field from (8): route by tenant and serve the upstream's signed card verbatim, no rewrite, signature intact. So your two points are really one answer... `tenant` is both the policy discriminator and what lets us keep signed cards valid under multi-agent hosting. It reframes the catalog from "rewrite each card" to "advertise gateway-url + tenant per agent, serve the signed cards untouched."

And (9) : agreed, that's where I was leaning too.., I'll do the A2A exploration in a fork so I can move fast and not carry MCP-regression risk while spiking, with the intent to upstream it back into mcp-gateway incrementally so MCP and A2A still share the one policy plane.., workshop in the fork, home in-tree. And you're right the broker's much thinner for A2A (card cache, no aggregation or sessions)... I'll stop the doc overstating that reuse.

I'll get a revised pass up against all of this; the tenant ↔ signed-card piece in particular feels like it simplifies the federation story a fair bit, so I'd genuinely like your read on whether routing-by-tenant changes how you'd want discovery to look. Once that revision's up, I'd love to land this as the agreed baseline so I can start the implementation against something stable rather than a moving target. And separately, the test server (#1200) is independent of this discussion and gives the e2e work a real target to run against.., it'll track the v1.0 surface as that firms up, so if you're good with it I'd like to get that one in too. Thanks again for going this deep on it.

carlesarnal:
Nice design doc. One thought on the agent card storage layer.

The design uses an in-memory broker cache for agent cards (poll + conditional GET from upstream). This works for the PoC, but for production multi-gateway deployments you'd likely want a shared metadata store so that all gateway replicas see the same set of agents without each one polling independently.

Apicurio Registry already supports `AGENT_CARD` as a native artifact type (aligned with A2A v1.0), with versioning, compatibility rules, and three-tier discovery endpoints (public/entitled/private). It could serve as that shared backend — the broker would read from the registry instead of polling each upstream directly, and operators could publish/manage agent cards through the registry's API alongside their schema governance.

Not suggesting this for the PoC scope, but worth considering as a pluggable backend option in the design so it doesn't get locked out. Happy to discuss if useful.

Aman-Cool:
Thanks @carlesarnal, really useful pointer... I read through #7887 and the `a2a` package, and there's a lot here worth not locking out.

On the core point you're right : the in-memory poll cache is a PoC choice, and the card store is exactly the layer I'd rather keep behind an interface than bake in. So I'll make storage a pluggable backend in the design; in-memory for the steel thread, a shared store as a first-class option.., so a registry like Apicurio could slot in without re-architecting. It also converges with a point @david-martin made in review, that the A2A broker is essentially a thin card cache, so keeping that layer pluggable rather than fixed feels clearly right.

A couple of things stood out while reading. #7887 lines up almost exactly with where I'm taking the design... I'm targeting A2A v1.0, and you've already got `supportedInterfaces`, the named `securitySchemes` map, `capabilities.extendedAgentCard`, and `AgentCardSignature`/JWS; so a field mapping would be clean, and I'll adopt the `/.well-known/a2a` path for the v1.0 surface. The other thing I noticed is you model `MCP_TOOL` alongside `AGENT_CARD` ; since the gateway routes both MCP and A2A, that symmetry's interesting... in principle the same governed source could back both halves, though that's well beyond anything I'd scope now.

One thing I genuinely don't have an answer to, where you'd know far better than me.., how does signing work if the registry is the card source? v1.0 cards are JWS-signed over the canonicalized card, and the way I'd planned to serve federated cards without breaking that is to route by `tenant` and pass the agent's signed card through verbatim. If cards live in Apicurio instead, does the registry preserve the upstream agent's signature as-is, or re-sign on publish with its own key; i.e. does the registry become the trust authority for the card? That changes who the client is actually verifying, and it's the bit I'd want to understand before leaning on a registry as a source. Curious how you've been thinking about it... happy to dig into the mapping whenever suits.

Aman-Cool:
@david-martin, The revised pass is up. Everything from your review is now reflected in the doc itself : v1.0 as the target with the version-specific surface isolated, `Ready` = configured for status, cross-namespace `targetRef` allowed, oldest-wins dropped, skill-filtering marked not applicable to A2A, the tenant-as-header enforcement correction, and the v1.0 signed-card / multi-tenancy findings written into the card integrity section.

Two things are flagged `[OPEN]` inline and are the only decisions left before this can settle:

1. **v1.0 target** : I've written the doc as moving to v1.0 per your comment, so unless you object I'll treat that as resolved and migrate the remaining v0.3 references in the body as part of the same pass.
2. **Prefix collision** : dropping oldest-wins leaves namespace-qualified paths (`/a2a/{namespace}/{prefix}`) as my recommendation; collisions become structurally impossible rather than timing-dependent. It changes the path form everywhere in the doc, so I'd rather get your yes/no before rewriting all the examples.

Once those two land I'd like to merge this as the agreed baseline so the fork work builds against something stable.

cc @maleck13 and @jasonmadigan

Aman-Cool:
@david-martin, Quick follow-up on (2) from the review... the mid-request mode change is now derisked against real Envoy, not just the elicitation precedent, and it surfaced one constraint the doc should record.

I ran it as the first spike in the fork ([[fork PR](https://github.com/Aman-Cool/mcp-gateway/pull/1)](https://github.com/Aman-Cool/mcp-gateway/pull/1), transcripts there): router detects `/a2a/{prefix}` at request-headers, reads the method at request-body, and at response-headers flips `ModeOverride` per method; `BUFFERED` for `message/send`/`tasks/get`, `STREAMED` for `message/stream`/`tasks/resubscribe`. Envoy (Istio 1.27, `allow_mode_override: true`) honors both directions: the buffered path delivered the full response body in one message and the client got the rewritten gateway task ID back, and the streamed path carried SSE events through incrementally.., same code path, method-keyed. `/mcp` stayed `200` throughout.

The finding worth recording: the first buffered run failed `500`, and Envoy's `ext_proc` debug named it; "mismatch between content length and the length of the mutated body." The upstream's `Content-Length` is committed at the headers phase, so a buffered rewrite that changes body length (any task-ID rewrite will) trips fail-closed Envoy... the fix is stripping `Content-Length` in the same response-headers response that sets the override, so Envoy re-frames the response. MCP never hits this because tool-call responses are SSE; no `Content-Length`.., so it's specific to the buffered JSON path. I'll add a line to the doc's `message/send` rewrite section so it's captured where the next implementer will look.

So the piece you flagged as "haven't seen it done before" works, with one known constraint and a one-line fix. The fork workflow from (9) is earning its keep already 🙂

carlesarnal:
Good question on signing. Today the registry stores agent cards as opaque content — when you publish a card with JWS signatures, it preserves them byte-for-byte. The `/.well-known/a2a` and search endpoints serve the stored content as-is, so a signed card published to Apicurio comes back with the original signature intact. The registry validates the JSON structure of signatures on ingest (required `protected` + `signature` fields per JWS) but does not verify or re-sign — it's a metadata store, not a trust authority.

So for the gateway integration, the model would be: the agent (or its operator) signs the card and publishes it to Apicurio. The gateway reads it from the registry and serves it verbatim. The client verifies against the agent's key, not the registry's. The registry's role is versioned storage + discovery + governance (compatibility rules, visibility controls), not identity.

That said, we don't have a signing/verification feature on the roadmap today — it's deliberately out of scope (#7134 was closed for this reason). If v1.0 signed cards become common and there's demand for the registry to verify signatures on ingest or to act as a re-signing proxy, that's something we could revisit, but the current design assumes signing is between the agent and its consumers.

This actually aligns well with your `tenant`-based routing approach — the gateway routes by tenant, serves the upstream's signed card through untouched, and the registry is the shared store that all gateway replicas read from instead of each one polling independently.

david-martin (approved):
Approving, with phase 1 changes being implementable

The intent is to have a `--enable-a2a` flag on the router, and only then parse traffic as a2a if it's on a /a2a prefixed path. No CRDs in this phase. Enables auth, auditing and observability at this phase. No a2a aggregation of cards or a2a catalog endpoint at this time.

PR #1133, opened 2026-06-12
Fix tls hairpin

### LINKED ISSUE #1130: HTTPS hairpin init fails due to TLS hostname mismatch
## Problem

The broker's hairpin initialize request fails for all HTTPS Gateway listeners due to TLS hostname verification.

The hairpin flow (`clients.Initialize` in `internal/clients/clients.go`) connects to the Gateway's internal service address (e.g. `mcp-gateway-istio.ns.svc.cluster.local:443`), but the Gateway's TLS certificate has SANs for the public hostname (e.g. `mcp.example.com`). Go's TLS verifier rejects the mismatch:

```
x509: certificate is valid for mcp.example.com, not mcp-gateway-istio.gateway-system.svc.cluster.local
```

This affects every HTTPS listener regardless of whether the CA is public or private.

For private CAs, the default system root pool also lacks the CA certificate, producing:

```
x509: certificate signed by unknown authority
```

## Context

- `derivePrivateHost` already prepends `https://` for HTTPS listeners (issue #917)
- `mcp-go` v0.53.0 exposes `transport.WithHTTPBasicClient(client *http.Client)` to inject a custom HTTP client
- The public hostname is already available via `--mcp-gateway-public-host`

### 1. Fix public CA: set `ServerName` on the hairpin TLS config

Build an `*http.Client` with `tls.Config{ServerName: publicHost}` when the private host uses HTTPS. This tells Go to verify the cert's SANs against the public hostname while the TCP connection goes to the internal address.

The broker-router already has `--mcp-gateway-public-host` available. At startup, build the hairpin client:

```go
func buildHairpinHTTPClient(privateHost, publicHost, caCertPath string) (*http.Client, error) {
	// only needed when the hairpin target is HTTPS
	if !strings.HasPrefix(strings.ToLower(privateHost), "https://") {
		return nil, nil
	}

	pool, err := x509.SystemCertPool()
	if err != nil {
		pool = x509.NewCertPool()
	}

	// optional: private CA support
	if caCertPath != "" {
		pem, err := os.ReadFile(caCertPath)
		if err != nil {
			return nil, fmt.Errorf("failed to read gateway CA cert: %w", err)
		}
		if !pool.AppendCertsFromPEM(pem) {
			return nil, fmt.Errorf("failed to parse gateway CA cert PEM")
		}
	}

	t := http.DefaultTransport.(*http.Transport).Clone()
	t.TLSClientConfig = &tls.Config{
		MinVersion: tls.VersionTLS12,
		RootCAs:    pool,
		ServerName: publicHost, // verify SANs against the public hostname
	}
	return &http.Client{Transport: t}, nil
}
```

`ServerName: publicHost` is the key fix — the TCP connection goes to `mcp-gateway-istio.ns.svc.cluster.local:443` (the internal address), but Go's TLS verifier checks the cert's SANs against `mcp.example.com` (the public hostname the cert was issued for). Without this, the handshake fails because the internal DNS name won't appear in the cert's SAN list.

### 2. Fix private CA: add `--gateway-ca-cert` flag

A new CLI flag `--gateway-ca-cert=/path/to/ca.crt` on the broker-router binary. When set, the PEM is read at startup and appended to the system root pool on the hairpin HTTP client (shown above).

The operator does not automatically mount this certificate. For now, users using a private CA must manually patch the broker-router deployment to mount the CA certificate and add the `--gateway-ca-cert` flag. The operator must preserve user-applied volume mounts and args when reconciling the deployment — it must not overwrite manual patches to these fields.

### 3. Thread the client through `clients.Initialize`

`clients.Initialize` gains an `*http.Client` parameter (nil = use default, preserving plain HTTP behavior). It passes the client via `transport.WithHTTPBasicClient()`. The `InitForClient` function type and the router's call site in `initializeMCPSeverSession` are updated to pass it through.

## What stays the same

- Plain HTTP listeners — no custom client, existing behavior preserved
- Broker-to-upstream CA trust (`caCertSecretRef`) — separate path, unchanged
- The hairpin JWT validation flow — unchanged

## Affected files

- `cmd/mcp-broker-router/main.go` — new `--gateway-ca-cert` flag, build hairpin client at startup
- `internal/clients/clients.go` — accept and use custom `*http.Client`
- `internal/mcp-router/server.go` / `request_handlers.go` — thread the client through `InitForClient`

maleck13:
Fix: #1130  Set ServerName on the hairpin HTTP client's TLS config so the cert is verified against the public hostname.
  Add --gateway-ca-cert flag for private CA support. Update the operator to preserve user-applied volume mounts
  across reconciliations.

>Note we didn't add a new property to the MCPGatewayExtension resource here,  we should consider if we need that in a follow up.

>Note there are a lot of files changed but this is largely due to the hostname change in the e2e tests

If the gateway is using a well known CA this flag is not needed, so it is only in the case of a private CA being used to protect the gateway listener

This change also sets the main test listener as a HTTPS listener with a custom cert so we are always testing this change now

## Summary by CodeRabbit

* **New Features**
  * HTTPS Gateway listener on port 8443 with TLS termination; NodePort and hostname updates to *.mcp-gateway.local
  * New --gateway-ca-cert flag and support for supplying private CA bundles
  * Debug make target to enable verbose controller/gateway logging

* **Improvements**
  * CI/setup ordering and retry logic adjusted to ensure TLS/cert prerequisites are ready
  * Deployment reconciliation now preserves user-added volumes/mounts while updating managed entries

* **Tests**
  * E2E tests and helpers updated for HTTPS hairpin and CA-mounted scenarios

* **Documentation**
  * Guide added for HTTPS gateway and private CA configuration


tests/e2e/happy_path_test.go
```diff
@@ -1161,6 +1161,40 @@ var _ = Describe("MCP Gateway Registration Happy Path", func() {
 		}, TestTimeoutLong, TestRetryInterval).To(Succeed())
 	})
 
+	It("[Happy] should report prompt conflicts in MCPServerRegistration status when same prefix is used", func() {
```
maleck13: This exists on main but rebasing and changing the hostnames causes it to show up here

tests/e2e/mcp_client.go
```diff
@@ -18,6 +19,27 @@ import (
...
+	return &http.Client{
+		Transport: &http.Transport{
+			TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
+			DialContext: func(ctx context.Context, network, addr string) (net.Conn, error) {
+				host, port, _ := net.SplitHostPort(addr)
+				if strings.HasSuffix(host, ".local") {
+					addr = net.JoinHostPort("127.0.0.1", port)
+				}
+				return (&net.Dialer{}).DialContext(ctx, network, addr)
+			},
```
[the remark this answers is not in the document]
maleck13: Test code only

cmd/mcp-broker-router/main.go
```diff
@@ -245,6 +251,21 @@ func (a *app) setupSessionInfra(ctx context.Context) {
...
+	hc, err := clients.BuildHairpinHTTPClient(
+		a.brokerCfg.privateHost,
+		a.brokerCfg.publicHost,
+		a.brokerCfg.gatewayCACert,
+	)
+	if err != nil {
+		panic("failed to build hairpin HTTP client: " + err.Error())
+	}
+	a.hairpinClient = hc
+	if hc != nil {
```
david-martin: redundant?

Patryk-Stefanski: nit: No need for a nil check here

internal/clients/clients.go
```diff
@@ -15,6 +20,8 @@ import (
 	"github.com/mark3labs/mcp-go/mcp"
 )
 
+const hairpinTimeout = 5 * time.Second
```
jasonmadigan: the old path, from what I can tell, used mcp-go's client with no timeout, so this is a bit of a change. is this a bit small? should it be configurable?

config/istio/gateway/gateway.yaml
```diff
@@ -73,3 +73,15 @@ spec:
...
           from: All
+    - name: mcp-tls
+      hostname: '*.mcp-gateway.local'
+      port: 8443
+      protocol: HTTPS
+      tls:
+        mode: Terminate
+        certificateRefs:
+        - kind: Secret
+          name: mcp-gateway-tls-cert
```
jasonmadigan: what creates this?

internal/clients/clients.go
```diff
@@ -71,3 +81,36 @@ func Initialize(ctx context.Context, gatewayHost, initToken string, conf *config
...
+		if !pool.AppendCertsFromPEM(pem) {
+			return nil, fmt.Errorf("failed to parse gateway CA cert PEM")
+		}
+	}
+
+	t := http.DefaultTransport.(*http.Transport).Clone()
+	t.TLSClientConfig = &tls.Config{
+		MinVersion: tls.VersionTLS12,
+		RootCAs:    pool,
+		ServerName: publicHost,
```
jasonmadigan: 🤖  flagged (seems legit):

ServerName: publicHost — TLS cert checks compare against a hostname, never host:port. The operator-derived path strips the port, but a hand-set --mcp-gateway-public-host (which the docs in this PR tell users to do) isn't validated, so a value like mcp.example.com:8443 flows straight into ServerName and every hairpin fails with a confusing cert error. Strip via net.SplitHostPort here or reject it at flag parse.

config/istio/gateway/gateway.yaml
maleck13: The test setup sets up the secret

internal/clients/clients.go
maleck13: yes good spot! this opens a long lived connection. I had missed that 👍

config/istio/gateway/gateway.yaml
maleck13: will move it out of the test setup and into normal setup. Actually will move it out of this gateway and move it to the test setup altogether

maleck13:
@jasonmadigan fixes in the latest commit for your review findings 👍

jasonmadigan (approved):
lgtm. there was an e2e failure before will keep an eye

PR #1137, opened 2026-06-13
fix: preserve split-form (--flag value) user-managed broker-router co…

aniketpandey05:
## What does this PR do?

Fixes the broker-router command reconciliation logic to correctly handle user-managedflags written in split form (`--flag value`) instead of only `--flag=value`.Previously `mergeCommand` dropped the value token of such flags on reconcile, and
`filterManagedFlags` could report false drift for the orphaned value token.

## Details

`internal/controller/broker_router.go`:
- Added `commandEntry` / `parseCommandEntries` to split a container command into logical entries (binary name, `--flag=value`, `--flag value` pair, or bare `--flag`).
- Rewrote `filterManagedFlags` and `mergeCommand` to operate on these logical entries instead of raw argv tokens, so split-form flags and their valuesare preserved/compared as a unit.
- Switched managed-flag matching from `strings.HasPrefix` to an exact matchon the flag name (extracted before `=`), avoiding accidental prefixcollisions. All existing managed flags are not prefixes of one another, soexisting behavior is unchanged.

`internal/controller/deployment_test.go`:
- `TestFilterManagedFlags`: added cases for a user split-form flag (stripped)and a managed split-form flag (kept as a unit).
- `TestMergeCommand`: added cases for preserving user split-form flags, mixedmanaged `=`-form + user split-form flags, and a managed split-form flagbeing replaced by the desired `=`-form value.
- `TestDeploymentNeedsUpdate`: added a no-op case for a user-managedsplit-form flag (previously reported false drift).

## Summary by CodeRabbit

* **Refactor**
  * Improved command-parsing to treat command tokens as flag entries, correctly handling both `--flag=value` and split `--flag` `value` forms; preserves binary name and groups flag/value pairs.

* **Tests**
  * Added coverage ensuring user split-form flags are preserved/ignored as intended and managed flags are merged or replaced correctly across mixed scenarios.

### DIFF
--- a/internal/controller/broker_router.go
+++ b/internal/controller/broker_router.go
@@ -534,14 +534,60 @@ func deploymentNeedsUpdate(desired, existing *appsv1.Deployment) (bool, string) 
 	return false, ""
 }
 
-// filterManagedFlags returns only the binary name and flags the controller manages.
+// commandEntry represents a single logical entry in a container command:
+// either the binary name, a "--flag=value" token, or a "--flag value" pair
+// spanning two tokens.
+type commandEntry struct {
+	tokens   []string
+	flagName string
+}
+
+// extractFlagName extracts the flag name from a "--flag" or "--flag=value" token.
+func extractFlagName(arg string) string {
+	if idx := strings.Index(arg, "="); idx != -1 {
+		return arg[:idx]
+	}
+	return arg
+}
+
+// isManagedFlag reports whether name (e.g. "--log-level") is a controller-managed flag.
+func isManagedFlag(name string) bool {
+	return slices.Contains(managedCommandFlags, name)
+}
+
+// parseCommandEntries splits a container command into logical entries,
+// grouping "--flag value" split-form flags into a single entry with their
+// value token.
+func parseCommandEntries(command []string) []commandEntry {
+	var entries []commandEntry
+	for i := 0; i < len(command); i++ {
+		arg := command[i]
+		if !strings.HasPrefix(arg, "--") {
+			entries = append(entries, commandEntry{tokens: []string{arg}})
+			continue
+		}
+		name := extractFlagName(arg)
+		if strings.Contains(arg, "=") {
+			entries = append(entries, commandEntry{tokens: []string{arg}, flagName: name})
+			continue
+		}
+		if i+1 < len(command) && !strings.HasPrefix(command[i+1], "--") {
+			entries = append(entries, commandEntry{tokens: []string{arg, command[i+1]}, flagName: name})
+			i++
+			continue
+		}
+		entries = append(entries, commandEntry{tokens: []string{arg}, flagName: name})
+	}
+	return entries
+}
+
+// filterManagedFlags returns the binary name and the controller-managed flags
+// (with their values) as a flat token slice.
 func filterManagedFlags(command []string) []string {
 	var out []string
-	for _, arg := range command {
-		if !strings.HasPrefix(arg, "--") || slices.ContainsFunc(managedCommandFlags, func(flag string) bool {
-			return strings.HasPrefix(arg, flag)
-		}) {
-			out = append(out, arg)
+	for _, entry := range parseCommandEntries(command) {
+		if entry.flagName == "" || isManagedFlag(entry.flagName) {
+			out = append(out, entry.tokens...)
 		}
 	}
 	return out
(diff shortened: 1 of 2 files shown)


aniketpandey05:
@coderabbitai review

Patryk-Stefanski (commented):
The underlying bugs here are real and worth fixing — the prefix-match collision and the dropped split-form value token are genuine. Thanks for tackling this.

That said, I think the tokenizer is more machinery than the problem requires. The controller's own `buildBrokerRouterDeployment` always emits `--flag=value` form (the only exception is the bare `--enable-url-elicitation`), so the split-form parser only ever activates on user-added flags in the _existing_ deployment. The tokenizer also has an edge case: values that themselves start with `--` (e.g. `--my-flag --some-value`) are rejected by the `!strings.HasPrefix(command[i+1], "--")` lookahead, silently splitting the pair into two separate entries.

The simpler approach that fixes both original bugs without the new one:

```go
func extractFlagName(arg string) string {
    if i := strings.Index(arg, "="); i != -1 {
        return arg[:i]
    }
    return arg
}
```

Then replace the `strings.HasPrefix(arg, flag)` calls in `filterManagedFlags` and `mergeCommand` with `slices.Contains(managedCommandFlags, extractFlagName(arg))`. This fixes the prefix-collision bug exactly and is immune to the `--`-prefixed value edge case.

The trade-off: a user split-form `--flag value` pair would still result in only `--flag` being dropped and the orphaned value token being kept as a spurious user token — but that's a narrow edge case (split-form flags in a Kubernetes container command are uncommon) and the simpler code is easier to reason about.

Happy to discuss if there's a strong reason to handle that case too.

aniketpandey05:
@Patryk-Stefanski I think your fix is better as it resolves both real bugs (the prefix-collision and the dropped split-form value in `filterManagedFlags`/`mergeCommand`), and it's easier to  review 

One question  should we also update `filterManagedFlags` to compare by exact flag name/index instead of the current `!strings.HasPrefix(arg, "--")` check? That check currently lets any non-flag token through, including a stray value from a split-form flag, which could make the comparison see "drift" that isn't really there and trigger unnecessary updates on every reconcile.

thank you for revieiwing the pr

internal/controller/broker_router.go
```diff
@@ -534,33 +534,77 @@ func deploymentNeedsUpdate(desired, existing *appsv1.Deployment) (bool, string)
...
+// either the binary name, a "--flag=value" token, or a "--flag value" pair
+// spanning two tokens.
+type commandEntry struct {
+	tokens   []string
+	flagName string
+}
+
+// flagName extracts the flag name from a "--flag" or "--flag=value" token.
+func flagName(arg string) string {
+	if idx := strings.Index(arg, "="); idx != -1 {
```
Patryk-Stefanski, line 547: The function name `flagName` collides with the struct field `commandEntry.flagName` — both live in the same file and share the identifier. Consider renaming to `extractFlagName` or `parseFlagName` to make them distinct at a glance.

Patryk-Stefanski (commented):
One rename request before merge, and an answer to your question below.

**Rename  function → ** (or similar)

The package-level function  and the struct field  share the same identifier. Go resolves them correctly by context, but a reader scanning the struct sees  and the function named  in the same file and has to stop to distinguish them. Something like  or  removes the ambiguity for no cost.

---

**Your question about `filterManagedFlags`**

> Should we also update `filterManagedFlags` to compare by exact flag name/index instead of the current `!strings.HasPrefix(arg, "--")` check? That check currently lets any non-flag token through, including a stray value from a split-form flag, which could make the comparison see "drift" that isn't really there.

That's exactly the bug this PR fixes — and the fix is already here. The old token-at-a-time loop admitted stray value tokens (like `"3600"` from `--session-length 3600`) because they don't start with `--`, so `!strings.HasPrefix(arg, "--")` passed them through. The new `parseCommandEntries` groups the flag and its value into a single `commandEntry`, so a stray value can never appear as a standalone entry. The new `filterManagedFlags` then only needs to check `entry.flagName == ""` (binary/positional) and `isManagedFlag(entry.flagName)`. No further change needed — the PR is the answer to your own question.

aniketpandey05:
@Patryk-Stefanski I have renamed the package-level helper function to `extractFlagName` to eliminate any ambiguity with the `commandEntry.flagName` struct field.

Thanks for the clarification on `filterManagedFlags`

PR #1138, opened 2026-06-13
fix: restore automountServiceAccountToken=false on broker-router deployment reconcile

aniketpandey05:
Fixes broker-router Deployment reconciliation to compare and restore thepod-level `automountServiceAccountToken` field, which the controller intendsto keep `false` for security but previously did not enforce on existingDeployments.

## Details

`internal/controller/broker_router.go`:
- `deploymentNeedsUpdate`: added a comparison of`Spec.Template.Spec.AutomountServiceAccountToken` between desired and
  existing, right after the image check. Drift is reported with the old/new values.
- `reconcileBrokerRouter` update path: added`existingDeployment.Spec.Template.Spec.AutomountServiceAccountToken deployment.Spec.Template.Spec.AutomountServiceAccountToken` alongside the other container/pod field patches.


aniketpandey05:
@coderabbitai help

PR #1139, opened 2026-06-14
fix: honor MCPServerRegistration targetRef namespace

### LINKED ISSUE #1199: MCPServerRegistration: targetRef.namespace is ignored during HTTPRoute resolution
<h3 style="font-style: normal; font-variant-caps: normal; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-line: none; text-decoration-thickness: auto; text-decoration-style: solid; caret-color: rgb(255, 255, 255); color: rgb(255, 255, 255);">Summary</h3><p style="font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-line: none; text-decoration-thickness: auto; text-decoration-style: solid; caret-color: rgb(255, 255, 255); color: rgb(255, 255, 255);"><code>MCPServerRegistrationSpec.targetRef.namespace</code><span class="Apple-converted-space"> </span>is documented as functional ("optional, defaults to same namespace"), but<span class="Apple-converted-space"> </span><code>getTargetHTTPRoute</code><span class="Apple-converted-space"> </span>— the resolver used by<span class="Apple-converted-space"> </span><code>Reconcile</code><span class="Apple-converted-space"> </span>— ignores it and always looks in the registration's own namespace. Two sibling paths do honor it, so the field is half-implemented.</p><pre style="font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-line: none; text-decoration-thickness: auto; text-decoration-style: solid; caret-color: rgb(255, 255, 255); color: rgb(255, 255, 255);"><code class="language-go">// mcpserverregistration_controller.go:307 — ignores targetRef.namespace
namespaceName := types.NamespacedName{Namespace: mcpsr.Namespace, Name: mcpsr.Spec.TargetRef.Name}
</code></pre>
Path | File:line | Honors targetRef.namespace?
-- | -- | --
Route resolution (Reconcile) | controller.go:307 | No
Status write-back | controller.go:605-607 | Yes
Watch index | controller.go:767-769 | Yes

<h3 style="font-style: normal; font-variant-caps: normal; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-line: none; text-decoration-thickness: auto; text-decoration-style: solid; caret-color: rgb(255, 255, 255); color: rgb(255, 255, 255);">Impact</h3><p style="font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-line: none; text-decoration-thickness: auto; text-decoration-style: solid; caret-color: rgb(255, 255, 255); color: rgb(255, 255, 255);">A cross-namespace<span class="Apple-converted-space"> </span><code>targetRef</code><span class="Apple-converted-space"> </span>either fails with<span class="Apple-converted-space"> </span><code>failed to get targeted httproute</code>, or — if a same-named route exists in the registration's own namespace — silently binds to the wrong route/backend with no error.</p><h3 style="font-style: normal; font-variant-caps: normal; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-line: none; text-decoration-thickness: auto; text-decoration-style: solid; caret-color: rgb(255, 255, 255); color: rgb(255, 255, 255);">Reproduce</h3><ol style="font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-line: none; text-decoration-thickness: auto; text-decoration-style: solid; caret-color: rgb(255, 255, 255); color: rgb(255, 255, 255);"><li>HTTPRoute<span class="Apple-converted-space"> </span><code>weather-route</code><span class="Apple-converted-space"> </span>in<span class="Apple-converted-space"> </span><code>team-a</code>.</li><li><code>MCPServerRegistration</code><span class="Apple-converted-space"> </span>in<span class="Apple-converted-space"> </span><code>team-b</code><span class="Apple-converted-space"> </span>with<span class="Apple-converted-space"> </span><code>targetRef: {name: weather-route, namespace: team-a}</code>.</li><li>Reconcile looks up the route in<span class="Apple-converted-space"> </span><code>team-b</code><span class="Apple-converted-space"> </span>→<span class="Apple-converted-space"> </span><code>NotFound</code><span class="Apple-converted-space"> </span>→<span class="Apple-converted-space"> </span><code>Ready=False</code>, despite a correct<span class="Apple-converted-space"> </span><code>targetRef.namespace</code>.</li></ol><p style="font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-line: none; text-decoration-thickness: auto; text-decoration-style: solid; caret-color: rgb(255, 255, 255); color: rgb(255, 255, 255);">No test covers cross-namespace registration<span class="Apple-converted-space"> </span><code>targetRef</code><span class="Apple-converted-space"> </span>(existing cross-ns tests cover only<span class="Apple-converted-space"> </span><code>MCPGatewayExtension</code>→<code>Gateway</code>).</p><h3 style="font-style: normal; font-variant-caps: normal; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-line: none; text-decoration-thickness: auto; text-decoration-style: solid; caret-color: rgb(255, 255, 255); color: rgb(255, 255, 255);">Proposed fix</h3><ul style="font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-line: none; text-decoration-thickness: auto; text-decoration-style: solid; caret-color: rgb(255, 255, 255); color: rgb(255, 255, 255);"><li><strong>Option A (minimal):</strong><span class="Apple-converted-space"> </span>treat<span class="Apple-converted-space"> </span><code>targetRef</code><span class="Apple-converted-space"> </span>as same-namespace-only — make all paths use<span class="Apple-converted-space"> </span><code>mcpsr.Namespace</code>, reject non-empty<span class="Apple-converted-space"> </span><code>targetRef.namespace</code><span class="Apple-converted-space"> </span>via CEL.</li><li><strong>Option B (full support):</strong><span class="Apple-converted-space"> </span>honor<span class="Apple-converted-space"> </span><code>targetRef.namespace</code><span class="Apple-converted-space"> </span>in<span class="Apple-converted-space"> </span><code>getTargetHTTPRoute</code>, gated by a<span class="Apple-converted-space"> </span><code>ReferenceGrant</code><span class="Apple-converted-space"> </span>check reusing the existing<span class="Apple-converted-space"> </span><code>HasValidReferenceGrant</code><span class="Apple-converted-space"> </span>pattern (<code>mcpgatewayextension.go:24-77</code>).</li></ul><p style="font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-line: none; text-decoration-thickness: auto; text-decoration-style: solid; caret-color: rgb(255, 255, 255); color: rgb(255, 255, 255);">Recommend A unless cross-namespace agent routes are a real requirement.</p><h3 style="font-style: normal; font-variant-caps: normal; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-line: none; text-decoration-thickness: auto; text-decoration-style: solid; caret-color: rgb(255, 255, 255); color: rgb(255, 255, 255);">Why now</h3><p style="font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-line: none; text-decoration-thickness: auto; text-decoration-style: solid; caret-color: rgb(255, 255, 255); color: rgb(255, 255, 255);">The A2A controller (#1114) reuses<span class="Apple-converted-space"> </span><code>TargetReference</code><span class="Apple-converted-space"> </span>+<span class="Apple-converted-space"> </span><code>getTargetHTTPRoute</code>, so this would propagate into a second CRD. Fixing it in MCP first keeps the cross-namespace contract consistent.</p><h3 style="font-style: normal; font-variant-caps: normal; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-line: none; text-decoration-thickness: auto; text-decoration-style: solid; caret-color: rgb(255, 255, 255); color: rgb(255, 255, 255);">Acceptance criteria</h3><ul style="font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-line: none; text-decoration-thickness: auto; text-decoration-style: solid; caret-color: rgb(255, 255, 255); color: rgb(255, 255, 255);"><li>[ ] All<span class="Apple-converted-space"> </span><code>targetRef</code><span class="Apple-converted-space"> </span>paths agree on namespace handling.</li><li>[ ] Behavior matches the CRD docs (or docs updated).</li><li>[ ] Test for cross-namespace<span class="Apple-converted-space"> </span><code>targetRef</code>, including the same-name mis-binding case.</li></ul>

avinxshKD:
Fixes MCPServerRegistration reconciliation when `spec.targetRef.namespace` points at an HTTPRoute outside the registration namespace.

The API and docs allow this, and the field index/status update paths already account for it, but `getTargetHTTPRoute` always read the route from the MCPServerRegistration namespace.

This updates route lookup to use `targetRef.namespace` when set, with a regression test for cross-namespace refs.

Fixes: #1199

Tested:
- go test ./internal/controller

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed cross-namespace HTTPRoute handling for MCPServerRegistration by correctly honoring the target reference namespace (with default fallback).
  * Enforced Gateway API ReferenceGrant permissions for both route resolution and HTTPRoute status updates, preventing unauthorized cross-namespace cleanup/config removal.
  * Updated reconciliation to re-process affected MCPServerRegistrations when ReferenceGrants change.

* **Tests**
  * Added/extended unit tests for target-namespace resolution, ReferenceGrant-required error behavior, status cleanup rules, and ReferenceGrant-driven reconcile triggers.

### DIFF
--- a/internal/controller/mcpserverregistration_controller.go
+++ b/internal/controller/mcpserverregistration_controller.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"crypto/x509"
 	"encoding/pem"
+	"errors"
 	"fmt"
 	"net"
 	"net/url"
@@ -24,6 +25,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 	gatewayv1 "sigs.k8s.io/gateway-api/apis/v1"
+	gatewayv1beta1 "sigs.k8s.io/gateway-api/apis/v1beta1"
 
 	mcpv1alpha1 "github.com/Kuadrant/mcp-gateway/api/v1alpha1"
 	"github.com/Kuadrant/mcp-gateway/internal/config"
@@ -41,6 +43,8 @@ const (
 	maxCACertSize = 64 * 1024
 	// HTTPRouteIndex used to find MCPServerRegistrations
 	HTTPRouteIndex = "spec.targetRef.httproute"
+	// MCPServerRegistrationReferenceGrantIndex used to find MCPServerRegistrations for ReferenceGrants
+	MCPServerRegistrationReferenceGrantIndex = "spec.targetRef.referencegrant"
 	// ProgrammedHTTPRouteIndex used to find programmed httproutes
 	ProgrammedHTTPRouteIndex = "status.hasProgrammedCondition"
 
@@ -51,6 +55,8 @@ const (
 	// conditionReasonDisabled is the reason used when the MCPServerRegistration is disabled
 	conditionReasonDisabled = "Disabled"
 )
+
+var errReferenceGrantRequired = errors.New("ReferenceGrant required")
 
 // ServerInfo holds server information
 type ServerInfo struct {
@@ -83,6 +89,7 @@ type MCPReconciler struct {
 // +kubebuilder:rbac:groups=mcp.kuadrant.io,resources=mcpserverregistrations/status,verbs=get;update
 // +kubebuilder:rbac:groups=gateway.networking.k8s.io,resources=httproutes,verbs=get;list;watch
 // +kubebuilder:rbac:groups=gateway.networking.k8s.io,resources=httproutes/status,verbs=update
+// +kubebuilder:rbac:groups=gateway.networking.k8s.io,resources=referencegrants,verbs=list;watch
 // +kubebuilder:rbac:groups="",resources=secrets,verbs=get;list;watch;create;update;delete
 // +kubebuilder:rbac:groups="",resources=services,verbs=get
 
@@ -140,7 +147,14 @@ func (r *MCPReconciler) Reconcile(ctx context.Context, req reconcile.Request) (r
 	// get the HTTPRoute and gateway(s) this MCPServerRegistration targets
 	targetRoute, err := r.getTargetHTTPRoute(ctx, mcpsr)
 	if err != nil {
-		if err := r.updateStatus(ctx, mcpsr, false, conditionReasonNotReady, err.Error()); err != nil {
+		statusReason := conditionReasonNotReady
+		if errors.Is(err, errReferenceGrantRequired) {
+			statusReason = mcpv1alpha1.ConditionReasonRefGrantRequired
+			if removeErr := r.ConfigReaderWriter.RemoveMCPServer(ctx, mcpServerName(mcpsr)); removeErr != nil {
+				return ctrl.Result{}, removeErr
+			}
+		}
+		if err := r.updateStatus(ctx, mcpsr, false, statusReason, err.Error()); err != nil {
 			if apierrors.IsConflict(err) {
 				// don't log these as they are just noise
 				return ctrl.Result{RequeueAfter: defaultRequeueTime}, nil
@@ -316,7 +330,21 @@ func (r *MCPReconciler) findValidGatewaysForMCPServer(ctx context.Context, targe
 }
 
 func (r *MCPReconciler) getTargetHTTPRoute(ctx context.Context, mcpsr *mcpv1alpha1.MCPServerRegistration) (*gatewayv1.HTTPRoute, error) {
-	namespaceName := types.NamespacedName{Namespace: mcpsr.Namespace, Name: mcpsr.Spec.TargetRef.Name}
+	targetNamespace := targetRefNamespace(mcpsr.Namespace, mcpsr.Spec.TargetRef)
+	if targetNamespace != mcpsr.Namespace {
+		hasGrant, err := r.hasValidHTTPRouteReferenceGrant(ctx, mcpsr, targetNamespace)
+		if err != nil {
+			return nil, err
+		}
+		if !hasGrant {
+			return nil, fmt.Errorf("%w in %s to allow cross-namespace HTTPRoute reference from %s", errReferenceGrantRequired, targetNamespace, mcpsr.Namespace)
+		}
+	}
+
+	namespaceName := types.NamespacedName{
+		Namespace: targetNamespace,
+		Name:      mcpsr.Spec.TargetRef.Name,
+	}
 	logger := logf.FromContext(ctx).WithValues("method", "getTargetHTTPRoute")
 	logger.V(1).Info("httproute target ", "namespacename ", namespaceName)
 	targetRoute := &gatewayv1.HTTPRoute{}
@@ -325,6 +353,43 @@ func (r *MCPReconciler) getTargetHTTPRoute(ctx context.Context, mcpsr *mcpv1alph
 	}
 	return targetRoute, nil
 
+}
+
+func (r *MCPReconciler) hasValidHTTPRouteReferenceGrant(ctx context.Context, mcpsr *mcpv1alpha1.MCPServerRegistration, targetNamespace string) (bool, error) {
+	refGrantList := &gatewayv1beta1.ReferenceGrantList{}
+	if err := r.List(ctx, refGrantList, client.InNamespace(targetNamespace)); err != nil {
+		return false, fmt.Errorf("failed to list ReferenceGrants: %w", err)
+	}
+	for i := range refGrantList.Items {
+		if referenceGrantAllowsMCPServerRegistrationHTTPRoute(&refGrantList.Items[i], mcpsr) {
+			return true, nil
+		}
+	}
+	return false, nil
+}
+
+func referenceGrantAllowsMCPServerRegistrationHTTPRoute(rg *gatewayv1beta1.ReferenceGrant, mcpsr *mcpv1alpha1.MCPServerRegistration) bool {
+	fromAllowed := false
+	for _, from := range rg.Spec.From {
+		if string(from.Group) == mcpv1alpha1.GroupVersion.Group &&
+			string(from.Kind) == "MCPServerRegistration" &&
+			string(from.Namespace) == mcpsr.Namespace {
+			fromAllowed = true
+			break
+		}
+	}
+	if !fromAllowed {
+		return false
+	}
+
+	for _, to := range rg.Spec.To {
+		if string(to.Group) == gatewayv1.GroupVersion.Group &&
+			(to.Kind == "" || string(to.Kind) == "HTTPRoute") &&
+			(to.Name == nil || *to.Name == "" || string(*to.Name) == mcpsr.Spec.TargetRef.Name) {
+			return true
+		}
+	}
+	return false
 }
 
 func (r *MCPReconciler) getTargetGatewaysFromParentRef(ctx context.Context, parent *gatewayv1.ParentReference) (*gatewayv1.Gateway, error) {
@@ -561,15 +626,21 @@ func (r *MCPReconciler) updateHTTPRouteStatus(ctx context.Context, mcpsr *mcpv1a
 		return nil
 	}
 
-	namespace := mcpsr.Namespace
-	if targetRef.Namespace != "" {
-		namespace = targetRef.Namespace
+	targetNamespace := targetRefNamespace(mcpsr.Namespace, targetRef)
+	if targetNamespace != mcpsr.Namespace {
+		hasGrant, err := r.hasValidHTTPRouteReferenceGrant(ctx, mcpsr, targetNamespace)
+		if err != nil {
+			return err
+		}
+		if !hasGrant && mcpsr.DeletionTimestamp == nil {
+			return nil
+		}
 	}
 
 	httpRoute := &gatewayv1.HTTPRoute{}
 	err := r.Get(ctx, types.NamespacedName{
 		Name:      targetRef.Name,
-		Namespace: namespace,
+		Namespace: targetNamespace,
 	}, httpRoute)
 	if err != nil {
 		if apierrors.IsNotFound(err) {
@@ -659,6 +730,10 @@ func (r *MCPReconciler) SetupWithManager(ctx context.Context, mgr ctrl.Manager) 
 		return fmt.Errorf("failed to setup required index from MCPServerRegistration to httproutes %w", err)
 	}
 
+	if err := setupIndexMCPRegistrationToReferenceGrant(ctx, mgr.GetFieldIndexer()); err != nil {
+		return fmt.Errorf("failed to setup required index from MCPServerRegistration to referencegrants %w", err)
+	}
+
 	if err := setupIndexProgrammedHTTPRoutes(ctx, mgr.GetFieldIndexer()); err != nil {
 		return fmt.Errorf("failed to setup required index for programmed httproutes %w", err)
 	}
@@ -683,6 +758,10 @@ func (r *MCPReconciler) SetupWithManager(ctx context.Context, mgr ctrl.Manager) 
 			&mcpv1alpha1.MCPGatewayExtension{},
 			handler.EnqueueRequestsFromMapFunc(r.findMCPServerRegistrationsForMCPGatewayExtension),
 			builder.WithPredicates(predicate.ResourceVersionChangedPredicate{}),
+		).
+		Watches(
+			&gatewayv1beta1.ReferenceGrant{},
+			handler.EnqueueRequestsFromMapFunc(r.findMCPServerRegistrationsForReferenceGrant),
 		)
 
 	return controller.Complete(r)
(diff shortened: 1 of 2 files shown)


avinxshKD:
Hey @jasonmadigan fixed a small controller bug where MCPServerRegistration.spec.targetRef.namespace was ignored during HTTPRoute lookup....added a regression test and verified with go test ./internal/controller.

Aman-Cool:
Thanks for this @avinxshKD, it's exactly the `getTargetHTTPRoute` namespace bug, and the fix looks right: it now honors `targetRef.namespace` and matches the status-writeback and watch-index paths, so all three finally agree.

I'd actually filed #1199 for this same bug just before David pointed me here.., could you add `Fixes: #1199` to the description? (**also clears the `needs-issue` label.**)

A few notes while it's still draft, no rush:

- **Test:** nice cross-ns test 👍 could you also add the same-name edge case from #1199 (a route with the same name in both namespaces, asserting it picks the `targetRef` one)? That silent mis-bind was the dangerous bit. The empty-namespace fallback would be good to assert too.
- **DRY:** the default-namespace logic now lives in 3 spots in two slightly different shapes; a tiny `targetRefNamespace()` helper would keep them from drifting again (#1186 wants the same for `MCPGatewayExtension`).
- **One question:** should the cross-ns lookup be gated by a `ReferenceGrant`? Gateway API usually wants the target namespace's consent for a cross-ns ref, and we already do this for extension -> Gateway via `HasValidReferenceGrant`; totally fine as a follow-up if you'd rather keep this PR focused.

And whenever your primary PR merges (and if you've nothing else queued), would be great to bring this out of draft. happy to re-review and help get it over the line 🙂

avinxshKD:
Hey @Aman-Cool thanks for reviewing. Updated.

Added the same-name cross-namespace case and pulled the fallback into targetRefNamespace() so lookup/status/index use the same path now.

Left ReferenceGrant out for this one to keep the PR focused. Pls take a look.

cc @david-martin

maleck13 (changes requested):
We can't allow a cross namespace reference without a reference grant. The original code enforced same namespace specifically to stop cross namespace references. Being able to create an MCPServerRegistration is not enough permission wise. You need to have permission to expose that server on the HTTPRoute you are targeting. It being in the same ns is the first check. If it is not in the same namespace it must use a reference grant just like the MCPGatewayExtension does

avinxshKD:
@maleck13 Updated.
Cross-namespace HTTPRoute refs now require a ReferenceGrant before lookup. Added coverage for the same-name route case with a valid grant, and the no-grant rejection path.

cc @david-martin @Aman-Cool

Aman-Cool:
@avinxshKD. Agreeing with @maleck13... creating a registration shouldn't be enough to expose another namespace's server, so same-namespace as the first check and a grant otherwise is the right boundary. The updated fix looks right too; grant check before the lookup, and `targetRefNamespace()` keeping the getter, status update and index in agreement.

**Two things I'd flag though**: there's no `ReferenceGrant` watch, so a grant created or revoked after the registration does nothing until something unrelated triggers a reconcile; the extension controller already watches grants, so it's the other half of the pattern being copied here. And revoking a grant only flips status.., the already-written config stays in the Secret and the broker keeps serving the server after consent was withdrawn, so the no-grant path probably wants a `RemoveMCPServer` so revoking actually revokes.

avinxshKD:
@Aman-Cool Thanks for the review and agreed. Updated this to watch ReferenceGrants as well, and the no-grant path now removes the MCPServer config so revokes take effect.

Also filled in the fallback/cleanup test cases. Pls take a look when get chance.

avinxshKD:
This now covers the full cross-namespace permission path, not just the namespace lookup: ReferenceGrant gate, revoke requeue, stale config cleanup, and regression tests around each path.

maleck13:
Suggestion: when `errReferenceGrantRequired` is detected in Reconcile (controller L147-157), the status update uses `conditionReasonNotReady`. The MCPGatewayExtension controller uses a dedicated `mcpv1alpha1.ConditionReasonRefGrantRequired` (`"ReferenceGrantRequired"`) for this case — see `mcpgatewayextension_controller.go:324`. The constant already exists in the API types package.

Using the same reason here would keep status reasons consistent across controllers and let users distinguish "missing ReferenceGrant" from other NotReady causes when inspecting MCPServerRegistration status conditions.

avinxshKD:
@maleck13 good call, updated this to use ConditionReasonRefGrantRequired for the missing grant path and added a status assertion for it.

avinxshKD:
@Aman-Cool @jasonmadigan ptal when get chance and lmk if changes needed

internal/controller/mcpserverregistration_controller.go
```diff
@@ -41,6 +43,8 @@ const (
```
Patryk-Stefanski, line 40: **Blocking: `errReferenceGrantRequired` falls through to `return ctrl.Result{}, fmt.Errorf("reconcile failed %w", err)`, causing a tight requeue loop.**

When a ReferenceGrant is missing, the ReferenceGrant watcher already handles re-triggering reconciliation once the grant appears. Returning an error here causes the controller to requeue immediately with exponential backoff — burning cycles for nothing, identical to any transient error.

The right behaviour is to update status (already done above) and return `ctrl.Result{}, nil`, trusting the watch to re-trigger. Suggested fix:

```go
if errors.Is(err, errReferenceGrantRequired) {
    statusReason = mcpv1alpha1.ConditionReasonRefGrantRequired
    if removeErr := r.ConfigReaderWriter.RemoveMCPServer(ctx, mcpServerName(mcpsr)); removeErr != nil {
        return ctrl.Result{}, removeErr
    }
    if updateErr := r.updateStatus(ctx, mcpsr, false, statusReason, err.Error()); updateErr != nil {
        if apierrors.IsConflict(updateErr) {
            return ctrl.Result{RequeueAfter: defaultRequeueTime}, nil
        }
        return ctrl.Result{}, fmt.Errorf("reconcile failed: status update failed %w", updateErr)
    }
    return ctrl.Result{}, nil // ReferenceGrant watch re-triggers when grant is created
}
```

Compare with how the deletion finalizer path (line ~128) returns `nil` cleanly after completing its work.

internal/controller/mcpserverregistration_controller.go
```diff
@@ -561,15 +626,21 @@ func (r *MCPReconciler) updateHTTPRouteStatus(ctx context.Context, mcpsr *mcpv1a
...
+		hasGrant, err := r.hasValidHTTPRouteReferenceGrant(ctx, mcpsr, targetNamespace)
+		if err != nil {
+			return err
+		}
+		if !hasGrant && mcpsr.DeletionTimestamp == nil {
+			return nil
+		}
 	}
 
 	httpRoute := &gatewayv1.HTTPRoute{}
```
Patryk-Stefanski, line 640: **Nit: silent no-op when ReferenceGrant is revoked on a live registration should at least log.**

If a ReferenceGrant is revoked while the MCPServerRegistration is not being deleted, this returns `nil` silently — the stale route status stays on the HTTPRoute. That's a safe default, but without a log line an operator has no way to know why route status isn't being cleaned up. Consider:

```go
if !hasGrant && mcpsr.DeletionTimestamp == nil {
    logger.Info("skipping HTTPRoute status update: ReferenceGrant missing", "targetNamespace", targetNamespace)
    return nil
}
```

internal/controller/mcpserverregistration_controller.go
```diff
@@ -692,6 +771,14 @@ func httpRouteIndexValue(namespace, name string) string {
...
 }
 
+func mcpServerRegistrationToRefGrantIndexValue(mcpsr mcpv1alpha1.MCPServerRegistration) string {
+	return fmt.Sprintf("%s/MCPServerRegistration/%s", mcpv1alpha1.GroupVersion.Group, mcpsr.Namespace)
+}
+
+func refGrantFromToMCPServerRegistrationIndexValue(from gatewayv1beta1.ReferenceGrantFrom) string {
+	return fmt.Sprintf("%s/%s/%s", from.Group, from.Kind, from.Namespace)
+}
+
```
Patryk-Stefanski, line 781: **Nit: the two index-key functions form a matched pair but their symmetry is non-obvious.**

`mcpServerRegistrationToRefGrantIndexValue` produces `<group>/MCPServerRegistration/<namespace>` while `refGrantFromToMCPServerRegistrationIndexValue` produces `<group>/<kind>/<namespace>`. They match when the ReferenceGrant's `from.Kind == "MCPServerRegistration"` and `from.Group == mcp.kuadrant.io/v1alpha1`, which is correct — but nothing in the code makes this pairing visible to a future reader.

A short comment on either function linking them would help:

```go
// produces a key matching refGrantFromToMCPServerRegistrationIndexValue for an MCPServerRegistration from
func mcpServerRegistrationToRefGrantIndexValue(mcpsr mcpv1alpha1.MCPServerRegistration) string {
```

internal/controller/mcpserverregistration_controller_test.go
```diff
@@ -71,6 +81,323 @@ func TestMcpsrReferencesSecret(t *testing.T) {
...
+
+			r := &MCPReconciler{Client: fakeClient}
+			mcpsr := &mcpv1alpha1.MCPServerRegistration{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "server",
+					Namespace: "registrations",
+				},
+				Spec: mcpv1alpha1.MCPServerRegistrationSpec{
+					TargetRef: mcpv1alpha1.TargetReference{
+						Name:      "target-route",
```
Patryk-Stefanski, line 136: **Nit: test name is misleading.**

`TestUpdateHTTPRouteStatusRequiresReferenceGrantForCrossNamespaceCleanup` implies that a ReferenceGrant is *required* for cleanup to occur, but the test actually proves the opposite: status is removed even without a grant (during deletion). The name should reflect what is actually under test, e.g. `TestUpdateHTTPRouteStatusCrossNamespaceCleanupOnDeletion`.

internal/controller/mcpserverregistration_controller_test.go
```diff
@@ -71,6 +81,323 @@ func TestMcpsrReferencesSecret(t *testing.T) {
...
+	if requests[0].NamespacedName != client.ObjectKeyFromObject(mcpsr) {
+		t.Fatalf("request = %v, want %v", requests[0].NamespacedName, client.ObjectKeyFromObject(mcpsr))
+	}
+}
+
+func TestReconcileRemovesConfigWhenReferenceGrantIsMissing(t *testing.T) {
+	scheme := testScheme(t)
+
+	mcpsr := &mcpv1alpha1.MCPServerRegistration{
+		ObjectMeta: metav1.ObjectMeta{
```
Patryk-Stefanski, line 284: **Nit: this test asserts the current (broken) error-return behaviour.**

Lines 532–537 assert `err != nil` and that the error message contains `"ReferenceGrant required"`. If the blocking issue above is fixed (Reconcile returns `nil` when the grant is missing), this assertion will need to flip to `if err != nil { t.Fatal(...) }`. Worth updating both together so the test documents the *intended* behaviour, not the current one.

Patryk-Stefanski (changes requested):
Thanks for the fix — the core approach is right and the ReferenceGrant enforcement is the correct security model for cross-namespace references. A few things to address before this merges (see inline comments).

david-martin:
Closing due to inactivity

PR #1148, opened 2026-06-15
refactor: consolidate duplicated tool/prompt logic into entityRegistry

### LINKED ISSUE #1103: Reduce duplicated tool/prompt code in MCPManager
## Summary

`internal/broker/upstream/manager.go` has near-identical code for tools and prompts across several method pairs. The logic (diff, conflict detection, remove-all, get, managed-get, served-get, to-server-type) is structurally the same with only the type swapped.

## Duplicated pairs

| Tools | Prompts |
|-------|---------|
| `diffTools` | `diffPrompts` |
| `findToolConflicts` | `findPromptConflicts` |
| `removeAllTools` | `removeAllPrompts` |
| `getTools` | `getPrompts` |
| `GetManagedTools` | `GetManagedPrompts` |
| `GetServedManagedTool` | `GetServedManagedPrompt` |
| `toolToServerTool` | `promptToServerPrompt` |
| `SetToolsForTesting` | `SetPromptsForTesting` |

## Suggested approach

Extract a generic helper (or use a shared struct with function pointers) that handles diff, conflict detection, and lifecycle for any named MCP entity. The tool/prompt-specific bits (handler func, server add/delete interface) get passed in.

Keep it simple — if generics make this harder to read, a shared pattern with thin wrappers is fine.

## Acceptance criteria

- Duplicated logic is consolidated
- No behavior change
- Existing unit and e2e tests pass

aniketpandey05:
Refactors the upstream manager by extracting a generic `entityRegistry` to eliminate duplicated tool and prompt management logic. Previously, tools and prompts had separate implementations for several operations despite sharing nearly identical behavior. This change consolidates that logic into a single reusable registry abstraction while preserving the existing public API and behavior.

## Changes

- Added `internal/broker/upstream/registry.go` containing the new generic `entityRegistry`.
- Updated `MCPManager` to use separate tool and prompt registries backed by the shared implementation.

closes #1103 

## Summary by CodeRabbit

* **Refactor**
  * Unified upstream tool and prompt lifecycle tracking and gateway synchronization using a shared registry approach to reduce duplicated logic and keep states aligned.
* **Bug Fixes**
  * Improved gateway cleanup and resynchronization of registered tools/prompts when an upstream server is disabled or becomes unreachable (for example, after connection/ping failures).
* **Tests**
  * Updated unit tests to match the new unified tool/prompt management and revised diff/lookup behaviors.

### DIFF
--- a/internal/broker/upstream/manager.go
+++ b/internal/broker/upstream/manager.go
@@ -602,11 +602,6 @@ func (man *MCPManager) setStatus(err error, toolCount int, promptCount int, inva
 	man.status.InvalidToolList = invalidTools
 	man.status.InvalidPrompts = len(invalidPrompts)
 	man.status.InvalidPromptList = invalidPrompts
-	man.status.ProtocolValidation = ProtocolValidation{ExpectedVersion: mcp.LATEST_PROTOCOL_VERSION}
-	if info := man.mcp.ProtocolInfo(); info != nil {
-		man.status.ProtocolValidation.IsValid = true
-		man.status.ProtocolValidation.SupportedVersion = info.ProtocolVersion
-	}
 	if err != nil {
 		man.status.Message = err.Error()
 		man.status.Ready = false
@@ -616,15 +611,12 @@ func (man *MCPManager) setStatus(err error, toolCount int, promptCount int, inva
 	man.status.TotalPrompts = promptCount
 	man.status.Ready = true
 	man.status.Message = fmt.Sprintf("server added successfully. Total tools added %d. Total prompts added %d", toolCount, promptCount)
-<<<<<<< HEAD
 	// always report the version we expect; fill in the negotiated version once it is known
 	man.status.ProtocolValidation = ProtocolValidation{ExpectedVersion: expectedProtocolVersion}
 	if info := man.mcp.ProtocolInfo(); info != nil {
 		man.status.ProtocolValidation.IsValid = true
 		man.status.ProtocolValidation.SupportedVersion = info.ProtocolVersion
 	}
-=======
->>>>>>> 6e161b2 (fix: address coderabbit review comments)
 }
 
 // handleConnectionFailure records a connect/ping failure and backs off the
--- a/internal/broker/upstream/manager.go
+++ b/internal/broker/upstream/manager.go
@@ -668,6 +668,7 @@ func (man *MCPManager) recordBackendError(span trace.Span, err error) {
 		attribute.String("mcp.server", man.mcp.GetName()),
 	)
 }
+
 // SetStatusForTesting sets the status directly for testing purposes.
 // This bypasses the normal status update flow and should only be used in tests.
 func (man *MCPManager) SetStatusForTesting(status ServerValidationStatus) {
--- a/internal/broker/upstream/manager.go
+++ b/internal/broker/upstream/manager.go
@@ -602,6 +602,11 @@ func (man *MCPManager) setStatus(err error, toolCount int, promptCount int, inva
 	man.status.InvalidToolList = invalidTools
 	man.status.InvalidPrompts = len(invalidPrompts)
 	man.status.InvalidPromptList = invalidPrompts
+	man.status.ProtocolValidation = ProtocolValidation{ExpectedVersion: mcp.LATEST_PROTOCOL_VERSION}
+	if info := man.mcp.ProtocolInfo(); info != nil {
+		man.status.ProtocolValidation.IsValid = true
+		man.status.ProtocolValidation.SupportedVersion = info.ProtocolVersion
+	}
 	if err != nil {
 		man.status.Message = err.Error()
 		man.status.Ready = false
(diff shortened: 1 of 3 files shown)


aniketpandey05:
@coderabbitai review

aniketpandey05:
@coderabbitai review

aniketpandey05:
@maleck13 please review this one ,  thank you

Aman-Cool:
@rogueslasher,Read this with my guard up... it's a refactor of the federation state, so all that matters is whether it preserves behaviour. Went looking for a dropped guard and couldn't find one: the generic `findConflicts` keeps the same name + server-ID check, and it's still one `toolsLock` over both registries with `removeAll` releasing before the gateway call. Tidy.

One small thing: in `diff`, the remove path builds a whole `ServerTool` just to read its prefixed name, where the old code did a plain string concat.., minor, but cheap to prefix directly given how allocation-conscious the broker path is.

Nice cleanup... the generics read well.

aniketpandey05:
@coderabbitai review

aniketpandey05:
@coderabbitai review

aniketpandey05:
@coderabbitai review

aniketpandey05:
@coderabbitai review

aniketpandey05:
@maleck13 please review this

thank you

internal/broker/upstream/manager.go
```diff
@@ -522,6 +533,11 @@ func (man *MCPManager) setStatus(err error, toolCount int, promptCount int, inva
 	man.status.InvalidToolList = invalidTools
 	man.status.InvalidPrompts = len(invalidPrompts)
 	man.status.InvalidPromptList = invalidPrompts
+	man.status.ProtocolValidation = ProtocolValidation{ExpectedVersion: mcp.LATEST_PROTOCOL_VERSION}
+	if info := man.mcp.ProtocolInfo(); info != nil {
+		man.status.ProtocolValidation.IsValid = true
+		man.status.ProtocolValidation.SupportedVersion = info.ProtocolVersion
```
Patryk-Stefanski: **Blocking: moves `ProtocolValidation` to the error path — unreviewed behaviour change not in scope for a refactor.**

On `main`, `ProtocolValidation` was only written on the success path (after the `if err != nil { return }` guard). This PR moves it before the guard, so callers receiving `Ready: false` (connection failure, ping failure, disabled server) now also get a populated `ProtocolValidation` field.

Please revert to the original position (success path only), or if populating it on the error path is intentional, extract it as a separate commit with a rationale and a test.

internal/broker/upstream/registry.go
```diff
@@ -0,0 +1,311 @@
...
+		if _, exists := oldMap[name]; !exists {
+			toAdd = append(toAdd, r.toServer(item))
+		}
+	}
+	for name, item := range oldMap {
+		if _, exists := newMap[name]; !exists {
+			toRemove = append(toRemove, r.getServerName(r.toServer(item)))
+		}
+	}
+	return toAdd, toRemove
```
Patryk-Stefanski, line 111: **Blocking: `r.toServer(item)` is called here just to extract the prefixed name — unnecessary allocation on every reconcile.**

`r.toServer` for a tool constructs a full `server.ServerTool` including metadata and a handler closure, when all that's needed is the served name. On `main`, `diffTools` used `prefixedName(man.mcp.GetPrefix(), oldTool.Name)` directly.

Consider storing `prefix` on the registry (already threaded through `newToolRegistry`/`newPromptRegistry`) and using it directly:

```go
for name := range oldMap {
    if _, exists := newMap[name]; !exists {
        toRemove = append(toRemove, prefixedName(r.prefix, name))
    }
}
```

internal/broker/upstream/registry.go
```diff
@@ -0,0 +1,311 @@
...
+				r.logger.Error("nil entry in gateway items",
+					"upstream mcp server", r.serverID, r.entityTag, existingName)
+				continue
+			}
+			meta := r.getMetaFields(*existingPtr)
+			if meta == nil {
+				r.logger.Error("unable to check conflict, meta is nil",
+					"upstream mcp server", r.serverID, r.entityTag, existingName)
+				continue
+			}
```
Patryk-Stefanski, line 136: **Blocking: log level escalated from `Debug` to `Error` for expected skip conditions.**

On `main`, nil meta and missing ID were logged at `Debug` with "skipping conflict check" messages. These are expected for tools not managed by this gateway and will fire on every reconcile tick. Logging them as `Error` will produce spurious noise in operator logs.

Please revert the three skip/continue cases (nil gateway entry, meta is nil, id is missing) back to `Debug`. The actual conflict detection at the bottom of the loop is correctly at `Debug` and could reasonably stay or be promoted separately.

internal/broker/upstream/registry.go
```diff
@@ -0,0 +1,311 @@
...
+		entityTag:         entityTag,
+	}
+}
+
+// toServerItem converts an upstream entity to its gateway-facing server item.
+// Replaces toolToServerTool / promptToServerPrompt.
+func (r *entityRegistry[E, S]) toServerItem(item E) S {
+	return r.toServer(item)
+}
+
```
Patryk-Stefanski: **Nit: `toServerItem` is a one-liner wrapper over `r.toServer` with no added value.**

This only exists so tests can call it by a stable exported name. Either call `r.toServer(item)` directly in the tests, or remove this method.

Patryk-Stefanski (changes requested):
The DRY consolidation is justified and the `entityRegistry` abstraction is clean. Three issues need fixing before merge: a silent behaviour change in `setStatus`, log level escalation in `findConflicts`, and an unnecessary allocation in `diff`.

PR #1154, opened 2026-06-17
fix: align CI Kind cluster setup with pinned binary across all workflows

jasonmadigan:
## Summary

- GitHub Actions runners updated containerd to config v4, which the runner-preinstalled Kind can't handle. The benchmark workflow was using bare `kind create cluster` instead of the pinned `bin/kind` via `make kind-create-cluster-ci`, causing `kind load image-archive` to fail with `unknown containerd config version: 4`.
- Fixed benchmark and added the CI node image probe step to e2e-on-demand, e2e-auth, and conformance for consistency — all six Kind-using workflows now have the same setup pattern.

## Test evidence

Verified no remaining bare `kind` commands across all workflows:
```
grep -rn 'kind create cluster\|kind load' .github/workflows/ → no results
```

All six workflows now use `make kind-create-cluster-ci` with optional `KIND_CLUSTER_IMAGE` passthrough.

Relates to https://github.com/Kuadrant/mcp-gateway/actions/runs/27692802011

## Summary by CodeRabbit

* **Chores**
  * Updated multiple CI/CD workflows (conformance tests, e2e auth, e2e on-demand, and performance benchmarks) to explicitly set job-level permissions and optimize test infrastructure by probing for pre-built CI node images when available, with automatic fallback to default cluster creation for improved pipeline efficiency.

### DIFF
--- a/.github/workflows/conformance.yaml
+++ b/.github/workflows/conformance.yaml
@@ -33,6 +33,9 @@ jobs:
   conformance-tests:
     runs-on: ubuntu-latest
     timeout-minutes: 30
+    permissions:
+      contents: read
+      packages: read
 
     steps:
       - name: Checkout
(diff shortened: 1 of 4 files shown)


david-martin (approved):
These changes make sense.

Issue #1155, opened 2026-06-17
Hairpin client uses fixed SNI, breaking tools/call on HTTPS listeners with per-server hostnames


maleck13:
## Problem

The hairpin HTTP client uses a single `*http.Client` with a fixed TLS ServerName (SNI) set to the gateway's public hostname. Envoy selects filter chains at TLS handshake time using SNI from the ClientHello — before HTTP headers are parsed ([Envoy SNI docs](https://www.envoyproxy.io/docs/envoy/latest/faq/configuration/sni)). When an MCP server's HTTPRoute hostname differs from the gateway's public hostname, the hairpin sends the wrong SNI. While the ext_proc rewrites `:authority` for HTTP routing, the filter chain is already locked in by SNI.

This means `tools/call` fails for servers on HTTPS listeners when the server hostname differs from the gateway's public hostname.

## Solution

Replace the single `*http.Client` with a `HairpinClientPool` that caches per-SNI clients. `Initialize` calls `pool.Get(conf.Hostname)` to get a client with the server's hostname as SNI. For HTTP (no TLS), the default client is returned.

Three TLS cases are handled:

| Case | Listener | SNI |
|------|----------|-----|
| 1 | HTTP | none (no TLS) |
| 2 | HTTPS | server's hostname from HTTPRoute |
| 3 | TLS backend | DestinationRule for TLS origination |

## E2E Coverage

- Hairpin SNI test: plain HTTP backend registered on HTTPS listener, `tools/call` verifies correct filter chain selection
- DestinationRule test: internal TLS backend fails without DR, succeeds with it

maleck13:
@david-martin @trepel @jasonmadigan  this is the issue with the TLS and multiple listener setup. I have a fix in progress that we should then backport to 0.7

david-martin:
Fixed by #1156 (merged) and cherry-picked to 0.7.1 via #1159.

PR #1156, opened 2026-06-17
per-server SNI for hairpin TLS connections

maleck13:
- Replace single *http.Client hairpin with HairpinClientPool that caches per-SNI clients
- Initialize uses the server HTTPRoute hostname as TLS ServerName so Envoy selects the correct filter chain
- Add e2e tests for HTTPS hairpin use cases and DestinationRule TLS origination
- Document supported TLS configurations in the install guide

Fixes #

#1155 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved TLS connection handling for backend servers with multiple hostnames.
  * Fixed support for internal HTTPS backends by enabling TLS origination workaround via Istio DestinationRule.

* **Documentation**
  * Enhanced "Supported TLS Configurations" section with clearer guidance on TLS termination behavior and workarounds for internal HTTPS backends.

* **Tests**
  * Expanded e2e test coverage for TLS scenarios, including hostname-based routing and DestinationRule-driven TLS origination.

### DIFF
--- a/cmd/mcp-broker-router/main.go
+++ b/cmd/mcp-broker-router/main.go
@@ -81,7 +81,7 @@ type app struct {
 	jwtMgr         *session.JWTManager
 	elicitMap      idmap.Map
 	tokenElicitMap elicitation.Map
-	hairpinClient  *http.Client
+	hairpinPool    *clients.HairpinClientPool
 	mcpBroker      broker.MCPBroker
 	tokenHandler   http.Handler
 	elicitHandler  http.Handler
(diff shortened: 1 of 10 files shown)


david-martin (commented):
Reviewed the changes at a high level, and sounds reasonable with good test coverage and docs coverage.

One question: should we remove clients from the pool if a server is removed from the config? Small source of potential memory leak?

david-martin (approved):

maleck13:
I noticed this also . yes it is technically a leak but it is really small and we are talking about a relatively (<=500) static servers in an environment so the small memory impact is ok imo vs implementing a clean up . It might a nice improvement at somepoint

PR #1162, opened 2026-06-18
feat: local-env-setup should use HTTPS by default with a self-signed cert

### LINKED ISSUE #1142: local-setup-env should use HTTPS by default with a self-signed cert
## Summary

The e2e test infrastructure already uses HTTPS with a private CA certificate, but `make local-setup-env` still defaults to plain HTTP. The local dev environment should match the e2e setup and use HTTPS by default with a self-signed certificate.

## Motivation

- Closer parity between local dev and CI/e2e environments reduces "works on my machine" surprises
- TLS-related bugs (like the hairpin init path) are only caught if the local setup exercises HTTPS
- Developers testing AuthPolicy, OAuth, or URL elicitation flows locally benefit from realistic TLS

## Acceptance Criteria

- `make local-setup-env` provisions a self-signed CA and TLS certificate for the gateway
- The gateway listener defaults to HTTPS
- The MCP endpoint URL printed at the end uses `https://`
- Existing `make local-setup-env` users can still override to HTTP if needed

manik3160:
The e2e test infrastructure already uses HTTPS with a private CA certificate, but `make local-setup-env` still defaults to plain HTTP. This PR updates the local development environment to match the e2e setup and use HTTPS by default with a self-signed certificate.

Closes #1142

## Changes
- Add `config/istio/gateway/local-gateway-tls.yaml` to provision a self-signed CA and TLS certificate for the local gateway
- Modify the gateway `mcp` listener to default to `HTTPS` with TLS termination
- Update NodePort config to map port `30080` to the new `8443` HTTPS listener
- Update `local-env-setup` Makefile target to install `cert-manager` and apply the local TLS resources
- Update CI `ci-setup` target to also apply the new TLS resource so the gateway programs successfully
- Add `GATEWAY_SCHEME` variable (defaulting to `https`) to update all CLI URL output references (inspect, info, dev)

Existing users can still override to plain HTTP by running `make local-env-setup GATEWAY_SCHEME=http`.

## Summary by CodeRabbit

* **New Features**
  * Gateway now operates over HTTPS with automatic local TLS certificate provisioning and management.

* **Chores**
  * Updated gateway and NodePort mappings to use HTTPS (port 8443, port 30080 for MCP) instead of HTTP (port 8080).
  * Switched local/dev/inspection and printed URLs to respect the selected gateway scheme.
  * Improved CI/conformance HTTPS connectivity by updating the conformance endpoint and CI TLS handling.

* **Tests**
  * Updated e2e gateway listener port and default test URL fallbacks to match HTTPS.

* **Documentation**
  * Updated the MCP server registration sample to include the `/mcp` path.

### DIFF
--- a/.github/workflows/conformance.yaml
+++ b/.github/workflows/conformance.yaml
@@ -27,7 +27,8 @@ concurrency:
 
 env:
   CONFORMANCE_VERSION: "0.1.15"
-  MCP_URL: "http://mcp.127-0-0-1.sslip.io:8001/mcp"
+  MCP_URL: "https://mcp.127-0-0-1.sslip.io:8001/mcp"
+  NODE_TLS_REJECT_UNAUTHORIZED: "0"
 
 jobs:
   conformance-tests:
--- a/.github/workflows/conformance.yaml
+++ b/.github/workflows/conformance.yaml
@@ -89,6 +89,7 @@ jobs:
         run: |
           make ci-setup
           kubectl apply -k config/mcp-gateway/base/ -n mcp-system
+          make patch-broker-local-ca
 
       - name: Deploy conformance server
         run: make deploy-conformance-server
--- a/Makefile
+++ b/Makefile
@@ -475,6 +475,17 @@ deploy-test-servers: kind-load-test-servers ## Deploy test MCP servers for local
 	  echo "Gateway IP: $$GATEWAY_IP"; \
 	  kubectl patch deployment mcp-oidc-server -n mcp-test --type='json' -p="$$(cat config/keycloak/patch-hostaliases.json | envsubst)"
 
+# Patch broker deployment to trust local self-signed CA
+.PHONY: patch-broker-local-ca
+patch-broker-local-ca: ## Patch broker deployment to trust local self-signed CA
+	@echo "Patching broker-router deployment to trust local self-signed CA..."
+	@kubectl wait --for=condition=Available deployment/mcp-gateway -n mcp-system --timeout=120s
+	@kubectl get secret private-ca-keypair -n cert-manager -o jsonpath='{.data.ca\.crt}' | base64 -d > out/certs/ca.crt || true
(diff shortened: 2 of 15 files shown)


Patryk-Stefanski (changes requested):
Two blocking issues and a couple of nits.

**Blocking: Kind cluster YAML not updated**

The PR moves the `mcp` nodeport from 8080 to 8443 in `nodeport.yaml` and remaps keycloak from 30081 to 30089, but `config/kind/cluster.yaml` and `config/kind/cluster-ci.yaml` still map `30080 → mcp-gateway (standard)`, `30081 → keycloak`, `30082 → mcp-gateway HTTPS`. Kind host port mappings are baked in at cluster creation — they can't be changed without recreating the cluster. After `make local-env-setup`, traffic on `localhost:30080` hits an Envoy listener that now expects HTTPS, resulting in TLS errors with no obvious explanation. The Kind cluster YAML files must be updated, or the PR must document that users need to recreate their cluster.

**Blocking: Conformance server HTTPRoute now claims the main gateway hostname**

Adding `mcp.127-0-0-1.sslip.io` to `conformance-server-httproute.yaml` routes all traffic for the main gateway hostname to the conformance server. This will break any local dev environment where the everything-server or other registered servers are the expected default backend. If CI needs a specific URL, set it via the `MCP_URL` env var rather than attaching the route to the main hostname.

**Nit: NODE_TLS_REJECT_UNAUTHORIZED is too broad**

`NODE_TLS_REJECT_UNAUTHORIZED: "0"` in `.github/workflows/conformance.yaml` disables TLS verification for all outbound TLS calls in that job. Prefer `NODE_EXTRA_CA_CERTS` pointing at the extracted CA cert so the suppression is scoped to the self-signed cert only.

**Nit: Silent failure in patch-broker-local-ca**

The `|| true` on the `kubectl get secret ... | base64 -d > out/certs/ca.crt` line swallows failures. If the secret doesn't exist yet, the CA file will be empty and the subsequent `kubectl create secret` will silently create an empty bundle — broken TLS trust with no error. Drop `|| true` and let it fail loudly, or add an explicit existence check first.

david-martin:
Closing due to inactivity

PR #1163, opened 2026-06-18
update docs to call out single listener pattern for https

maleck13:
## What does this PR do?

Clarify requirements on listeners. 

## Summary by CodeRabbit

* **Documentation**
  * Updated MCP Gateway listener/router guidance to emphasize port usage (including the MCP Gateway on port 8080).
  * Clarified HTTP vs HTTPS routing behavior, noting HTTPS TLS filter-chain isolation and related `:authority` limitations.
  * Added/expanded guidance recommending a single HTTPS listener with a wildcard hostname to support both client and backend routing.
  * Included a new example `kubectl patch` workflow and improved formatting/structure for listener setup steps.

### DIFF
--- a/docs/guides/configure-mcp-gateway-listener-and-router.md
+++ b/docs/guides/configure-mcp-gateway-listener-and-router.md
@@ -11,12 +11,12 @@ This guide covers adding the required MCP listeners to your existing Gateway. Th
 
 ## Step 1: Add MCP Listeners to Gateway
 
-MCP Gateway requires two listeners on your Gateway:
+### HTTP (multiple listeners)
+
+With HTTP, you can use two listeners on the same port — one for client traffic and one for backend server routing:
 
 - **`mcp`**: the public-facing listener for MCP client traffic. The hostname must resolve to your Gateway's external address.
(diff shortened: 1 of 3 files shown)


docs/guides/configure-mcp-gateway-listener-and-router.md
```diff
@@ -57,6 +51,47 @@ kubectl get gateway your-gateway-name -n your-gateway-namespace -o jsonpath='{.s
...
+
+Verify the listener was added:
+
+```bash
+kubectl get gateway your-gateway-name -n your-gateway-namespace -o jsonpath='{.spec.listeners[*].name}'
+```
+
+You should see `mcp-tls` in the output.
+
+> **Important:** If you installed MCP Gateway using Helm, ensure the `gateway.publicHost` value in your Helm values matches the hostname above. For example:
```
jasonmadigan, line 95: nit: this note was removed from the HTTP bit above. so users following the HTTP path won't see it now. suggest: duplicate it there too or move it before step 2 so both paths get it

jasonmadigan (approved):

PR #1165, opened 2026-06-19
fix(broker): report served tool/prompt count on upstream status error path

### LINKED ISSUE #1164: broker: MCPServerRegistration keeps reporting a stale tool count after an upstream goes down
While testing what happens when a backend MCP server drops out, I noticed the tool count on its MCPServerRegistration never goes back down.

`MCPManager.manage()` runs on a ticker. When the upstream connection or ping fails (`manager.go:329`, `manager.go:341`) it calls `removeAllTools()` / `removeAllPrompts()` — which empties `man.tools` and pulls every tool off the gateway — then calls `setStatus(err, ...)` to mark the server not ready.

But `setStatus` returns early on the error path (`manager.go:525-528`) and never touches `TotalTools` / `TotalPrompts`:

```go
if err != nil {
    man.status.Message = err.Error()
    man.status.Ready = false
    return            // TotalTools / TotalPrompts keep the last successful value
}
man.status.TotalTools = toolCount
man.status.TotalPrompts = promptCount
```

So `status.TotalTools` keeps whatever the last healthy cycle wrote. The controller forwards it straight through — `mcpserverregistration_controller.go:368` writes `int32(gatewayServerStatus.TotalTools)` into `status.discoveredTools`, which is the `Tools` printer column on the CRD (`api/v1alpha1/types.go:37`).

Result: a server loses its connection, all its tools are removed from the gateway, the condition flips to `Ready=False`, but `kubectl get mcpserverregistration` still shows the old count beside it:

```
NAME              READY   TOOLS
weather-service   False   15
```

The readiness and the tool count disagree, which is misleading for anyone watching the CRD or a dashboard built on it.

One thing worth calling out so a fix doesn't overcorrect: on a transient `tools/list` failure under the default `FilterOut` policy (`manager.go:371`) the tools are NOT removed — the gateway keeps serving the cached set — so there the existing count is correct. So the fix isn't "force the count to 0 on error", it's "report the count that's actually being served":

```go
if err != nil {
    man.status.Message = err.Error()
    man.status.Ready = false
    man.toolsLock.RLock()
    man.status.TotalTools = len(man.tools)
    man.status.TotalPrompts = len(man.prompts)
    man.toolsLock.RUnlock()
    return
}
```

That reports 0 when tools were removed (connect/ping failure, or RejectServer) and keeps the real count when they're still served. It lines up with the success path, where `toolCount == len(man.tools)`.

Happy to send a small PR with a test that drops a server mid-cycle and asserts the count goes back to 0.

namansh70747:
Fixes #1164

When an upstream connection or ping fails, `manage()` removes the server's tools and prompts and then calls `setStatus` to mark it not ready. `setStatus` returned early on the error path without updating `TotalTools`/`TotalPrompts`, so the status kept the count from the last healthy cycle. The controller surfaces that value as `status.discoveredTools` (the `Tools` printer column), so `kubectl get mcpserverregistration` showed a server as `Ready=False` while still listing its old tool count.

This reports the count actually being served on the error path:

- `0` when the tools were removed (connect/ping failure, rejected server)
- the cached count when a transient `tools/list` error leaves the previously served set in place (default `FilterOut` policy), so a momentary list blip doesn't wrongly report 0 while tools are still served

The read is guarded by the existing `toolsLock`, which no `setStatus` caller holds, so there's no deadlock. It lines up with the success path, where `toolCount == len(man.tools)`.

Test: `TestMCPManager_setStatus_ErrorReportsServedCount` covers both cases (removed → 0, still served → keeps count). It fails on the old behaviour and passes with the fix.

Noted on the issue that the MCPServerRegistration/Status endpoint coupling is changing soon — happy for this to be superseded by that work; sending it as a small interim fix.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved accuracy of tool and prompt counts during error states, ensuring proper status reporting when the system encounters operational issues.

### DIFF
--- a/internal/broker/upstream/manager.go
+++ b/internal/broker/upstream/manager.go
@@ -525,6 +525,13 @@ func (man *MCPManager) setStatus(err error, toolCount int, promptCount int, inva
 	if err != nil {
 		man.status.Message = err.Error()
 		man.status.Ready = false
+		// report the tools/prompts actually being served, not a stale count: 0 when they were
+		// removed (connect/ping failure, rejected server), or the cached count when a transient
+		// tools/list error leaves the previously served set in place.
+		man.toolsLock.RLock()
+		man.status.TotalTools = len(man.tools)
+		man.status.TotalPrompts = len(man.prompts)
+		man.toolsLock.RUnlock()
 		return
 	}
 	man.status.TotalTools = toolCount
--- a/internal/broker/upstream/manager_test.go
+++ b/internal/broker/upstream/manager_test.go
@@ -420,6 +420,44 @@ func TestMCPManager_setStatus_ProtocolVersions(t *testing.T) {
 	}
 }
 
+// TestMCPManager_setStatus_ErrorReportsServedCount checks the error path reports the count of
+// tools/prompts actually being served rather than leaving a stale count from the last healthy
+// cycle. When the server goes not-ready and its tools were removed the count must drop to 0; when
+// a transient list error leaves the cached set in place the count must stay accurate.
+func TestMCPManager_setStatus_ErrorReportsServedCount(t *testing.T) {
+	logger := slog.New(slog.NewTextHandler(os.Stdout, nil))
+
+	t.Run("clears count when tools were removed", func(t *testing.T) {
+		mock := newMockMCP("test-server", "test_")
+		manager, err := NewUpstreamMCPManager(mock, newMockToolsAdderDeleter(), nil, logger, 0, mcpv1alpha1.InvalidToolPolicyFilterOut)
+		require.NoError(t, err)
+		// simulate a prior healthy cycle that reported tools, then a failure that removed them
+		manager.status.TotalTools = 5
+		manager.status.TotalPrompts = 2
+
+		manager.setStatus(fmt.Errorf("connection lost"), 0, 0, nil, nil)
+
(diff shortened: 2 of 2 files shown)


namansh70747:
Thanks again for the steer on #1164, @maleck13 — put the small fix up here whenever you get a chance. No rush.

jasonmadigan:
~~Thanks for filing the issue and the fix. This is part of a broader change we're planning internally (#629) which will remove the status syncing that causes this staleness. Since that work is already assigned to a maintainer, we'll address it there. Generally, if an issue is assigned it's best to check before sending a PR, saves duplicated effort. Appreciate you flagging it though.~~

oops, this was closed prematurely - I think this is fine as a follow on to #1187 when that lands. keeping open

namansh70747:
referencing #1164 as the linked bug — that one's still `triage/needs-triage`; if accepting it unblocks the `triage/needs-issue` label here, happy to wait. otherwise let me know if a separate issue would be cleaner.

also addressed coderabbit's comment: seeded `manager.prompts` and added the `TotalPrompts` assertion in the transient-error sub-test to lock that contract down too.

jasonmadigan:
superseded by #1187 which deleted `discoveredTools` and `ServerValidator` entirely as part of #629. thanks @namansh70747 for the fix and analysis.

PR #1141, opened 2026-06-15
fix(broker-router): honor `--log-level=4` (warn) instead of defaulting to debug

PRAteek-singHWY:
## What

`setupLogger()` selected the slog level with a switch that only handled `0`
(info), `8` (error), and `-4` (debug). The `--log-level` flag help advertises
`4=warn`, but `4` had no case and fell through to `default`, which set
`slog.LevelDebug`. As a result, requesting warn produced the most verbose log
level instead.

This replaces the switch with a direct conversion from the flag value to
`slog.Level`, so all slog levels (including warn) are honored:

```go
opts.Level = slog.Level(a.brokerCfg.logLevel)
```

A unit test is added for the level mapping (`setupLogger` was previously
untested).

## Why

The flag value is already a raw `slog.Level` (`Info=0`, `Warn=4`, `Error=8`,
`Debug=-4`), so `4` is valid, documented input. Mapping it to debug means an
operator asking for warn gets the noisiest level instead: more log volume, more
risk of sensitive data at debug, and extra work on the broker/router hot paths.
The level is reachable in normal operation because the controller forwards
`BROKER_ROUTER_LOG_LEVEL` to `--log-level` (see `docs/release-notes/0.0.8.md`).

## Changes

- `cmd/mcp-broker-router/main.go`: map `--log-level` directly to `slog.Level` in
  `setupLogger()`.
- `cmd/main.go`: same fix for the controller binary, which had the identical
  switch.
- `cmd/mcp-broker-router/main_test.go`: add a test covering info/warn/error/debug
  and an arbitrary value.

## Testing

- `make test-unit` passes.
- `gofmt`/`goimports` clean on the changed files.

## Notes

Bugfix, exempt from the issue-first requirement per `CONTRIBUTING.md`. Tracking
issue: #1143. Commit is signed off (`git commit -s`).

Fixes #1143

## Summary by CodeRabbit

* **New Features**
  * Updated `--log-level` help text to document the expected numeric values.
* **Refactor**
  * Simplified and standardized log level handling so configured integer values are applied consistently to the logger.
* **Tests**
  * Added unit tests validating log level behavior for debug, info, warn, error, and an arbitrary/unexpected value.

PR #1172, opened 2026-06-19
fix(broker): stop forwarding `Cookie`/`Proxy-Authorization` to user-specific upstreams

### LINKED ISSUE #1170: broker: client `Cookie`/`Proxy-Authorization` headers are forwarded to every user-specific upstream
## Summary

When building a federated tool list, the broker forwards the client's request
headers to each `userSpecificList` upstream so the upstream can return a
per-user tool list. The header filter (`filterUserHeaders`) strips only
`mcp-session-id` and `x-mcp-*`, so it also forwards **`Cookie`** and
**`Proxy-Authorization`** — credentials that are scoped to the gateway
origin/hop, not the upstream. These are sent to **every** user-specific
upstream queried for a single `tools/list`.

## Where

`internal/broker/user_specific_tools.go`

- `filterUserHeaders` (the filter) keeps everything except `mcp-session-id` and `x-mcp-*`:

```go
func filterUserHeaders(h http.Header) map[string]string {
	headers := make(map[string]string, len(h))
	for key, vals := range h {
		lower := strings.ToLower(key)
		if lower == "mcp-session-id" {
			continue
		}
		if strings.HasPrefix(lower, "x-mcp-") {
			continue
		}
		if len(vals) > 0 {
			headers[key] = vals[0]
		}
	}
	return headers
}
```

- `FetchUserSpecificTools` computes these headers once and fans them out to all
  user-specific servers concurrently; `doFetchTools` opens a direct
  broker→upstream connection (`transport.WithHTTPHeaders(userHeaders)`), so this
  path does not traverse Envoy.

## Impact

- A client's gateway-/origin-scoped session `Cookie` is disclosed to third-party
  upstream MCP servers. A malicious or compromised user-specific upstream
  receives the cookie and can replay it against the gateway origin.
- `Proxy-Authorization` is a hop-by-hop credential and likewise must not be
  proxied to upstreams.
- Because a `tools/list` queries every user-specific server, the leak is
  broadcast to all of them, not just one backend.

## Expected behaviour

`Cookie` and `Proxy-Authorization` should never be forwarded to upstreams. The
client's `Authorization` header should continue to be forwarded — user-specific
servers rely on it for per-user identity (see
`docs/design/security-architecture.md`).

## Scope / related

This issue is limited to the clearly gateway-scoped headers above. The broader
question of `Authorization` reaching multiple distinct upstreams on the
user-specific fan-out (and whether RFC 8693 token exchange — the blast-radius
mitigation described in `security-architecture.md` — can apply on a path that
bypasses Envoy/Authorino) is a separate design discussion and is intentionally
out of scope here.

## Suggested fix

Add `Cookie` and `Proxy-Authorization` to the set of headers stripped by
`filterUserHeaders`, and add test coverage to the existing `TestFilterUserHeaders`
table.

PRAteek-singHWY:
## What

When the broker builds a federated tool list, it forwards the client's request
headers to each `userSpecificList` upstream so the upstream can return a
per-user tool list. `filterUserHeaders` stripped only `mcp-session-id` and
`x-mcp-*`, so it also forwarded `Cookie` and `Proxy-Authorization`.

This adds those two headers to the strip set, so they are never sent to
upstreams. The client's `Authorization` header is intentionally still forwarded
— user-specific servers rely on it for per-user identity.

A case is added to the existing `TestFilterUserHeaders` table covering the new
behaviour.

## Why

`Cookie` and `Proxy-Authorization` are scoped to the gateway origin/hop, not the
upstream. `FetchUserSpecificTools` computes the forwarded headers once and fans
them out to every user-specific upstream concurrently, over direct
broker→upstream connections that bypass Envoy. As a result a client's
gateway-/origin-scoped session cookie was disclosed to every user-specific
upstream queried for a single `tools/list`; a malicious or compromised upstream
could replay it against the gateway origin. `Proxy-Authorization` is a
hop-by-hop credential and must not be proxied either.

Forwarding `Authorization` to upstreams is documented and intended
(`docs/design/security-architecture.md`), so it is left unchanged.

## Changes

- `internal/broker/user_specific_tools.go`: add `sensitiveForwardHeaders`
  (`cookie`, `proxy-authorization`) and skip them in `filterUserHeaders`;
  clarify in the doc comment that `Authorization` is preserved by design.
- `internal/broker/user_specific_tools_test.go`: add a `TestFilterUserHeaders`
  case asserting `Cookie`/`Proxy-Authorization` are stripped and `Authorization`
  is preserved.

## Testing

- `go test ./internal/broker/...` passes.
- `gofmt`/`goimports` clean on the changed files.

## Notes

Security fix, exempt from the issue-first requirement per `CONTRIBUTING.md`.
Tracking issue: #1170. Commit is signed off (`git commit -s`).

Out of scope: whether `Authorization` should reach multiple distinct upstreams
on the user-specific fan-out, and how RFC 8693 token exchange applies on a path
that bypasses Envoy/Authorino — tracked separately for design discussion.

Fixes #1170

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced security by preventing sensitive gateway credentials from being forwarded to upstream servers during user-specific tool operations. The system now filters out cookies and proxy authorization headers to mitigate credential leakage risks.

### DIFF
--- a/internal/broker/user_specific_tools.go
+++ b/internal/broker/user_specific_tools.go
@@ -228,8 +228,19 @@ func gatewaySessionTTL(gatewaySessionID string) time.Duration {
 	return ttl
 }
 
+// sensitiveForwardHeaders are client headers that must never be forwarded to
+// upstream MCP servers. cookie and proxy-authorization are scoped to the
+// gateway origin/hop, not the upstream, so forwarding them would leak
+// gateway-scoped credentials to every user-specific upstream queried.
+var sensitiveForwardHeaders = map[string]struct{}{
+	"cookie":              {},
+	"proxy-authorization": {},
+}
+
 // filterUserHeaders returns user headers suitable for forwarding to upstream,
-// stripping internal gateway headers.
+// stripping internal gateway headers and gateway-scoped credentials. the
+// client's Authorization header is intentionally preserved: user-specific
+// servers rely on it to return a per-user tool list.
 func filterUserHeaders(h http.Header) map[string]string {
 	headers := make(map[string]string, len(h))
 	for key, vals := range h {
@@ -238,6 +249,9 @@ func filterUserHeaders(h http.Header) map[string]string {
 			continue
 		}
 		if strings.HasPrefix(lower, "x-mcp-") {
+			continue
+		}
+		if _, sensitive := sensitiveForwardHeaders[lower]; sensitive {
 			continue
 		}
 		if len(vals) > 0 {
--- a/internal/broker/user_specific_tools_test.go
+++ b/internal/broker/user_specific_tools_test.go
@@ -75,6 +75,17 @@ func TestFilterUserHeaders(t *testing.T) {
 				"Accept": "text/html",
 			},
 		},
+		{
+			name: "strips cookie and proxy-authorization, preserves authorization",
+			input: http.Header{
+				"Cookie":              []string{"session=secret"},
+				"Proxy-Authorization": []string{"Basic abc"},
+				"Authorization":       []string{"Bearer user-token"},
+			},
+			expected: map[string]string{
+				"Authorization": "Bearer user-token",
+			},
+		},
 	}
 
 	for _, tc := range tests {


PRAteek-singHWY:
Hello @jasonmadigan @david-martin , could you please have a look whenever you get a chance?

This change is small and self-contained: it stops the broker forwarding `Cookie` and `Proxy-Authorization` to user-specific upstreams on the tools/list fan-out (Fixes #1170), with a unit test added to `TestFilterUserHeaders`. CodeRabbit had no actionable comments and all checks are green.

Whenever you get a chance to review, I'd really appreciate it. Once this and #1141 are reviewed I'll pick up the next change and keep exploring other areas of the gateway. 

Looking forward to your guidance and happy to iterate on anything.

david-martin (approved):
These two headers are extremely unlikely to be intended for upstream MCP servers. `Cookie` in particular could contain a value scoped to the gateway or something in front of it (which ideally gets dropped after verification, but that is not guaranteed). Safe change.

If this list needs to grow in future, users have options at the gateway layer: AuthPolicy or HTTPRoute header filters can strip headers before they reach the broker, so this does not need to become an unbounded blocklist.

david-martin:
Thanks @PRAteek-singHWY

PRAteek-singHWY:
> These two headers are extremely unlikely to be intended for upstream MCP servers. `Cookie` in particular could contain a value scoped to the gateway or something in front of it (which ideally gets dropped after verification, but that is not guaranteed). Safe change.
> 
> If this list needs to grow in future, users have options at the gateway layer: AuthPolicy or HTTPRoute header filters can strip headers before they reach the broker, so this does not need to become an unbounded blocklist.


Hi @david-martin Sir

Noted on the blocklist point. For anything beyond these clearly gateway-scoped headers I'll lean on the AuthPolicy / HTTPRoute filter approach rather than expanding the strip set. Good learning, thanks for your guidance and reviewing it with clarity.

Looking forward to finding more interesting ways I can help.

Thank you

PR #1180, opened 2026-06-22
 fix missing params for upstream.MCPNamein filtered_tools_handler

aniketpandey05:
## What does this PR do?

Fixes incorrect logging in filtered_tools_handler.go by invoking upstream.MCPName() instead of referencing the method value. This ensures the actual upstream MCP server name is included in log messages, improving observability and debugging.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed incorrect method invocation in debug logging for tool filtering, improving log accuracy for troubleshooting and debugging.

### DIFF
--- a/internal/broker/filtered_tools_handler.go
+++ b/internal/broker/filtered_tools_handler.go
@@ -192,7 +192,7 @@ func (broker *mcpBrokerImpl) filterToolsByServerMap(allowedTools map[string][]st
 		}
 		tools := upstream.GetManagedTools()
 		if tools == nil {
-			broker.logger.Debug("no tools registered for upstream server", "server", upstream.MCPName)
+			broker.logger.Debug("no tools registered for upstream server", "server", upstream.MCPName())
 			continue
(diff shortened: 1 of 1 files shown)


aniketpandey05:
@coderabbitai review

aniketpandey05:
@david-martin @jasonmadigan @maleck13 this one can be merged too it fixes the missing ()

Aman-Cool:
@rogueslasher, Confirmed.., `upstream.MCPName` was passing the method value instead of calling it, so the line logged a func, not the name. Grepped the tree too: this is the only instance left, everything else already calls `MCPName()`.

Optional: you can actually drop the call here. `serverName` is already in scope from the loop above, and since `findServerByName` only returns the upstream whose `MCPName() == serverName`, the two are equal.., so `"server", serverName` does the same thing without the call, and matches the Error log just above. Fine to leave as-is though.

maleck13 (approved):

PR #1182, opened 2026-06-23
fix(e2e): add missing CreateAuthorizedPromptsJWT helper to release branch

jasonmadigan:
## Summary

Cherry-pick of hairpin TLS fixes (#1159) brought in test code referencing `CreateAuthorizedPromptsJWT`, but the function was added on main in #1118 and never cherry-picked to `release-0.7.0`. This caused a compile error: `./happy_path_test.go:1122:20: undefined: CreateAuthorizedPromptsJWT`

**Fixes:**
- Adds `CreateAuthorizedPromptsJWT` and refactors `CreateAuthorizedCapabilitiesJWT` to share a `createCapabilitiesJWT` helper, matching main
- Adds an e2e compile check step (`go build -tags e2e ./tests/e2e/...`) to the existing unit test job in `tests.yaml` -- catches this class of error on release branch PRs without needing a cluster

Companion PR for main: #1183

PR #1183, opened 2026-06-23
ci: add e2e compile check to test workflow

jasonmadigan:
## Summary

e2e tests use `//go:build e2e`, so `go test ./...` (unit tests) never compiles them. `e2e.yaml` only triggers on PRs to `main`, not `release-*` branches. Result: cherry-picks to release branches get zero e2e coverage -- not even a compile check. This caused a compile error to ship in v0.7.1 (`undefined: CreateAuthorizedPromptsJWT`).

Adds an e2e compile step (`go build -tags e2e ./tests/e2e/...`) to the existing unit test job in `tests.yaml`, which already triggers on `main` and `release-*`. No separate job needed -- just an extra step after `make test-unit`. Fast, no cluster required, catches symbol and import errors.

Whether to extend full e2e runs to release branches is a separate decision.

Companion fix for release-0.7.0: #1182

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -45,3 +45,5 @@ jobs:
       - name: Run make unit test
         run: |
           make test-unit
+      - name: Verify e2e tests compile
+        run: go build -tags e2e ./tests/e2e/...


jasonmadigan:
cc @trepel

Patryk-Stefanski (approved):

PR #1187, opened 2026-06-23
remove controller->broker status polling for MCPServerRegistration

### LINKED ISSUE #629: Tech Debt - Improve how checking the status of the MCPServers with the gateway is handled

jasonmadigan:
Closes #629

## Summary

Rips out the controller-to-broker status polling. The controller was setting MCPServerRegistration Ready only after HTTP-polling the broker's `/status` endpoint, which was unreliable: tool counts went stale, protocol validation lagged behind reality. Now the controller sets Ready once the config secret is written. Runtime status lives in the broker's `/status` endpoint where it belongs.

Removing the polling exposed two pre-existing bugs that were hiding behind the polling delay:

**ext_proc `Value` vs `RawValue`**: Since Envoy 1.27, `send_header_raw_value` is enabled by default. Envoy reads `RawValue` (bytes) and ignores `Value` (string). The router's `WithImmediateResponse` and `WithImmediateJSONRPCResponse` were using `Value`, producing empty Content-Type and mcp-session-id headers on all error responses. The `HeadersBuilder` used by non-error paths already used `RawValue` correctly. This was latent because the old polling delay meant tools were always registered by the time a call arrived, so the error paths never actually ran in practice.

**Redundant config writes**: `UpsertMCPServer` wrote the Secret on every reconcile even when content was unchanged, bumping `resourceVersion` each time. Each bump triggered kubelet volume sync and broker config reload, which could restart MCPManagers and temporarily empty the tool registry. Added a `ConfigChanged` guard to skip no-op writes, and fixed the guard to include URL comparison -- without it, changes to `spec.path` or backend port would silently produce stale config.

## What changed

**Controller** (-280 lines):
- Deleted `ServerValidator` (HTTP polling to broker)
- Simplified `updateStatus`: removed `toolCount` param, Ready on config write
- Removed `DiscoveredTools` from CRD status, regenerated CRDs

**Router** (bug fix):
- `WithImmediateResponse`/`WithImmediateJSONRPCResponse`: `Value` to `RawValue`
- Error-path `mcp-session-id` headers: `Value` to `RawValue`
- Added `Content-Type: text/plain` to `WithImmediateResponse`

**Config writer**:
- `UpsertMCPServer` skips write when server config unchanged
- `ConfigChanged` includes URL to prevent stale config on path/port changes

**E2e tests**:
- Protocol validation, conflict, TLS negative, unavailable server tests rewritten for new semantics (controller no longer reports broker-side errors)
- Tool count verifiers removed

**Conformance workflow**:
- Replaced `sleep 10` with broker `/status` polling to wait for tool discovery after kubelet volume sync

**Docs**: removed `discoveredTools` from API ref, clarified status ownership in design doc

## Test evidence

- `make test-unit`: pass
- `make test-controller-integration`: 41/41
- CI e2e: 51/51
- CI conformance: 10/10

## Summary by CodeRabbit

* **New Features**
  * Registration status now highlights category instead of tool count.
  * Tool discovery is now checked asynchronously after the gateway becomes ready.

* **Bug Fixes**
  * Improved readiness handling so registrations report ready as soon as configuration is written.
  * Reduced unnecessary updates when server configuration hasn’t changed.
  * Immediate error responses now return clearer content-type headers.

* **Documentation**
  * Updated guides and reference docs to match the new readiness and status behavior.
  * Clarified where to check tool availability and what status information means.


malladinagarjuna2:
hey @jasonmadigan sir   rather than using portforwarding in confrmance tests why can't we use  kubectl get --raw ?

jasonmadigan:
@jasonmadigan done

jasonmadigan:
@malladinagarjuna2 port forwarding isn't used. this is a draft PR, there was a port forward while I was debugging something

david-martin:
👀

david-martin (changes requested):
The ConfigChanged thing looks like a good catch, and possible source of timing based bugs.

Not a blocker, but could the /status endpoint be helpful in reducing test time instead of Eventuallys, or not much time gained overall?

A few docs flagged as needing updates:
  docs/guides/register-mcp-servers.md and docs/guides/external-mcp-server.md still show a TOOLS column in
  kubectl get mcpsr output, and register-mcp-servers.md says "the broker needs a moment to connect and
  discover tools and prompts" when describing what Ready means. Both need updating for the new semantics.

jasonmadigan:
>The ConfigChanged thing looks like a good catch, and possible source of timing based bugs.

very likely yes

> Not a blocker, but could the /status endpoint be helpful in reducing test time instead of Eventuallys, or not much time gained overall?

I went back and forth on this a few times. ~~Checking via the ep is probably better, but I was wary of e2e tests interacting with APIs under the covers. on reflection though, that ep is public, and with this change we're recommending folks use it to debug, so having extra test coverage for it via this is no bad thing? I've amended.~~ stuck with the existing checks, as I didn't really want add another port-forward...

pushed updates for the docs too, good catches

maleck13:
👀

maleck13 (approved):
Changes look good. Nice one @jasonmadigan

david-martin (approved):

PR #1201, opened 2026-06-26
fix: use BUFFERED request body mode for reliable ClearRouteCache

maleck13:
STREAMED body mode causes Envoy to silently ignore header mutations and ClearRouteCache in the ext_proc body phase. This led to intermittent (~15-25%) routing failures where hairpin initialize requests were handled by the broker instead of the backend MCP server.

BUFFERED mode guarantees the route cache clear takes effect before Envoy selects an upstream. Limited to 1MB request body (Envoy default) which is sufficient for MCP JSON-RPC payloads.

Also adds x-client-id header to broker and lazyinit requests, and improves server2 test server logging.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved request handling so routing mutations are applied more reliably, reducing intermittent `tools/call` routing failures.
  * Added default client identifiers to outbound MCP requests for more consistent tracking and handling.

* **Documentation**
  * Expanded routing guidance with details on request body buffering and its impact on routing behavior and request size limits.

* **Chores**
  * Enhanced server-side logging for MCP requests and responses to make session and client activity easier to trace.

### DIFF
--- a/config/istio/envoyfilter.yaml
+++ b/config/istio/envoyfilter.yaml
@@ -32,7 +32,7 @@ spec:
             processing_mode:
               request_header_mode: 'SEND'
               response_header_mode: 'SEND'
-              request_body_mode: 'STREAMED'
+              request_body_mode: 'BUFFERED'
               response_body_mode: 'NONE'
               request_trailer_mode: 'SKIP'
               response_trailer_mode: 'SKIP'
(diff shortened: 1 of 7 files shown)


jasonmadigan (approved):
lgtm. the docs CI failure looks like something that was broken on main already - found because router.md was touched

PR #1214, opened 2026-06-29
fix: add /status route to gateway HTTPRoute and fix conformance CI poll

### LINKED ISSUE #1213: conformance CI readiness poll reads totalTools from wrong server
## what broke

Conformance CI has failed on main since #975. The broker discovers the conformance server tools successfully, but the CI readiness check times out.

## root cause

The conformance workflow broker readiness poll at `.github/workflows/conformance.yaml:119` grabs the **first** `totalTools` from the `/status` JSON response:

```
TOOLS=$(curl --max-time 5 -s "$STATUS_URL" | grep -o '"totalTools":[0-9]*' | head -1 | cut -d: -f2)
```

`test-user-specific-server` is registered before `conformance-server` and reports `totalTools: 0` (tools are fetched per-user, not at registration time). If it appears first in the JSON, `head -1` returns 0 forever, even though conformance-server has 14 tools and is fully healthy.

Evidence from failed main run 28386287848:
- gateway route active at 16:22:36
- controller wrote conformance server config at 16:22:34
- broker fetched 14 conformance tools at 16:23:38
- CI still timed out at 16:24:40
- no conflict errors in the logs

## fix

Use `jq` to target the conformance server specifically rather than taking the first `totalTools` from the response. See #1214.

jasonmadigan:
Closes #1213

## Summary

Two fixes for conformance CI broken since #975:

1. **Add `/status` route to the controller-generated HTTPRoute.** #975 removed the static HTTPRoute which had a catch-all `PathPrefix /`. The broker `/status` endpoint is no longer reachable through the gateway. Adds a `/status` rule (no filters, read-only health endpoint) to `buildGatewayHTTPRoute`.

2. **Fix the conformance CI tools poll.** The `grep | head -1` approach grabbed `totalTools` from whichever server appeared first in the `/status` JSON. `test-user-specific-server` reports `totalTools: 0` (per-user fetch) and can appear before `conformance-server`. Replaced with `jq` targeting the conformance server specifically. Also fixed the `HTTP_STATUS: 200000` bug (SSE endpoint causes curl timeout with non-zero exit, which overwrote the valid 200).

## Test evidence

- `make test-unit`: pass
- `make test-controller-integration`: 41/41
- `make lint`: clean

### DIFF
--- a/.github/workflows/conformance.yaml
+++ b/.github/workflows/conformance.yaml
@@ -98,7 +98,8 @@ jobs:
           echo "Waiting for Gateway to wire up the MCP routes and broker to fetch tools..."
           # wait for the gateway route to be active (non-502/503)
           for i in {1..30}; do
-            HTTP_STATUS=$(curl --max-time 2 -s -o /dev/null -w "%{http_code}" ${{ env.MCP_URL }} || echo "000")
+            HTTP_STATUS=$(curl --max-time 2 -s -o /dev/null -w "%{http_code}" "${{ env.MCP_URL }}" 2>/dev/null || true)
+            HTTP_STATUS=${HTTP_STATUS:-000}
             if [[ "$HTTP_STATUS" != "000" && "$HTTP_STATUS" != "0" && "$HTTP_STATUS" != "502" && "$HTTP_STATUS" != "503" ]]; then
               echo "Gateway route is active! HTTP Status: $HTTP_STATUS"
               break
@@ -116,7 +117,7 @@ jobs:
           echo "Waiting for broker to discover conformance server tools..."
           STATUS_URL="${MCP_URL%/mcp}/status"
           for i in {1..60}; do
-            TOOLS=$(curl --max-time 5 -s "$STATUS_URL" | grep -o '"totalTools":[0-9]*' | head -1 | cut -d: -f2)
+            TOOLS=$(curl --max-time 5 -s "$STATUS_URL" | jq '[.servers[] | select(.name | contains("conformance")) | .totalTools] | add // 0')
             if [ -n "$TOOLS" ] && [ "$TOOLS" -gt 0 ] 2>/dev/null; then
               echo "Broker has $TOOLS tools from conformance server."
               break
--- a/internal/controller/broker_router.go
+++ b/internal/controller/broker_router.go
@@ -650,6 +650,7 @@ func (r *MCPGatewayExtensionReconciler) buildGatewayHTTPRoute(mcpExt *mcpv1alpha
 	pathType := gatewayv1.PathMatchPathPrefix
 	mcpPath := "/mcp"
 	wellKnownPath := "/.well-known/oauth-protected-resource"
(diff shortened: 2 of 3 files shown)


Patryk-Stefanski:
The fix here is correct for `MCPGatewayExtension`, but `MCPServerRegistration.updateStatus` (`internal/controller/mcpserverregistration_controller.go:608`) has the same stale-object bug and is in the critical path for this failure.

That function does a bare `r.Status().Update(ctx, mcpsr)` against the in-memory object passed in by the caller. The call sites catch `IsConflict` and requeue, but that still triggers reconciler-level backoff — the same spiral that delayed things here. In the failing CI run the conformance-server registration was reconciling repeatedly before its config finally landed in the secret at 13:16:31, leaving the broker only ~2 minutes to connect before the 120s probe expired.

Since `MCPServerRegistration` Ready is what gates the config secret write (and therefore broker tool discovery), it would be worth applying the same `retry.RetryOnConflict` + fresh `Get` pattern there too, otherwise the fix is only half-complete for the conformance failure path.

jasonmadigan:
@Patryk-Stefanski changed this quite a bit, wasn't convinced by 🤖 that the bug was what it thought. I believe this current PR/issue are a proper fix.

Patryk-Stefanski:
Thanks for the correction — the evidence in the updated description makes it clear. The probe was reading  from  which appears first in the JSON, while the conformance server had already loaded its 14 tools. The -based fix is the right call. The  point I raised is a separate issue and not what caused these failures — will open a separate issue for that if it warrants one.

Patryk-Stefanski (dismissed):

Patryk-Stefanski:
Still failing after the jq fix. Found why — the `/status` probe URL is wrong.

`STATUS_URL="${MCP_URL%/mcp}/status"` strips `/mcp` from `http://mcp.127-0-0-1.sslip.io:8001/mcp` to give `http://mcp.127-0-0-1.sslip.io:8001/status`. That hits **Envoy on port 8001**, which has no route for `/status` (the gateway HTTPRoute only covers `/mcp` and `/.well-known/`). So `jq` gets a 404 body, parses nothing, and returns 0 — every iteration, forever.

From this run's broker logs (17:36:22):
```
level=DEBUG msg="updating gateway tools" ... adding=14 removing=0
```
The broker had 14 tools from 17:36:22, but the probe never saw them because it was talking to Envoy, not the broker.

`/status` is served by the broker on port 8080 and is never proxied through the gateway. The fix needs to reach the broker directly, e.g.:

```bash
STATUS_URL="http://$(kubectl get svc mcp-gateway -n mcp-system -o jsonpath='{.spec.clusterIP}'):8080/status"
```

Or alternatively use `kubectl port-forward` and poll over that, or add `/status` to the gateway HTTPRoute rules.

jasonmadigan:
@jasonmadigan conformance green. all CI passing.

Patryk-Stefanski (approved):

Issue #1111, opened 2026-06-09
Switch from mark3labs/mcp-go to official MCP Go SDK (investigate & impl)


david-martin:
We currently use mark3labs/mcp-go. The official Go SDK (https://github.com/modelcontextprotocol/go-sdk) is now maintained by the MCP team + Google, with full spec compliance.

The official SDK supports streamable HTTP, notifications, and elicitations. APIs are similar but incompatible — migration is straightforward (GitHub MCP Server already migrated: github/github-mcp-server#1184).

Investigate (timebox: 2 hours):
- Feature parity with our mark3labs usage across broker and router
- Migration effort — identify all import sites and API translation needed
- Gaps or regressions vs mark3labs

Outcome: pros/cons and a go/no-go recommendation for switching in 0.8.

david-martin:
Raising priority. The [upcoming MCP spec](https://blog.modelcontextprotocol.io/posts/2026-07-28-release-candidate/) (2026-07-28) removes protocol sessions, changes the handshake model, adds required headers, and deprecates several features. The official Go SDK will track these changes as a Tier 1 SDK. Staying on mark3labs/mcp-go means either waiting for community support of the new spec or forking.

This investigation is now on the critical path for spec migration (#256).

PR #1215, opened 2026-06-30
SDK migration from mark3labs/mcp-go to official go-sdk

maleck13:
Re: #1111

Adds design doc for migrating the gateway from github.com/mark3labs/mcp-go to github.com/modelcontextprotocol/go-sdk. Covers
  broker server/client migration, router client migration, hooks-to-middleware conversion, performance implications, and security
  considerations.

  Migrating to the official Tier 1 SDK unblocks 2026-07-28 protocol adoption — the official SDK (v1.7.0) supports both 2025-11-25
  and 2026-07-28 with automatic version negotiation.

### DIFF
--- a/docs/design/sdk-migration/sdk-migration-design.md
+++ b/docs/design/sdk-migration/sdk-migration-design.md
@@ -8,7 +8,7 @@ The official `github.com/modelcontextprotocol/go-sdk` (v1.6.1) is a Tier 1 SDK m
 
 ## Summary
 
-Full replacement of `github.com/mark3labs/mcp-go` with `github.com/modelcontextprotocol/go-sdk/mcp` across broker, router, upstream client, and test servers. No coexistence period. The migration is structured in three phases: types and constants, server/broker, client/upstream. Performance must not regress in the broker's tool/call serving and tools/list aggregation paths.
+Full replacement of `github.com/mark3labs/mcp-go` with `github.com/modelcontextprotocol/go-sdk/mcp` across broker, router, upstream client, and test servers. The migration is structured in two phases: broker first (server, upstream client, hooks→middleware, test servers), then router. The broker is the highest-risk component — it touches every request path — so it migrates first to surface integration issues early. The router migration follows once the broker is stable. Performance must not regress in the broker's tool/call serving and tools/list aggregation paths.
 
 ## Goals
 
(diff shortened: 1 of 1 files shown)


david-martin (commented):
Zero perf regression is reasonable for our own code paths. If something regresses because of how the SDK itself handles things (JSON parsing, middleware overhead), that shouldn't block the move. Assess severity, discuss upstream if needed. Perf tests should also catch connection lifecycle issues around the `Client`/`ClientSession` separation and latent open connections.

---

Prefer this done wholesale in a single PR. Splitting it risks having both SDKs in the tree at the same time, no matter how short the timeline. Trust the tests, all or nothing. GitHub landed theirs the same way ([github/github-mcp-server#1428](https://github.com/github/github-mcp-server/pull/1428), +11k/-8.5k, single merge).

---

GitHub also wrote a [Copilot sub-agent template](https://github.com/github/github-mcp-server/blob/main/.github/agents/go-sdk-tool-migrator.md) for the mechanical renames. Worth referencing in the design doc, and we could build something similar.

---

On session IDs: the SDK's `Stateless` mode ([go-sdk#307](https://github.com/modelcontextprotocol/go-sdk/pull/307)) is one option. Gets the SDK's session tracking out of the way, lets us own the full JWT lifecycle. GitHub used this for their remote server. Alternatively, leave sessions on and handle JWT validation in a middleware. Either way, [go-sdk#148](https://github.com/modelcontextprotocol/go-sdk/issues/148) is being closed without a `SessionStore`, so no SDK-level solution is coming.

docs/design/sdk-migration/sdk-migration-design.md
```diff
@@ -0,0 +1,135 @@
+# SDK Migration: mark3labs/mcp-go → modelcontextprotocol/go-sdk
+
+## Problem
+
+The gateway depends on `github.com/mark3labs/mcp-go` (v0.53.0) for JSON-RPC types, client/server SDK, and HTTP transport. This is a community SDK with no `2026-07-28` tracking issues and a release cadence decoupled from spec milestones.
+
+The official `github.com/modelcontextprotocol/go-sdk` (v1.6.1) is a Tier 1 SDK maintained by the MCP organization. It ships partial `2026-07-28` support (`Mcp-Method`/`Mcp-Name` headers, extensions framework, error code standardization, multi-round-trip requests) with open tracking issues for the remaining SEPs (stateless protocol SEP-2575, TTL for list results SEP-2549). Staying on mark3labs blocks `2026-07-28` adoption.
```
jasonmadigan, line 7: I see 1.7.0 pre1 was made available last week (full support for `2026-07-28`) https://github.com/modelcontextprotocol/go-sdk/releases

wondering if we should:

1) align with 1.6.1, and update to 1.7.0 when non pre-release. 
2) Update to 1.7.0 pre, keep this upstream, although to do this we'd want a firm understanding of when it's likely 1.7.0 would become latest upstream

1 means two-updates, one now, one "later" (although I guess the the 2nd one is likely to be small effort?)

2 means potentially using a pre-release lib early, with all that entails (bugs), but probably gets us into the right long-term "shape" earlier?

jasonmadigan: data point: there was about 7-8d between 1.6.0 pre and 1.6.0 final, so 1.7.0 may be imminent anyway making this moot

jasonmadigan:
> agreed that we'd make it goal of next sprint to switch to new sdk, alpha one, and also then add alpha support for new spec. sdk still first though

PR #1218, opened 2026-06-30
feat: migrate from mark3labs/mcp-go to official MCP Go SDK

### LINKED ISSUE #1111: Switch from mark3labs/mcp-go to official MCP Go SDK (investigate & impl)

### LINKED ISSUE #1242: Flaky e2e: UserSpecificList intermittent 503 failures
`[Security,UserSpecificList] internal headers not forwarded to upstream` and `[UserSpecificList] tool call routing works for user-specific tools` fail intermittently with `503: no healthy upstream`. Reproduces on both PR branches and main.

**Recent failures:**
- main: https://github.com/Kuadrant/mcp-gateway/actions/runs/28652840272
- main: https://github.com/Kuadrant/mcp-gateway/actions/runs/28652496969
- main: https://github.com/Kuadrant/mcp-gateway/actions/runs/28660909716
- PR #1224: https://github.com/Kuadrant/mcp-gateway/actions/runs/28655530680
- PR #1224: https://github.com/Kuadrant/mcp-gateway/actions/runs/28653218883
- PR #1246: https://github.com/Kuadrant/mcp-gateway/actions/runs/28661001717

**Likely cause: cross-process deployment scaling**

CI runs with `E2E_PROCS=4`. The `MCP Gateway User-Specific Tool Lists` Describe has no `Serial` or `Ordered` decorator, so Ginkgo distributes its `It` blocks across processes.

One of those tests, `user-specific server down does not break tools/list` (`user_specific_list_test.go:239`), scales `mcp-test-user-specific-server` to 0 replicas to verify graceful degradation. While that test holds the server at 0, another Ginkgo process could be running `tool call routing` or `internal headers`, both of which call tools against the same `mcp-test-user-specific-server` backend. With no healthy pods, Envoy returns 503.

The `BeforeEach` at line 23 scales the server back to 1, but that only runs within a single Ginkgo process and does not synchronize across processes. It also does not wait for pod readiness (`ScaleDeployment` fires `kubectl scale` and returns immediately).

This would explain why the failure is intermittent: it depends on Ginkgo scheduling the scale-down test in a different process than the tool-call tests, and on the timing overlapping.

**Possible fixes:**
- Add `Serial` to the scale-down test so it never runs in parallel with other tests
- Add `Ordered, Serial` to the whole `Describe` so the suite runs sequentially in a single process
- Move the scale-down test to its own `Describe` with `Serial`, leaving the rest parallel

jasonmadigan:
## Summary

- Replaces `mark3labs/mcp-go` with `modelcontextprotocol/go-sdk v1.7.0-pre.1` across broker, router, upstream, transport, and all tests
- Zero mark3labs references remain; outward behaviour is preserved -- clients cannot tell the library changed
- Standalone test servers (`tests/servers/*`) stay on SDK v1.4.1 as a backwards-compatibility canary

Design doc: #1215

## Compatibility layer (`http_compat.go`)

The official SDK's HTTP surface differs from mark3labs in framing, error shapes, and edge-case handling. Rather than pushing breaking changes to every client, a boundary shim restores mark3labs' exact observable behaviour: plain `application/json` POST responses, mark3labs error codes for unknown methods and malformed payloads, capability-honest `-32601` for unsupported domains, `*bool` tool-annotation fidelity, unpaginated lists, batch rejection, and lenient Accept handling. Each shim has a regression test asserting mark3labs' exact status code, content-type, and body bytes.

Two accepted deltas:

1. **Duplicate in-flight JSON-RPC request IDs** are now rejected per spec. mark3labs tolerated them, but remapping IDs at the boundary risks response cross-wiring for clients already violating the protocol.
2. **Notification failure key in `select_tools`** responses is no longer possible. SDK notification dispatch is async; failure was already rare and advisory-only.

## Session resurrection (`session_resurrection.go`)

mark3labs validated sessions statelessly (JWT only). The official SDK keeps a pod-local session table and 404s IDs it has never seen, which would break sessions across replicas and restarts and orphan Redis-keyed state. The resurrection handler intercepts the 404: if the ID is a valid gateway JWT, it connects a fresh server session with the same ID and synthesised initialize state, then replays the request. Invalid or expired JWTs still 404. Idle table entries are evicted after 30 minutes (client-invisible since they resurrect on demand).

## Two-phase config reload

`OnConfigChange` now deregisters stale managers under the lock, then stops them *outside* the lock before starting replacements. Stopping under the lock deadlocked with `IsReady()` and middleware readers (Stop blocks on session teardown / network I/O). Starting replacements before the old manager finishes risks `removeAllTools` deleting the replacement's tools by name with no ownership check.

## Other notable changes

- **Hooks to middleware**: `BeforeAny`/`AfterListTools`/`OnSuccess`/`OnError` hooks replaced with `mcp.Middleware` (tracing + filtering)
- **Router**: `client.Client` replaced with `mcp.Client` + `mcp.ClientSession` two-step connect; hairpin init skips the standalone SSE stream and short-circuits the SDK's `server/discover` probe in-process
- **Upstream**: `HeaderRoundTripper` consolidated into `internal/transport/`; tools with nil/invalid schemas rejected at validation instead of panicking the SDK; `roots.listChanged` capability and pre-connect notification registration preserved
- **Session reuse**: per-user upstream session pool for `userSpecificList` servers, cached across `tools/list`, evicted on error with single retry, purged on session end
- **Per-session notifications**: `tools/list_changed` targeted via expiring per-session claims + `SendingMiddleware` filter, delivered on the standalone GET stream with mark3labs frame bytes
- **Transport**: consolidated into `internal/transport/transport.go` (roundtripper, discover, SSE and peek helpers)
- **Benchmarks**: new suites for router and broker hot paths

## Performance

Microbenchmarks (benchstat, interleaved, n=10): broker filter internals geomean ~32% faster, ~68% less memory, ~15% fewer allocs. Router hot-path parsing unchanged (allocations bit-identical, time within noise). Below the compat layer: list-response reshape adds ~226us/109KiB per 50-tool `tools/list` (~20us per 10-prompt `prompts/list`); `tools/call` pays only ~0.9us envelope parse.

E2E (CI k6, same runners): tool-call avg 2.91ms / p95 7ms including compat layer, inside main's own band. Regression gate passes.

## Test evidence

- `go build ./...`, `go vet ./...`, `golangci-lint run` clean
- `make test-unit` all pass, `go test -race ./internal/broker/...` clean
- Compat layer: per-shim regression tests asserting mark3labs' exact bytes
- E2E: 71/71 locally against unmodified SDK v1.4.1 test-server images; `e2e-tests` and `conformance-tests` green in CI
- Session resurrection verified live: initialize, rollout restart, same-session `tools/list` 200 on new pod; invalid JWT 404
- Controller integration 41/41

**85 files changed, +6773/-2007**

Closes #1111

## Summary by CodeRabbit

* **New Features**
  * Enhanced MCP compatibility across tools, prompts, and notifications, including tool-hint propagation.
  * Improved session continuity via session resurrection and idle handling.
  * More precise `tools/list_changed` notification delivery with targeted updates and controlled broadcast behavior.
* **Bug Fixes**
  * Fixed tool/prompt filtering to avoid mutating shared in-memory objects during repeated or concurrent requests.
  * Hardened shutdown and invalid-session behavior for more reliable teardown and responses.
* **Chores**
  * Updated MCP SDK integration and expanded routing/compatibility and concurrency test coverage.


docs/release-notes/mcp-go-sdk-migration.md
```diff
@@ -0,0 +1,25 @@
...
+1. Find affected policies:
+   ```bash
+   kubectl get authpolicies -A -o yaml | grep -n "readOnly=unspecified\|idempotent=unspecified"
+   ```
+2. Replace `readOnly=unspecified` with `readOnly=false` and `idempotent=unspecified` with `idempotent=false` in the matching expressions, treating an absent hint as `false`.
+
+## Gateway sessions are pod-local; clients must re-initialise after a broker restart
+
+### What changed
+After a broker restart (upgrade, rollout, rescheduling), requests carrying a previous session ID now receive HTTP 404. This is the MCP specification's signal for the client to start a new session. Previously the broker silently re-initialised the session, hiding the restart from clients. Idle sessions are also now closed once the session JWT lifetime (`--session-length`, default 24h) passes without a request.
```
maleck13: even with Redis?

jasonmadigan:
> Config reload: two-phase manager replacement -- old managers stop before replacements start, so same-named tools cannot be deleted by a stale teardown

investigating this deeper, there's a few findings here I don't think are actually issues..

docs/release-notes/mcp-go-sdk-migration.md
jasonmadigan: i've not fully reviewed the agent output yet, this and a bunch of other things look wrong to me (e.g. test server updates make me sus)

jasonmadigan: fixed. restarts and multi-replica routing behave as they did on mark3labs lib. redis-keyed backend sessions/tokens re-attach.

maleck13:
👀 on this one today

internal/broker/http_compat.go
```diff
@@ -0,0 +1,763 @@
...
+	"net/http"
+	"strings"
+	"unicode"
+
+	"github.com/Kuadrant/mcp-gateway/internal/broker/upstream"
+	"github.com/Kuadrant/mcp-gateway/internal/transport"
+	"github.com/modelcontextprotocol/go-sdk/mcp"
+)
+
+// TODO: remove this compatibility layer when we adopt streamable-HTTP
```
jasonmadigan, line 19: @maleck13 like we chatted about, added a big TODO here noting that this should be removed over time

internal/broker/upstream/hints_test.go
```diff
@@ -0,0 +1,128 @@
...
+	"net/http/httptest"
+	"testing"
+	"time"
+
+	"github.com/Kuadrant/mcp-gateway/internal/config"
+	"github.com/modelcontextprotocol/go-sdk/mcp"
+	"github.com/stretchr/testify/require"
+)
+
+func boolPtr(v bool) *bool { return &v }
```
maleck13: we use `ptr.To(value)` quite a lot elsewhere. Might be better for consistency? Although this is just a test so 🤷

internal/broker/upstream/mcp.go
```diff
@@ -142,166 +185,183 @@ func (up *MCPServer) GetName() string {
...
-				},
-				Elicitation: &mcp.ElicitationCapability{},
-			},
-			ClientInfo: mcp.Implementation{
-				Name:    "mcp-broker",
-				Version: "0.0.1",
-			},
-		},
-	})
+	connectCtx, cancel := context.WithTimeout(ctx, 30*time.Second)
```
maleck13, line 245: 30s seems like a long timeout? 10s would seem more than enough?

internal/broker/broker.go
```diff
@@ -205,88 +248,175 @@ func NewBroker(logger *slog.Logger, opts ...Option) MCPBroker {
...
-
-	serverOpts := []server.ServerOption{
-		server.WithHooks(hooks),
-		server.WithToolCapabilities(true),
-		server.WithPromptCapabilities(true),
+	serverOpts := &mcp.ServerOptions{
+		HasPrompts: true,
+		// mark3labs never paginated list results; a page size no client can
+		// reach keeps every list a single page with no nextCursor.
+		PageSize:     unpaginatedPageSize,
```
maleck13, line 260: interesting. Wondering if there is a benefit in the gateway to use pagination at a certain limit?

internal/broker/broker.go
```diff
@@ -205,88 +248,175 @@ func NewBroker(logger *slog.Logger, opts ...Option) MCPBroker {
...
+	for _, man := range toStop {
+		m.logger.InfoContext(ctx, "stopping manager", "server", man.MCPName())
+		man.Stop()
+	}
+
+	// phase 3: start new/replacement managers and swap derived state
+	m.startManagers(ctx, servers, virtualServers)
+
+	m.logger.DebugContext(ctx, "Broker OnConfigChange done", "total servers", len(servers))
+}
```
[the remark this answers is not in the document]
maleck13: @jasonmadigan seems valid but I would say a low risk. No harm in fixing if low cost

internal/broker/broker.go
```diff
@@ -15,19 +15,25 @@ import (
...
+	"github.com/modelcontextprotocol/go-sdk/mcp"
 	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/trace"
 )
 
 var _ config.Observer = &mcpBrokerImpl{}
 
+// unpaginatedPageSize disables list pagination in practice. not math.MaxInt:
+// the SDK's paginateList computes pageSize+1, which must not overflow.
+const unpaginatedPageSize = 1 << 30
```
maleck13, line 29: this made me wonder if we should consider a pagination option in the gateway

internal/broker/user_specific_tools.go
```diff
@@ -204,12 +192,100 @@ func (broker *mcpBrokerImpl) doFetchTools(ctx context.Context, srv userSpecificS
...
+	}, nil)
+
+	session, err := mcpClient.Connect(ctx, t, nil)
+	if err != nil {
+		return nil, err
+	}
+
+	cached.session = session
+	// if another goroutine raced us, close ours and use theirs with our
+	// headers, which are at least as fresh
```
maleck13, line 243: This got highlighted as potential race that we could prob solve with a simple mutex lock on the session cache miss path

jasonmadigan:
Last commit fixes flake from #1242

internal/broker/gateway_server.go
```diff
@@ -0,0 +1,217 @@
...
+
+var _ upstream.ToolsAdderDeleter = &gatewayServer{}
+var _ upstream.PromptsAdderDeleter = &gatewayServer{}
+
+func newGatewayServer(server *mcp.Server) *gatewayServer {
+	return &gatewayServer{
+		server:        server,
+		tools:         make(map[string]*upstream.GatewayTool),
+		prompts:       make(map[string]*upstream.GatewayPrompt),
+		notifyTargets: make(map[string]time.Time),
```
maleck13, line 50: I think we should consider updating the notification docs around this targeting mechanism and ensure our technical docs are up to date with the design implemented in this PR
- targeted sessions and how it works
- sentinel tool and why that is needed 
Or maybe something that ties the different pieces together? 
- how the compat layer, resurrection handler, and notification targeting form a full handler chain, and why each layer exists at that position?

internal/broker/upstream/mcp.go
jasonmadigan: the sdk seems to use 60s as a default, so this is reduced a bit anyway

internal/broker/broker.go
jasonmadigan: i think so, worthy of a follow up issue?

internal/broker/gateway_server.go
jasonmadigan: agree, added some updates to notifications.md and session-mgmt.md

internal/broker/broker.go
maleck13: yes lets do that 👍

internal/transport/transport.go
```diff
@@ -0,0 +1,149 @@
...
+// DiscoverShortCircuit answers the SDK client's SEP-2575 server/discover
+// probe in-process with the method-not-found error a mark3labs-era gateway
+// produced, so the client falls straight back to the legacy initialize
+// handshake with zero network round trips for the probe.
+type DiscoverShortCircuit struct {
+	Base http.RoundTripper
+}
+
+// RoundTrip implements http.RoundTripper.
+func (d *DiscoverShortCircuit) RoundTrip(req *http.Request) (*http.Response, error) {
```
maleck13, line 99: These round trippers are useful but add complexity. Wondering how best to communicate that we have these to short circuit certain requests that are not needed?

maleck13 (approved):
I have given this a good read over. I understand the changes. Although a lot to take in and digest, there are minimal changes to the tests that show we are consistent with what came before. In addition from probing with AI there are several clear improvements also. 

+1

PR #1220, opened 2026-06-30
fix(broker): keep user-specific tools when the x-mcp-authorized filter runs

### LINKED ISSUE #1219: broker: userSpecificList tools are dropped by the x-mcp-authorized capability filter
## What

When a server is registered with `userSpecificList` and tool-level authorization
is also in use (the signed `x-mcp-authorized` header), the user-specific tools
never make it into the `tools/list` response. They are silently dropped, so the
client cannot see or call them.

The two features work fine on their own. The problem only shows up when they are
combined, which the guide explicitly says is supported.

## Where

`internal/broker/filtered_tools_handler.go`

`FilterTools` runs the `x-mcp-authorized` filter first (`applyAuthorizedCapabilitiesFilter`).
When the header is present, that calls `filterToolsByServerMap`, which rebuilds the
response from each upstream's cached `GetManagedTools()` instead of filtering the
list it was handed.

userSpecificList servers cache no managed tools — `MCPManager` returns early for
them (`internal/broker/upstream/manager.go:353`, "tools fetched per-user"). Their
tools are fetched per request and merged into the result earlier by
`FetchUserSpecificTools`. Because `filterToolsByServerMap` ignores that merged
list and reads from the (empty) managed-tool cache, the user-specific tools are
discarded.

The virtual-server filter (`applyVirtualServerFilter`) does not have this problem
— it filters the merged slice directly.

## Impact

Any deployment that combines `userSpecificList` with `x-mcp-authorized` tool
authorization loses all of its user-specific tools from `tools/list`. The drop is
silent — no error, the tools are just absent. The guide
(`docs/guides/user-specific-tools.md`) documents this combination as supported:

> User-specific tools go through the same filter pipeline as cached tools:
> 1. Fetch ... 2. JWT filter (`x-mcp-authorized`): filters the merged list by user
> permissions ... You can combine `userSpecificList` with virtual servers and
> JWT-based authorization.

The code does not filter the merged list; it rebuilds it from the cache, so the
documented behaviour does not hold.

## Steps to reproduce

1. Register an MCPServerRegistration with `userSpecificList: Enabled` and a `prefix`.
2. Configure tool-level authorization so the `x-mcp-authorized` header reaches the
   broker (allow at least one of that server's tools).
3. Call `tools/list` as an authorized user.

Expected: the user-specific tools allowed by the JWT appear in the response
(merged with the cached tools from standard servers).

Actual: the user-specific tools are missing; only cached tools from standard
servers come back.

## Expected behaviour

The `x-mcp-authorized` filter should filter the already-merged tool list (the same
list every other filter step operates on) rather than reconstructing it from the
managed-tool cache, so user-specific tools allowed by the JWT survive.

PRAteek-singHWY:
## What

`FilterTools` applies the `x-mcp-authorized` capability filter first. When the
header is present, `filterToolsByServerMap` rebuilt the response from each
upstream's cached `GetManagedTools()` rather than filtering the list it was
given.

userSpecificList servers cache no managed tools — `MCPManager` returns early for
them ("tools fetched per-user"), and their tools are merged into the result
per-request by `FetchUserSpecificTools` before filtering. Because the filter read
from the empty managed-tool cache instead of the merged list, every user-specific
tool was silently dropped whenever `x-mcp-authorized` was in play.

This makes the capability filter work on the already-merged slice — the same list
the virtual-server filter already operates on — so user-specific tools allowed by
the JWT are kept.

## Why

The guide documents combining `userSpecificList` with `x-mcp-authorized`
authorization as supported, and says the JWT step "filters the merged list". The
old code reconstructed the list from the managed-tool cache, so for user-specific
servers (which cache nothing) the documented behaviour did not hold and authorized
per-user tools disappeared from `tools/list` with no error.

## Changes

- `internal/broker/filtered_tools_handler.go`: `filterToolsByServerMap` now builds
  the set of allowed prefixed tool names from the JWT and filters the merged tool
  slice, instead of rebuilding from `GetManagedTools()`. Behaviour for standard
  (cached) servers is unchanged; user-specific tools are no longer dropped.
- `internal/broker/filtered_tools_handler_test.go`: add a case covering a
  user-specific server (no cached tools) alongside a standard one, asserting both
  survive the capability filter.

## Testing

- `make test-unit` passes.
- The new test fails on the previous behaviour (user-specific tool dropped) and
  passes with the fix.
- `gofmt` / `goimports` clean on the changed files.

## Notes

Bugfix, exempt from the issue-first requirement per `CONTRIBUTING.md`. Tracking
issue filed for visibility.

Fixes #1219

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved filtering of authorized tools during JWT-based access checks, ensuring the correct set is returned.
  * Fixed an issue where user-specific tools could be dropped when filtering combined tool sources (including cached tools).
  * Filtering now reliably preserves the expected tool names in the final results.

### DIFF
--- a/internal/broker/filtered_tools_handler.go
+++ b/internal/broker/filtered_tools_handler.go
@@ -125,7 +125,7 @@ func (broker *mcpBrokerImpl) applyAuthorizedCapabilitiesFilter(headers http.Head
 		return tools
 	}
 
-	return broker.filterToolsByServerMap(allowedTools)
+	return broker.filterToolsByServerMap(allowedTools, tools)
 }
 
 // parseAuthorizedCapabilitiesJWT validates and extracts allowed capabilities from the JWT header.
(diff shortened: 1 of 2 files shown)


PRAteek-singHWY:
Hi @david-martin sir,
I'll resolve this conflict shortly. and then we can get it meregd,
and Thank you for reviewing, Means a lot.

PRAteek-singHWY:
Hi @david-martin 

Rebased onto latest main and resolved the conflict from the MCP Go SDK migration - filterToolsByServerMap is ported to []*mcp.Tool and now filters the already-merged tool slice (bug still present on main, just under the new types). Build, unit, e2e, and conformance all green, including the user-specific-tools regression case. 

Ready for review whenever you get a chance.

Thank You.

david-martin (approved):
This is a nice find @PRAteek-singHWY 
Thanks for fixing this. The test is a welcome addition as well.

PR #1224, opened 2026-07-01
feat: add caCertBundleRef to MCPGatewayExtension

### LINKED ISSUE #1108: Configure TLS from MCPGatewayExtension
The broker currently requires `--gateway-ca-cert` to be set directly as a CLI flag. This should be configurable via `MCPGatewayExtension` instead, so TLS configuration lives in the CRD rather than requiring direct flag management.

There is a design doc covering part of this: `docs/design/gateway-ca-cert-bundle/`. The `caCertBundleRef` field on `MCPGatewayExtension` would replace the need for `--gateway-ca-cert`, and also solve the duplication problem when multiple upstream servers share the same CA.

Scope:
- Add `caCertBundleRef` to `MCPGatewayExtension` (per the existing design doc)
- Operator propagates the referenced CA bundle to the broker deployment

david-martin:
## What does this PR do?

Adds a `caCertBundleRef` field to `MCPGatewayExtension` that references a shared CA certificate bundle Secret. The broker loads this as a base trust pool for all upstream connections, eliminating duplication when many servers share the same private CA.

Fixes #1108

### Changes

- CRD: `caCertBundleRef` (optional) on `MCPGatewayExtensionSpec` with `CACertBundleReference` type (`name` + optional `key`, default `ca.crt`)
- Controller: validates referenced Secret (exists, labeled, valid PEM, ≤256 KiB), writes `gatewayCACertPEM` to config Secret, re-reconciles on Secret changes
- Broker: loads gateway CA bundle as base trust pool, detects changes and recreates all upstream managers
- Upstream: three-layer additive trust pool (system roots, gateway bundle, per-server CA)
- Config: `gatewayCACertPEM` field in `BrokerConfig` and `MCPServersConfig`, `WriteCACertBundle` on `ConfigWriterDeleter`
- Docs: updated guide, API reference, CLAUDE.md
- Tests: unit tests for controller validation and upstream TLS, 6 e2e test cases

## Pre-review checklist

- [ ] Checked the CodeRabbit walkthrough for "Possibly related issues" and confirmed the PR uses `Fixes` or `Closes` syntax for any it addresses
- [ ] Checked the CodeRabbit walkthrough for "Possibly related PRs" and confirmed this PR is not a duplicate

## Summary by CodeRabbit

* **New Features**
  * Added `caCertBundleRef` support on `MCPGatewayExtension` for a gateway-level CA bundle used as the base trust for upstream TLS.
  * Extended HTTPS end-to-end coverage for gateway CA rotation, and combined gateway + per-server trust.

* **Bug Fixes**
  * Trust is applied additively (gateway bundle plus optional per-server additions).
  * Improved behavior when the gateway bundle changes and readiness errors for missing/invalid CA bundle secrets.

* **Documentation**
  * Expanded guides and the `MCPGatewayExtension` reference with setup steps, examples, and troubleshooting.


tests/e2e/ca_cert_bundle_test.go
```diff
@@ -0,0 +1,431 @@
...
+
+		By("Verifying tools are absent (TLS handshake fails)")
+		// weak assertion: only proves tools don't appear, not why.
+		// stronger: hit broker /status and assert x509 error in Message field.
+		Consistently(func(g Gomega) {
+			toolsList, err := mcpGatewayClient.ListTools(ctx, mcpgo.ListToolsRequest{})
+			if err != nil {
+				return
+			}
+			g.Expect(verifyMCPServerRegistrationToolsPresent("cab_wrong_", toolsList)).To(BeFalse())
```
david-martin, line 267: Weak assertion: only proves tools are absent, not that TLS is the cause. Broker `/status` endpoint would let us assert the actual x509 error. Follow-up.

david-martin: @jasonmadigan i know you backed out the /status endpoint check in some tests recently.
Thoughts on this kind of test where i want to assert that a cert error has happened, and not something else that's causing no tools to be listed?

tests/e2e/custom_tls_test.go
```diff
@@ -49,25 +46,6 @@ var _ = Describe("Custom TLS Configuration", Ordered, func() {
 		mcpGatewayClient *NotifyingMCPClient
 	)
 
-	BeforeAll(func() {
```
david-martin, line 52: Note this being removed.
I think this is a future source of regression if there's infra issues bringing up cert manager or the tls test server, and those tests start getting ignored without any obvious signal.
They should always be there.

Patryk-Stefanski (commented):
Four issues surfaced during review — each commented inline below.

internal/controller/mcpserverregistration_controller.go
```diff
@@ -336,7 +341,7 @@ func (r *MCPReconciler) getTargetGatewaysFromParentRef(ctx context.Context, pare
 	return g, nil
 }
 
-func (r *MCPReconciler) buildMCPServerConfig(ctx context.Context, targetRoute *gatewayv1.HTTPRoute, mcpsr *mcpv1alpha1.MCPServerRegistration) (*config.MCPServer, error) {
+func (r *MCPReconciler) buildMCPServerConfig(ctx context.Context, targetRoute *gatewayv1.HTTPRoute, mcpsr *mcpv1alpha1.MCPServerRegistration, hasGatewayCACertBundle bool) (*config.MCPServer, error) {
 	if mcpsr.DeletionTimestamp != nil {
```
Patryk-Stefanski, line 345: **Bug: gateway CA bundle forces `https://` scheme upgrade on all servers, including plain-HTTP ones**

`hasGatewayCACertBundle` is folded across every `MCPGatewayExtension` on every valid gateway. If any peer extension has `CACertBundleRef != nil`, `buildMCPServerConfig` coerces *every* server's endpoint to `https://` — including plain internal `http://` services that have no TLS at all. The broker then attempts a TLS handshake against a plaintext server, fails, and that server's tools are permanently unavailable for as long as any co-hosted extension has a bundle set.

The scheme upgrade was designed to signal "this server speaks TLS" — a property that belongs to the individual `MCPServerRegistration` (via `caCertSecretRef`), not to the gateway. The gateway bundle is a trust pool configuration, not a protocol assertion.

Suggested fix: keep the gateway CA cert available to the trust pool, but remove the `|| hasGatewayCACertBundle` from the scheme-upgrade condition:

```go
// scheme upgrade only when the registration itself declares TLS
if mcpsr.Spec.CACertSecretRef != nil {
    // ... enforce https
}
```

The broker will still pick up the gateway bundle for TLS servers (those already coerced by their own `caCertSecretRef`); plain-HTTP servers are left alone.

internal/broker/broker.go
```diff
@@ -280,12 +283,20 @@ func (m *mcpBrokerImpl) OnConfigChange(ctx context.Context, conf *config.MCPServ
 	// concurrently replacing conf.Servers/VirtualServers under its own write lock.
 	servers := conf.ListServers()
 	virtualServers := conf.ListVirtualServers()
+	newGatewayCACert := conf.GetGatewayCACertPEM()
 
 	m.logger.DebugContext(ctx, "Broker OnConfigChange start", "Total managers for upstream mcp servers", len(m.mcpServers), "total servers", len(servers))
 	// unregister decommissioned servers
```
Patryk-Stefanski, line 289: **Bug: CA bundle rotation tears down all upstream managers, including HTTP-only servers**

When `gatewayCACertChanged` is true, the condition `gatewayCACertChanged || mcpServer.ConfigChanged(man.Config())` stops and recreates *every* manager unconditionally — including plain `http://` servers that have no TLS dependency on the gateway CA.

A routine CA rotation (e.g. certificate refresh before expiry) therefore causes a simultaneous reconnect storm across all registered servers. During re-initialisation each server goes through the full MCP handshake again, so all tool discovery and `tools/call` requests fail for the duration.

The teardown should be gated on whether the server actually uses TLS:

```go
serverUsesTLS := mcpServer.CACert != "" || newGatewayCACert != ""
if (gatewayCACertChanged && serverUsesTLS) || mcpServer.ConfigChanged(man.Config()) {
    man.Stop()
    delete(m.mcpServers, mcpServer.ID())
}
```

HTTP-only servers can keep their existing connections across a CA rotation; only TLS-using servers need to be rebuilt with the new trust pool.

internal/controller/ca_cert_bundle.go
```diff
@@ -0,0 +1,60 @@
...
+	}
+
+	if secret.Labels == nil || secret.Labels[ManagedSecretLabel] != ManagedSecretValue {
+		return newValidationError(mcpv1alpha1.ConditionReasonSecretInvalid,
+			fmt.Sprintf("CA bundle secret %s missing required label %s=%s", ref.Name, ManagedSecretLabel, ManagedSecretValue))
+	}
+
+	key := ref.Key
+	if key == "" {
+		key = "ca.crt"
```
Patryk-Stefanski, line 43: **Bug: `validateCACertPEM` accepts leaf certificates — silent failure at TLS handshake time**

`validateCACertPEM` checks that each PEM block parses as a valid `x509.Certificate` but does not check `cert.IsCA` or `cert.BasicConstraintsValid`. A leaf/end-entity certificate (e.g. the server's own TLS cert rather than its issuing CA) passes validation here, gets written to the config, and is loaded into the trust pool via `AppendCertsFromPEM`.

At TLS handshake time Go's chain builder ignores non-CA certs as trust anchors, so all upstream connections fail with `certificate signed by unknown authority`. The `MCPGatewayExtension` stays `Ready` and there's no indication the bundle contained a non-CA cert.

Adding an `IsCA` check surfaces the problem at admission time instead:

```go
for _, cert := range certs {
    if !cert.IsCA {
        return fmt.Errorf("certificate for %q is not a CA certificate", cert.Subject.CommonName)
    }
}
```

api/v1alpha1/mcpgatewayextension_types.go
```diff
@@ -159,6 +166,20 @@ type OAuthProtectedResource struct {
 	ScopesSupported []string `json:"scopesSupported,omitempty"`
 }
 
+// CACertBundleReference identifies a Secret containing a PEM-encoded CA bundle.
+type CACertBundleReference struct {
+	// name is the name of the Secret resource.
+	// +required
+	// +kubebuilder:validation:MinLength=1
```
Patryk-Stefanski, line 173: **Cleanup: `CACertBundleReference` is structurally identical to `CACertSecretReference`**

Both types have the same two fields with the same names, types, JSON tags, kubebuilder markers, and defaults (`+default="ca.crt"`). The only difference is the type name and doc comment.

If they stay as separate types they'll silently diverge over time — a field added to one (e.g. `Namespace` for cross-namespace references) has to be remembered for the other. `CACertSecretReference` could be reused directly here, or a shared `CACertReference` base type introduced for both.

internal/controller/mcpserverregistration_controller.go
david-martin: Edge case, as it would be unusual for someone to set a gateway-level CA bundle and then mix HTTP and HTTPS servers, but let's allow for it. I'll update.

internal/broker/broker.go
david-martin: Probably better to check the URL scheme. Will add this check for HTTP servers.

internal/controller/ca_cert_bundle.go
david-martin: This is the existing validateCACertPEM used for per-server caCertSecretRef validation too, not introduced by this PR. Adding an IsCA check sounds clean but would break legitimate self-signed certs that don't set BasicConstraints.IsCA = true. Older root CAs and dev certs often omit it entirely. Worth a follow-up to figure out the right behaviour there. Thoughts on whether we should pursue that separately?

api/v1alpha1/mcpgatewayextension_types.go
david-martin: Inclined to keep them separate for readability in CRD docs.

internal/controller/mcpserverregistration_controller.go
david-martin: Actually, reverting this one. The gateway CA bundle is the only signal that internal services speak TLS (they default to http://). Removing the scheme upgrade breaks the core use case: register a TLS server without per-server caCertSecretRef, relying on the gateway bundle alone. The mixed HTTP/HTTPS scenario needs a different mechanism (explicit protocol field or appProtocol on the Service). Happy to track that separately.

david-martin: To clarify: the architectural point is valid. The CA bundle is a trust pool config, not a protocol assertion, and we are conflating the two. But this is existing behaviour: per-server `caCertSecretRef` already does the same double duty (trust pool + scheme upgrade to https). This PR extends the same pattern to the gateway-level bundle.

The reason it works this way: `determineProtocol()` always returns `http` for internal services. There is no standalone "this server speaks TLS" field on MCPServerRegistration. The CA cert references are the only signal available to trigger https.

Separating trust pool config from protocol assertion would need new API surface (e.g. a `protocol` field on MCPServerRegistration). Worth tracking as a follow-up, but out of scope here.

Patryk-Stefanski (dismissed):

Patryk-Stefanski (approved):

internal/controller/ca_cert_bundle.go
david-martin: Tracked in #1244.

PR #1246, opened 2026-07-03
fix: add Serial decorator to CA Cert Bundle e2e test

david-martin:
## Summary
- Add `Serial` decorator to CA Cert Bundle e2e test suite, which mutates the shared `MCPGatewayExtension` and can race with other parallel suites

Flagged by CodeRabbit in #1224.

## Test plan
- [ ] Existing e2e tests pass (no behavioural change, just scheduling constraint)


trepel (approved):
LGTM

PR #1257, opened 2026-07-06
Add accesspolicy controller

vibhor-5:
### Description

This PR introduces the XAccessPolicy Controller, a Kubernetes controller designed to bridge the gap between high-level, gateway-agnostic MCP authorization intent and the concrete enforcement mechanisms provided by Kuadrant's Authorino.

It watches `XAccessPolicy` resources targeting `Gateway` objects and enables declarative, tool-level access control for MCP (Model Context Protocol) servers.

Key Responsibilities:

1. CEL Translation: Translates domain-specific CEL variables (e.g., `request.mcp.tool_name`) into their data-plane equivalents (e.g., `request.headers['x-mcp-toolname']`) that Authorino can evaluate at runtime.
2. Policy Aggregation: Combines multiple `XAccessPolicy` rules targeting the same Gateway into a single Kuadrant `AuthPolicy`, seamlessly handling Kuadrant's 1:1 policy-to-target constraint.

Relates to #804


accesspolicy/api/v1alpha1/xaccesspolicy_types.go
```diff
@@ -0,0 +1,178 @@
...
+	Name  string `json:"name"`
+}
+
+// LocalPolicyTargetReferenceWithSectionName identifies a target resource and optionally a specific section.
+type LocalPolicyTargetReferenceWithSectionName struct {
+	LocalPolicyTargetReference `json:",inline"`
+	SectionName                *string `json:"sectionName,omitempty"`
+}
+
+type AccessPolicySpec struct {
```
guicassolato, line 40: We should probably not redefine the type in our codebase but import it from [`sigs.k8s.io/kube-agentic-networking`](https://pkg.go.dev/sigs.k8s.io/kube-agentic-networking@v0.1.0/api/v1alpha1).

Rebuilding the CRD from code ourselves is unexpected, as we don't own the API group, and risks becoming inconsistent with the source of the upstream. Instead, we should make having the CRD installed a requirement for using the controller.

accesspolicy/internal/controller/xaccesspolicy_controller.go
```diff
@@ -0,0 +1,297 @@
...
+			return ctrl.Result{}, nil
+		}
+		return ctrl.Result{}, err
+	}
+
+	if len(policy.Spec.TargetRefs) == 0 {
+		return ctrl.Result{}, nil
+	}
+
+	targetRef := policy.Spec.TargetRefs[0]
```
guicassolato, line 74: An XAccessPolicy resource could still target multiple different gateways, no? Are planning to support multiple targets?

accesspolicy/internal/controller/xaccesspolicy_controller.go
```diff
@@ -0,0 +1,297 @@
...
+		return ctrl.Result{}, client.IgnoreNotFound(err)
+	}
+
+	r.updateStatus(&policy, agenticv1alpha1.PolicyConditionResolvedRefs, metav1.ConditionTrue, agenticv1alpha1.PolicyReasonResolved, "Gateway found")
+
+	// Fetch all XAccessPolicies targeting this gateway
+	var policyList agenticv1alpha1.XAccessPolicyList
+	if err := r.List(ctx, &policyList, client.InNamespace(policy.Namespace)); err != nil {
+		return ctrl.Result{}, err
+	}
```
guicassolato, line 99: Because what we'll be actually reconciling is the state of the _gateways_, not the policies, there's a different pattern that could be used here. Instead of implementing a reconciler for XAccessPolicy custom resources, we can implement a _Gateway reconciler_ instead – although the name doesn't really matter. The important part is that, with this pattern, the object that the reconciler receives in the request argument is of the `Gateway` kind, not of the `XAccessPolicy` kind. We then define an event mapping function/handler that perceives XAccessPolicy-related events but enqueues calls to reconcile each of the Gateway objects referenced in the `spec.targetRefs` of the policy instead.

This could save from having to list all gateways for each individual policy reconciliation and, within the loop, all policies affected by the gateway, while multiple reconciliation events related to multiple XAccessPolicy resources may be enqueue – effectively to yield the same outcome but with more cycles.

See controller-runtime's [`EnqueueRequestsFromMapFunc`](https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/handler#example-EnqueueRequestsFromMapFunc).

accesspolicy/internal/controller/xaccesspolicy_controller.go
```diff
@@ -0,0 +1,297 @@
...
+		if len(p.Spec.TargetRefs) == 0 || string(p.Spec.TargetRefs[0].Name) != gateway.Name {
+			continue
+		}
+		if p.Spec.Action == agenticv1alpha1.ActionTypeExternalAuth {
+			continue
+		}
+
+		for _, rule := range p.Spec.Rules {
+			if rule.Source.Type == agenticv1alpha1.AuthorizationSourceTypeSPIFFE {
+				continue
```
guicassolato, line 116: I'll have to think this through, but I think there may be something we can do regarding authentication here.

I'm thinking maybe using the [`plain`](https://docs.kuadrant.io/latest/authorino/docs/features/#plain-authenticationplain) type of authentication from the AuthPolicy when an SPIFFE ID is known to be injected by the proxy, say, for example, because it has mTLS authentication enabled.

PR #1261, opened 2026-07-07
docs: MCP Resources Federation design doc

kkukadia:
## Summary

Design doc for federating MCP Resources through the gateway, following the pattern established for tools and prompts. Split from #208, tracked in #788.

This is a **design proposal, not implementation**. Opening it for review before anything gets merged or built against it.

## Scope

- Federates `resources/list` and `resources/read` across upstream servers
- Rewrites `ui://` URIs with per-server prefixes to avoid collisions
- Rewrites `_meta.ui.resourceUri` in `tools/call` responses to stay consistent

Non-goals and full rationale are in the doc itself.

## Addressed from the first review round

- Response rewrite: new sibling `resourceURIRewriter`, not an extension of `sseRewriter` (renamed to `elicitationRewriter` now that a second rewriter exists)
- Prefix safety: explicit re-check at URI-injection time, independent of the CRD's pattern validation
- Partial resource list on upstream failure: acceptable, consistent with how tools/prompts already behave, see "Open Questions" in the doc for the full reasoning
- Added a "Forward Compatibility" section explaining why this design mostly sidesteps the upcoming SEP-2575/SEP-2567 spec changes (no session-scoped cache, no subscriptions, shared upstream connection abstraction, prefix-based routing, stream-scoped response rewriting)

## New, please take a look

- `AddAfterListResources` needs to fan out to upstreams concurrently, not sequentially. This fetch now runs in the client-facing request path (unlike tools/prompts, which fetch in the background), so a sequential loop would add every slow upstream's timeout to every client's latency
- `resources/list` JWT filtering only controls listing, not `resources/read` access, same known limitation `MCPVirtualServer` already has for tools. `HandleResourceRead` needs to set `x-mcp-servername` for AuthPolicy to actually enforce it

## Future considerations (deliberately out of scope, not blocking)

- Capability negotiation without the `initialize` handshake once SEP-2575 lands. Per maintainer guidance on #788, this design doesn't build against an unreleased spec change; flagged for whoever picks up that migration later

Ref: #788, split from #208

## Summary by CodeRabbit

* **Documentation**
  * Added a new design document, “MCP Resources Federation,” covering gateway support for federating MCP `resources/list` and `resources/read` across multiple upstream servers.
  * Documents `ui://` URI namespacing and deterministic longest-prefix routing for resource reads.
  * Explains request-time aggregation for listings with partial results on upstream timeouts/failures, including how pagination is handled.
  * Details authorization-driven filtering for listings, response URI rewriting for tool calls, and a testing strategy.


kkukadia:
@Patryk-Stefanski @david-martin could you please review

david-martin:
Will do. Thanks @k1chik 
I'll assign #788 to you so we can reopen the pr. (to do that you'll need to add a comment first to 788, or i think accepting the invite to the org should suffice)

david-martin:
Notes from initial review:

* Prefix validation must exclude URI delimiter characters (/, ?, #), since the prefix is injected into a URI authority segment (unlike tools/prompts where it's plain string concat).
* `_meta.ui.resourceUri` response rewrite would be implemented in internal/mcp-router/response_handlers.go, called from https://github.com/Kuadrant/mcp-gateway/blob/2f0b77c4507dfcd5e8a3eb8ecd3ff39bf41ff4cd/internal/mcp-router/server.go#L314-L326
* As the new mcp spec hasn't shipped yet, we should implement against the current stateful spec, but also present possible approaches for stateless as future consideration (but not commit to or design around the unreleased change in this for now). That can follow.
* Adding `ui://` resources to an existing test server is reasonable, no need for a dedicated one.

kkukadia:
> Will do. Thanks @k1chik I'll assign #788 to you so we can reopen the pr. (to do that you'll need to add a comment first to 788, or i think accepting the invite to the org should suffice)

@david-martin just self assigned, thank you!

kkukadia:
@david-martin thanks for the feedback!

**URI validation** : the CRD's `prefix` pattern already excludes `/`, `?`, `#`, (happy coincidence). IIUC It's written for tool-name collisions, not URI safety. May I suggest that we add an explicit re-check at injection time, independent of the CRD validation, so a future loosening of that pattern can't silently break URI safety for resources.

**Response rewrite** : I could be wrong here, but reusing `sseRewriter` strikes me as not quite right (please correct me if I'm wrong), it only kicks in for elicitation sessions, and it assumes SSE streaming, so a plain JSON tool response wouldn't get rewritten. I think a sibling `resourceURIRewriter` gated on `isToolCall()`, using the same closure-scoped `mcpRequest` you pointed at, is a better fit. But open to feedback if I'm missing something.

**Stateless spec** : Agreed, I moved it to a `Future Considerations` note instead. The one thing I want to flag: `internal/broker/upstream/manager.go` is the shared piece that'd need to change if the handshake model ever does, so I'm calling that out now rather than leaving it a surprise later.

**Test server** : adding ui:// resource to the existing server1

david-martin:
capturing comments from sync call

> **URI validation** : the CRD's `prefix` pattern already excludes `/`, `?`, `#`, (happy coincidence). IIUC It's written for tool-name collisions, not URI safety. May I suggest that we add an explicit re-check at injection time, independent of the CRD validation, so a future loosening of that pattern can't silently break URI safety for resources.

This suggestion makes sense. 
A separate validation check for resource uris vs. tool names to avoid the silent break in future.


> **Response rewrite** : I could be wrong here, but reusing `sseRewriter` strikes me as not quite right (please correct me if I'm wrong), it only kicks in for elicitation sessions, and it assumes SSE streaming, so a plain JSON tool response wouldn't get rewritten. I think a sibling `resourceURIRewriter` gated on `isToolCall()`, using the same closure-scoped `mcpRequest` you pointed at, is a better fit. But open to feedback if I'm missing something.

Makes sense as well. A separate writer specifically for resource uri rewriting the resource read response body.
The existing `rewriter` struct instance could be renamed to reflect it's for elicitations?

docs/design/resources-federation/resources-federation-design.md
```diff
@@ -0,0 +1,197 @@
...
+```text
+ui://template.html  →  ui://{prefix}template.html
+```
+
+For example, with prefix `insights_`:
+```text
+ui://template.html  →  ui://insights_template.html
+```
+
+The router extracts the prefix back out by matching the authority against registered server prefixes.
```
david-martin, line 44: How should prefixes that are a substring of another prefix be handled?
e.g. 
  prefix "app_"       →  ui://app_template.html
  prefix "app_admin_" →  ui://app_admin_dashboard.html

is ui://app_admin_template.html a resource called admin_template.html on server app_, or template.html on server app_admin_?
Longest-prefix match picks app_admin_? and document?

docs/design/resources-federation/resources-federation-design.md
```diff
@@ -0,0 +1,197 @@
...
+```
+
+### Component Changes
+
+| Component | File | Change |
+|---|---|---|
+| Upstream client | `internal/broker/upstream/mcp.go` | Add `SupportsResources()` and `ListResources()` to the `MCP` interface |
+| Upstream connection | `internal/broker/upstream/manager.go` | Add `ListResources()` for pull-time fetching; no pre-registration |
+| Broker | `internal/broker/broker.go` | Enable resource capabilities, gated on at least one upstream supporting resources, following the same pattern as prompts; register `AddAfterListResources` hook; add `GetServerInfoByResource()` to `MCPBroker` interface |
+| Router request | `internal/mcp-router/request_handlers.go` | Add `HandleResourceRead()`, reusing the existing `validateSession()` check `HandleToolCall` already uses; add `resources/read` case in `RouteMCPRequest`; add `ResourceURI()` to `MCPRequest` |
```
david-martin: also, sets x-mcp-servername routing header?

docs/design/resources-federation/resources-federation-design.md
```diff
@@ -0,0 +1,197 @@
...
+| Router response | `internal/mcp-router/server.go`, new `internal/mcp-router/resource_rewrite.go` | Construct `resourceURIRewriter` in the `ResponseHeaders` case (mirroring how `sseRewriter` is wired today); detect and rewrite `_meta.ui.resourceUri` in `tools/call` response bodies |
+| Router response (rename) | `internal/mcp-router/elicitation.go` | Rename `sseRewriter` to `elicitationRewriter` so its name reflects what it's for, now that a second response rewriter exists |
+| Config / CRD | `internal/config/types.go`, `api/v1alpha1/types.go` | No changes (VirtualServer filtering out of scope) |
+
+`GetServerInfoByResource(uri string)` parses the authority segment of the URI and does longest-prefix matching against registered server prefixes - the same approach `GetServerInfo` already uses for tools.
+
+The `AddAfterListResources` hook calls `ListResources()` on each active upstream with a per-upstream timeout (same default as the broker's existing upstream timeout), rewrites the `ui://` URIs, and merges results. Upstreams that error or time out are skipped with a log - the request is not failed. Upstreams with no prefix are skipped entirely.
```
david-martin: Note the change of mcp library on main to the official mcp go sdk.
A middleware pattern is now used.

docs/design/resources-federation/resources-federation-design.md
```diff
@@ -0,0 +1,197 @@
...
+
+The existing `sseRewriter` (`internal/mcp-router/elicitation.go`), used today for rewriting elicitation request IDs, is the closest precedent for the Process/Flush lifecycle, but isn't a good fit to extend directly:
+
+- It's only constructed when `clientElicitation && statusCode == "200"` - resource-URI rewriting needs to run on any tool call response, not just sessions that registered for elicitation.
+- It assumes SSE framing (splits on `\n`, only rewrites `data:`-prefixed lines). A typical non-streaming `tools/call` JSON response has neither, so reusing it as-is would silently fail to rewrite plain JSON bodies.
+
+Proposed instead: a new sibling rewriter (`resourceURIRewriter`), constructed in the `ResponseHeaders` case gated on `mcpRequest.isToolCall()` alone, following the same Process/Flush pattern but handling both plain single-JSON bodies and SSE-framed bodies (tool responses can be either, depending on the upstream's transport). It runs independently of the elicitation rewriter - composed, not merged into one struct, for responses that need both.
```
david-martin: > gated on `mcpRequest.isToolCall()` alone

can we make this gate narrower so it doesn't happen with all tool call responses, regardless of if resource uris meed to be rewritten or not?

david-martin: also, is response headers phase best place to construct the writer, or request body? 
(i don't know, asking the question)

docs/design/resources-federation/resources-federation-design.md
```diff
@@ -0,0 +1,197 @@
...
+
+### Forward Compatibility
+
+Two spec changes are coming that will reshape how clients and servers connect: SEP-2575 drops the `initialize`/`initialized` handshake, and SEP-2567 drops protocol sessions and restricts server-to-client requests. This design mostly stays out of the way of both, so there's little to unwind later:
+
+- **No session-scoped cache.** `AddAfterListResources` fetches live on every `resources/list` call instead of caching per-session state, unlike tools/prompts.
```
david-martin: OK.
This is the trade off as resources are more likely to be user specific than tools.
(tools can be too, but we have user specific tools lists capability for that)

docs/design/resources-federation/resources-federation-design.md
```diff
@@ -0,0 +1,197 @@
...
+- URI templates (`resources/templates/list`)
+- Stateless (Streamable HTTP) protocol support
+- VirtualServer filtering for resources
+- `cacheScope` / `ttlMs` cache-aware proxying (SEP-2549, future consideration - see [scoping discussion on #788](https://github.com/Kuadrant/mcp-gateway/issues/788#issuecomment-4682923399)). If built, invalidation should be `ttlMs`-based rather than relying on `notifications/resources/list_changed`, since SEP-2567 restricts the server-to-client push that notification depends on.
+- Pagination - `resources/list` supports cursor-based pagination in the spec; aggregating cursors across multiple upstreams is a non-trivial problem deferred to a follow-up. In the meantime, each upstream is only fetched for its first page: a `nextCursor` in an upstream's response is not followed, so that upstream's remaining resources are left out of the federated list. `AddAfterListResources` logs when an upstream's response includes a `nextCursor`, so a paginated upstream showing up as partial is observable rather than silent, the same treatment as an upstream that errors or times out.
```
david-martin: OK on this, to limit scope of support in stateful spec version

docs/design/resources-federation/resources-federation-design.md
```diff
@@ -0,0 +1,197 @@
...
+  "resources": { "insights-server": ["ui://insights_template.html"] }
+}
+```
+
+A new `filtered_resources_handler.go` mirrors `filtered_prompts_handler.go`. Tools and prompts filter a pre-populated set; resources filter **per-upstream, inside the `AddAfterListResources` hook, before the results get merged** - each upstream's list is filtered on its own, matching the per-server structure of the `resources` claim in the JWT.
+
+Enforcement doesn't change here: a missing `resources` key makes no assertion about resources (`enforceCapabilityFilter` still governs behavior), and an empty map (`"resources": {}`) explicitly denies all resources.
+
+### Security Considerations
+
```
david-martin, line 166: any updates to https://github.com/Kuadrant/mcp-gateway/blob/main/docs/design/security-architecture.md needed ?

david-martin (changes requested):

docs/design/resources-federation/resources-federation-design.md
kkukadia: I like longest-prefix match - it also matches existing precedent in GetServerInfo's userSpecificList fallback (broker.go:575-593), which already resolves ambiguous prefix matching 

One thing to note though, it doesn't eliminate the collision. 

app_'s admin_template.html becomes unreachable via resources/read once app_admin_'s template.html shadows it - both compute to the identical URI.

I'll add this note into the doc as well ( longest-prefix-match rule) and calling out the shadowing trade-off

docs/design/resources-federation/resources-federation-design.md
kkukadia: yes, it was in "security considerations" but I moved it to the table

docs/design/resources-federation/resources-federation-design.md
kkukadia: Thanks, the doc still uses the old AddAfterListResources hook API instead of the new filteringMiddleware pattern, so I'll reword that everywhere. I'll also explicitly advertise the resources capability, since we're not pre-registering anything

docs/design/resources-federation/resources-federation-design.md
kkukadia: > > gated on `mcpRequest.isToolCall()` alone
> 
> can we make this gate narrower so it doesn't happen with all tool call responses, regardless of if resource uris meed to be rewritten or not?

We can gate on whether the server has a prefix instead, those without one are already excluded from resource federation, so nothing's lost. Won't catch it per-tool, but it's free and I'll add it to the doc.

> also, is response headers phase best place to construct the writer, or request body? (i don't know, asking the question)

imo ResponseHeaders bc:
1. Envoy only honors the streaming override when it's set on a headers-phase response - it has to happen here regardless of when we decide.
2. We get the status code by then too, so we can skip streaming on error responses (RequestBody can't do that).
3. Same pattern the elicitation rewriter already uses today, proven out by the e2e tests

docs/design/resources-federation/resources-federation-design.md
kkukadia: No major updates needed. One thing I'll fix: the doc currently compares this to the tools/MCPVirtualServer limitation, but it's actually worse for resources since there's no per-URI header, only server-level. I'll correct that in the Security Considerations section instead of the risk table!

david-martin (approved):
Thanks for the work on this @kkukadia. Much appreciated

PR #1269, opened 2026-07-09
Isolated router

maleck13:
## What does this PR do?

Extracts routing logic from the ext_proc adapter into a transport-agnostic internal/routing/ package with zero Envoy imports. This is Phase 1 groundwork for 2026-07-28 protocol support and future Praxis port.

## Summary by CodeRabbit

* **New Features**
  * Introduced a routing layer for tools, prompts, and prefixes backed by live routing-table snapshots.
  * Added MCP protocol transition support, including header-based routing for the newer variant.
  * Improved request/response handling for elicitation-aware tool calls, including correct response streaming behaviour.
* **Bug Fixes**
  * Hardened gateway identity headers by setting the correct authority and stripping any client-forged verified identity headers.
  * Ensured consistent `413` responses when buffered request bodies exceed the configured limit.
* **Documentation**
  * Added design and task-plan documentation for rolling out multi-version MCP routing semantics.


internal/routing/router_202511.go
```diff
@@ -0,0 +1,623 @@
...
+	case mcpReq.IsElicitationResponse():
+		span.SetAttributes(attribute.String("mcp.route", "elicitation-response"))
+		return r.routeElicitationResponse(ctx, mcpReq)
+	case mcpReq.Method == MethodToolCall:
+		span.SetAttributes(attribute.String("mcp.route", "tool-call"))
+		return r.routeToolCall(ctx, table, mcpReq)
+	case mcpReq.Method == MethodPromptGet:
+		span.SetAttributes(attribute.String("mcp.route", "prompt-get"))
+		return r.routePromptGet(ctx, table, mcpReq)
+	default:
```
david-martin, line 75: should probably check `BrokerPass` here to be more explicit

docs/design/router-2026-07-28/router-2026-07-28-design.md
```diff
@@ -0,0 +1,303 @@
...
+- Broker changes for `2026-07-28` (`server/discover`, `InputRequiredResult`, `ttlMs` — separate design)
+- `2025-11-25` deprecation or removal
+- Independent deployment of router and broker
+- Changes to controller or operator
+
+## Job Stories
+
+### When a `2026-07-28` client calls a tool
+
+When an MCP client sends `tools/call` with `MCP-Protocol-Version: 2026-07-28`, `Mcp-Method: tools/call`, and `Mcp-Name: github_search`, the router wants to route the request to the github backend using only headers so that no body parsing is needed for the routing decision.
```
david-martin: does this contradict this:

> Validate `Mcp-Name` and `Mcp-Method` headers match the body (spec requirement: server must reject mismatches with `HeaderMismatch` error)

Though maybe if there's no prefix, we defer that check to the upstream mcp server

david-martin (dismissed):
Approving with non blocking comments (can be addressed in follow up if agreed)

internal/routing/router_202511.go
maleck13: It actually not needed here as there is no re-writes etc needed

docs/design/router-2026-07-28/router-2026-07-28-design.md
maleck13: I actually removed these job stories as they are really not use cases as such

david-martin (approved):

Issue #161, opened 2025-09-10
Observability Phase 3 - Metrics: Journey 2 complete


david-martin:
Building on #428, implement metrics as per the observability plan in https://github.com/kagenti/mcp-gateway/pull/350

* Define and implement metrics in proposal
* Optional: leverage Istio Telemetry tag overrides for per-tool metrics.
* Provide default (Grafana?) dashboards

*Outcome: Gateway admins get visibility into traffic patterns and hotspots*

For reference, example of pushing metrics to an otel collector and bridging from prometheus metrics in kuadrant operator https://github.com/Kuadrant/kuadrant-operator/pull/1712/files

david-martin:
**Additional use case: tool output size observability for context bloat awareness**

Gateway operators need visibility into the size of tool definitions and tool call responses flowing through the gateway. As tool counts grow across federated MCP servers, aggregate metadata and tool outputs become significant contributors to context bloat in downstream LLM clients.

Interest from @evaline-ju in measuring and correlating tool output sizes to token usage, with the goal of identifying redundant or expired tool calls that inflate context windows. Related research on trajectory reduction for LLM agents: https://arxiv.org/pdf/2509.23586

Metrics to consider as part of this phase:
- Size of tools/list response payloads (per-server and aggregate)
- Size of individual tool call response payloads
- Tool count per server and per gateway
- Response size percentiles to identify outlier tools

This would give operators early warning when tool metadata or outputs are inflating context windows, and help inform decisions about which servers to expose, filter, or prioritize for output compression.

PR #1281, opened 2026-07-14
docs: metrics phase 3 design

### LINKED ISSUE #1282: Metrics design doc

david-martin:
closes #1282 

## Summary

- Design doc for observability phase 3 metrics, covering broker-side OTel metrics for discovery health, tool inventory, upstream stability, and context contribution
- Documents existing gateway metrics and provides a reference Istio Telemetry config for adding MCP dimensions (tool name, server name) via header-based tag overrides
- Scoped to the gateway operator persona: configuration and infrastructure visibility, not MCP-level application errors

Relates to #161

## Review

Design doc only, no code changes. Looking for feedback on the metric definitions, job stories, and scope boundaries before writing the implementation plan.

## Summary by CodeRabbit

* **Documentation**
  * Added a Phase 3 “Metrics: Observability” design for the MCP Gateway, covering gateway, broker, and router metrics.
  * Documented existing gateway Istio metrics and provided example PromQL queries for server/tool usage and latency.
  * Defined broker OpenTelemetry metrics for discovery health, reconnection failures, tool inventory counts, and tools/list response size.
  * Outlined the planned metrics-labeling approach (adding MCP dimensions) and included scope, limitations, security, and future dashboard/alert considerations.

### DIFF
--- a/docs/design/metrics-phase3/metrics-phase3-design.md
+++ b/docs/design/metrics-phase3/metrics-phase3-design.md
@@ -215,8 +215,9 @@ mcp_broker_tools_list_response_bytes
   type: gauge
   labels: server_name
   description: last observed size in bytes of tools/list response per upstream server.
-               updates on each discovery. sum across servers gives total
-               gateway context footprint.
+               updates on each discovery. set to 0 when a server becomes
+               unreachable. removed when the server registration is deleted.
+               sum across servers gives total gateway context footprint.
 ```
 
 ```promql
--- a/docs/design/metrics-phase3/metrics-phase3-design.md
+++ b/docs/design/metrics-phase3/metrics-phase3-design.md
@@ -35,8 +35,6 @@ Add OpenTelemetry metrics to the broker for discovery health, tool counts, and r
 
 The gateway operator: the person deploying the MCP Gateway, creating `MCPServerRegistration` resources, and wiring upstream MCP servers. They need to know if their configuration is working and how traffic is flowing.
 
-MCP-level application errors (a tool returning `isError: true` inside an HTTP 200 JSON-RPC response) are out of scope. If the gateway routed the request and got a response, the gateway did its job. Tool execution errors are the concern of MCP server developers, not gateway operators. See [MCP spec: Tools - Error Handling](https://modelcontextprotocol.io/specification/2025-06-18/server/tools).
-
 ## Metrics Sources
 
 Three components can surface metrics, each covering a different layer:
@@ -65,7 +63,7 @@ When a gateway operator receives reports of slow tool calls, they want to see pe
 
 When a gateway operator wants to understand how much tool metadata the gateway is contributing to conversations across all federated servers, they want to see tool list response sizes and tool counts per server, so that they can identify servers contributing disproportionate context and make informed decisions about which servers to expose, filter, or prioritize for output compression.
 
-This concern is not unique to gateway operators. Context bloat from tool metadata is a common problem for any MCP client pulling tools from multiple servers. The gateway is the natural aggregation point for this data, providing centralized visibility that individual MCP clients and server developers cannot easily get on their own.
+Context bloat from tool metadata is a common problem for any MCP client pulling tools from multiple servers. The gateway is the natural aggregation point, providing centralized visibility across all servers.
 
 ### When the operator needs to confirm gateway infrastructure is healthy
 
@@ -136,12 +134,7 @@ These metrics already exist via Istio and require no new code:
 - `istio_requests_total` — request count by response code, reporter, source, destination
 - `istio_request_duration_milliseconds` — request latency histogram
 
-For the operator, these answer:
-- Are client requests reaching upstreams and getting responses? (filter by status code)
-- How fast are tool calls end-to-end? (latency by destination)
-- Is one route/server seeing more errors than others? (filter by destination and status)
-
-HTTP 4xx/5xx from the gateway indicates configuration or upstream problems the operator needs to see. HTTP 200 means the gateway did its job, regardless of MCP-level tool errors in the response body.
+These cover request rate, latency, and HTTP-level error rates per route. No new instrumentation needed, but they should be documented for operators.
 
 ### Istio Telemetry Tag Overrides (optional)
 
@@ -181,9 +174,7 @@ spec:
           value: "request.headers['x-mcp-toolname']"
 ```
 
-This adds per-tool and per-server dimensions to gateway request rate and latency metrics. Cardinality scales with the number of unique tools, which is operator-controlled.
-
-A reference configuration will be provided in `examples/otel/istio-mcp-metrics.yaml`.
+A reference configuration will be provided in `examples/otel/istio-mcp-metrics.yaml`. Cardinality scales with the number of unique tools, which is operator-controlled.
 
 ### Metrics Endpoint
 
@@ -208,7 +199,7 @@ A tool call returning HTTP 200 with `isError: true` in the JSON-RPC result is in
 
 ### Router metrics deferred
 
-The router (ext_proc) processes every client request and has access to MCP method, tool name, and server name. Adding metrics here would provide native MCP-aware request rate and latency without depending on Istio tag overrides. This is deferred to a future phase to keep scope small.
+Adding metrics to the router would provide native MCP-aware request rate and latency without depending on Istio tag overrides. Deferred to keep scope small (see Metrics Sources table).
 
 ### No dashboards
 
--- a/docs/design/metrics-phase3/metrics-phase3-design.md
+++ b/docs/design/metrics-phase3/metrics-phase3-design.md
@@ -71,62 +71,6 @@ When a gateway operator wants to confirm that the gateway, router, and broker ar
 
 ## Design
 
-### Broker Metrics
-
-The broker is the only component that sees discovery outcomes and upstream reachability. All metrics use the OpenTelemetry metrics API, consistent with existing OTel tracing in the broker.
-
-#### Discovery health
-
-```
-mcp_broker_discovery_total
-  type: counter
-  labels: server_name, status (success | failure)
-  description: number of discovery attempts per upstream server
-```
-
-```
-mcp_broker_discovery_duration_seconds
-  type: histogram
-  labels: server_name
-  description: time taken for tools/list calls during discovery
-```
-
-#### Tool inventory
-
-```
-mcp_broker_tools_discovered
-  type: gauge
-  labels: server_name
-  description: current number of tools discovered per upstream server.
-               replaces the removed discoveredTools status field.
-               set to 0 when a server becomes unreachable.
-```
-
-#### Upstream stability
-
-```
-mcp_broker_upstream_reconnections_total
-  type: counter
-  labels: server_name
-  description: number of upstream reconnection attempts.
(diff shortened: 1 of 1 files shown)


docs/design/metrics-phase3/metrics-phase3-design.md
```diff
@@ -0,0 +1,270 @@
...
+
+### When the operator wants to understand context contribution
+
+When a gateway operator wants to understand how much tool metadata the gateway is contributing to conversations across all federated servers, they want to see tool list response sizes and tool counts per server, so that they can identify servers contributing disproportionate context and make informed decisions about which servers to expose, filter, or prioritize for output compression.
+
+Context bloat from tool metadata is a common problem for any MCP client pulling tools from multiple servers. The gateway is the natural aggregation point, providing centralized visibility across all servers.
+
+### When the operator needs to confirm gateway infrastructure is healthy
+
+When a gateway operator wants to confirm that the gateway, router, and broker are running and processing requests, they want to use existing Kubernetes probes and gateway metrics (`istio_requests_total`), so that they can monitor infrastructure health without additional instrumentation.
```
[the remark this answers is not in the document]
david-martin: No. focusing on gateway operator only

docs/design/metrics-phase3/metrics-phase3-design.md
```diff
@@ -0,0 +1,270 @@
...
+
+## Design
```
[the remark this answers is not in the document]
david-martin: No API, storage, or flow changes in this design. Metrics are emitted by the broker and scraped externally. Doesn't warrant empty subsections.

docs/design/metrics-phase3/metrics-phase3-design.md
```diff
@@ -0,0 +1,270 @@
...
+```
+
+```promql
+# total tools/list payload across all servers
+sum(mcp_broker_tools_list_response_bytes)
+```
+
+#### Cardinality
+
+`server_name` is bounded by the number of `MCPServerRegistration` resources, which is operator-controlled and typically in the tens, not thousands. No high-cardinality labels (session IDs, tool call IDs) are used. These belong in traces and logs.
```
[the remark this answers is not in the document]
david-martin: tool label not included on any broker metrics. 
It's optional for gateway metrics via Telemetry resource, as detailed earlier, with cardinality called out.

docs/design/metrics-phase3/metrics-phase3-design.md
```diff
@@ -0,0 +1,270 @@
...
+  labels: server_name
+  description: last observed size in bytes of tools/list response per upstream server.
+               updates on each discovery. sum across servers gives total
+               gateway context footprint.
+```
+
+```promql
+# total tools/list payload across all servers
+sum(mcp_broker_tools_list_response_bytes)
+```
```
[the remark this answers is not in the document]
david-martin: updated. reset to 0 on failure. remove if MCPSR is removed.

docs/design/metrics-phase3/metrics-phase3-design.md
```diff
@@ -0,0 +1,271 @@
...
+
+- Router metrics for native MCP-aware request rate and latency
+- Grafana dashboard definitions
+- Tool response size metrics (requires response body access in the router)
+- Per-tool schema size metrics for identifying individual outlier tool definitions within a server
+- Alerting rule templates (e.g., discovery failure rate > threshold)
+
+## Execution
+
+See [tasks/tasks.md](tasks/tasks.md) for the implementation plan.
```
Patryk-Stefanski: The `tasks/tasks.md` file referenced here does not exist in this PR or on the branch. Either include it or remove the reference (\"implementation plan to follow\" would be fine).

docs/design/metrics-phase3/metrics-phase3-design.md
```diff
@@ -0,0 +1,271 @@
...
+  type: gauge
+  labels: server_name
+  description: current number of tools discovered per upstream server.
+               replaces the removed discoveredTools status field.
+               set to 0 when a server becomes unreachable.
+```
+
+```promql
+# current tool count per server
+sum(mcp_broker_tools_discovered) by (server_name)
```
Patryk-Stefanski, line 193: The definition of \"reconnection attempt\" needs to be more precise. In the current code (`internal/broker/upstream/manager.go`), there is no explicit reconnect event — the ticker fires, `manage()` calls `mcp.Connect()`, and `OnConnectionLost` just logs. Two implementors would instrument different things without clarity here. Is it: every `Connect()` call after a failed ping? Every `OnConnectionLost` callback? Every ticker tick while in a failed state? The instrumentation point needs to be specified.

Patryk-Stefanski (commented):
Review of the metrics Phase 3 design doc. A few items worth addressing before implementation starts, plus some discussion points.

docs/design/metrics-phase3/metrics-phase3-design.md
Patryk-Stefanski: The `tasks/tasks.md` file doesn't exist in this PR or on the branch. Either include it or change this to something like "implementation plan to follow."

docs/design/metrics-phase3/metrics-phase3-design.md
```diff
@@ -0,0 +1,271 @@
...
+mcp_broker_upstream_reconnections_total
+  type: counter
+  labels: server_name
+  description: number of upstream reconnection attempts.
+               frequent reconnections signal instability.
+```
+
+```promql
+# servers currently experiencing reconnections
+sum(rate(mcp_broker_upstream_reconnections_total[5m])) by (server_name) > 0
```
Patryk-Stefanski: "Reconnection attempt" needs a more precise definition. In the current code (`internal/broker/upstream/manager.go`), there is no explicit reconnect event — the ticker fires on an interval, `manage()` calls `mcp.Connect()`, and `OnConnectionLost` just logs. Without specifying the exact instrumentation point, two implementors would count different things. Options to clarify: every `Connect()` call that follows a failed ping? Every `OnConnectionLost` callback? Every ticker tick while in a disconnected state?

docs/design/metrics-phase3/metrics-phase3-design.md
```diff
@@ -0,0 +1,271 @@
...
+
+```
+mcp_broker_tools_list_response_bytes
+  type: gauge
+  labels: server_name
+  description: last observed size in bytes of tools/list response per upstream server.
+               updates on each discovery. set to 0 when a server becomes
+               unreachable. removed when the server registration is deleted.
+               sum across servers gives total gateway context footprint.
+```
```
Patryk-Stefanski, line 224: Worth discussing: `histogram` may be a better fit here than `gauge`. A histogram gives you percentiles across discovery cycles, which is more useful for detecting size drift or sudden bloat. A gauge works if the intent is strictly a "current footprint snapshot", but then the lifecycle semantics should be stated more explicitly — the doc says the value goes to 0 on disconnect but the metric is removed on deregistration. Those behave differently in alerting (a disappearing series vs. a zero).

docs/design/metrics-phase3/metrics-phase3-design.md
```diff
@@ -0,0 +1,271 @@
...
+mcp_broker_discovery_total
+  type: counter
+  labels: server_name, status (success | failure)
+  description: number of discovery attempts per upstream server
+```
+
+```promql
+# discovery failure rate per server
+sum(rate(mcp_broker_discovery_total{status="failure"}[5m])) by (server_name)
```
Patryk-Stefanski, line 162: The manager has an explicit early-exit path for disabled servers (`!man.mcp.IsEnabled()`) that is distinct from a connection failure. Should `disabled` be a third `status` value here? Counting disabled-server skips as `failure` would skew the failure rate PromQL queries shown below.

docs/design/metrics-phase3/metrics-phase3-design.md
```diff
@@ -0,0 +1,271 @@
...
+The router already sets `x-mcp-servername`, `x-mcp-toolname`, and `x-mcp-method` as request headers. Istio's Telemetry API can promote these to metric labels without code changes:
+
+```yaml
+apiVersion: telemetry.istio.io/v1
+kind: Telemetry
+metadata:
+  name: mcp-metrics
+  namespace: gateway-system
+spec:
+  metrics:
```
Patryk-Stefanski, line 105: Minor: the existing `examples/otel/istio-telemetry.yaml` uses `telemetry.istio.io/v1alpha1`. This doc uses `v1`. `v1` graduated in Istio 1.20 and is correct going forward, but the inconsistency will cause confusion. Either update the existing example or add a note explaining the difference.

docs/design/metrics-phase3/metrics-phase3-design.md
david-martin: Convention is to add the tasks dir after design approval. Will update the reference to say "implementation plan to follow".

docs/design/metrics-phase3/metrics-phase3-design.md
david-martin: ~Agreed, this needs to be precise. Looking at the code, the clearest instrumentation point is `handleConnectionFailure` in `manager.go`. It fires on every `Connect()` or `Ping()` failure inside `manage()`, which covers both initial connection failures and health check failures on an existing connection. `OnConnectionLost` just logs and waits for the next tick, so it is not a distinct reconnection event.~

~Will update the doc to define it as: incremented each time `manage()` enters `handleConnectionFailure`, covering both failed `Connect()` and failed `Ping()` on an already-connected client.~

i didn't say this. see next comment

docs/design/metrics-phase3/metrics-phase3-design.md
david-martin: Had it as histogram first, but the bucket boundaries are hard to define sensibly. Response size per server is slow-changing and deliberate (driven by how many tools are registered), not bound to factors like load or network latency. Percentiles across discovery cycles are not useful here since we are not looking for outliers over time.

A gauge stores the last value and keeps querying simple. The lifecycle semantics (0 on failure, removed on deregistration) are already documented. If size drift matters, `changes()` or recording rules over the gauge work without histogram complexity.

docs/design/metrics-phase3/metrics-phase3-design.md
david-martin: Inclined to not emit the metric at all for a disabled server. The manager already removes tools, disconnects, and sets status to not ready when `IsEnabled()` returns false. Including it as a third status value would just add noise to the discovery counter for something that is an explicit operator choice, not a failure.

docs/design/metrics-phase3/metrics-phase3-design.md
david-martin: `v1` is correct going forward (graduated in Istio 1.20). The existing example should be updated separately.

docs/design/metrics-phase3/metrics-phase3-design.md
david-martin: Changing to a failures count instead. Renamed to `mcp_broker_upstream_connection_failures_total`, instrumentation point is `handleConnectionFailure` in `manager.go`. Description now calls out the backoff decay.

Patryk-Stefanski (approved):

PR #1294, opened 2026-07-21
feat(e2e): support running e2e tests on OpenShift clusters

trepel:
## What does this PR do?

Makes the e2e test suite portable across Kind and OpenShift by parameterizing environment-specific values (hostnames, TLS certificates, CA bundles) and adding a gateway template for the elicitation test suite.

## Changes

  - Add `buildHostnames()` to consolidate HTTPRoute hostname construction based on cluster type
  - Add `gateway-elicitation.yaml.template` with parameterized listener hostnames and TLS secret
  - Parameterize `GATEWAY_TLS_SECRET`, `GATEWAY_TLS_LISTENER_HOSTNAME`, and `GATEWAY_CA_BUNDLE_CONFIGMAP`
  - Consolidate CA patching into `PatchBrokerCA` with Kind/OCP branching
  - Add `elicitationPublicHostDefault()` to match KIND gateway URL hostname
  - Remove hardcoded `PrivateHost` and `WithListenerPort` (now handled by the operator)
  - Replace hardcoded hostnames in tool discovery tests with configurable `ToolDiscoveryServerHost`
  - Mark two multi-gateway tests `Serial` for parallel safety
  - Fix cleanup ordering for `conflictExt` in shared-gateway test
  - Add `WaitForToolsWithPrefix` before JWT-scoped tool discovery test
  - Disable `FASTMCP_HTTP_HOST_ORIGIN_PROTECTION` on server3 for cross-origin OCP routing
  - Bump `WAIT_TIME` to 150s for slower OCP gateway provisioning

## Verification

  - Run e2e tests on Kind: `make test-e2e-ci`
  - Run e2e tests on OpenShift with appropriate env vars:
    ```
    E2E_DOMAIN=<domain> E2E_SCHEME=https GATEWAY_CLASS_NAME=<class> \
    GATEWAY_TLS_SECRET=<secret> GATEWAY_TLS_LISTENER_HOSTNAME=*.<domain> \
    GATEWAY_CA_BUNDLE_CONFIGMAP=<cm> make test-e2e-ci
    ```
  - Verify that all the tests pass on both platforms

## Pre-review checklist

## Summary by CodeRabbit

* **New Features**
  * Extended E2E gateway generation/deployment with gateway TLS configuration and an additional elicitation gateway manifest.
* **Bug Fixes**
  * Improved E2E readiness checks to wait for both elicitation and shared-gateway endpoints.
  * Fixed gateway-routed test traffic by disabling host-origin protection where needed.
  * Enhanced trusted CA provisioning for gateway deployments across domain modes.
* **Tests**
  * Updated E2E routing to derive gateway hostnames/TLS listener hostname from `E2E_DOMAIN`, and adjusted multi-gateway isolation/cleanup.
* **Documentation**
  * Updated E2E generated-artifact guidance and clarified TLS/hostname expectations.

### DIFF
--- a/tests/e2e/multi_gateway_test.go
+++ b/tests/e2e/multi_gateway_test.go
@@ -420,7 +420,7 @@ var _ = Describe("MCP Gateway Multi-Gateway", func() {
 		}, TestTimeoutMedium, TestRetryInterval).To(Succeed())
 	})
 
-	It("[multi-gateway] Shared Gateway with team isolation via sectionName", func() {
+	It("[multi-gateway] Shared Gateway with team isolation via sectionName", Serial, func() {
 		// This test verifies that a single Gateway with multiple listeners can be used
 		// by different teams, each with their own MCPGatewayExtension targeting their
 		// specific listener. Each team should only see tools from their own registrations.
@@ -453,7 +453,6 @@ var _ = Describe("MCP Gateway Multi-Gateway", func() {
 			TargetingGateway(SharedGatewayName, GatewayNamespace).
 			WithSectionName(TeamBMCPListenerName).
 			WithPublicHost(TeamBPublicHost).
-			WithListenerPort(8081).
 			Build()
 		teamBSetup.Clean(ctx).Register(ctx)
 		defer teamBSetup.TearDown(ctx)
@@ -619,7 +618,8 @@ var _ = Describe("MCP Gateway Multi-Gateway", func() {
 			WithTarget(SharedGatewayName, GatewayNamespace).
 			WithSectionName(TeamAMCPListenerName). // targeting Team A's listener
 			Build()
-		testResources = append(testResources, conflictExt)
+		// clean up before teamBSetup.TearDown deletes the namespace (LIFO ordering)
+		defer CleanupResource(ctx, k8sClient, conflictExt)
 		Expect(k8sClient.Create(ctx, conflictExt)).To(Succeed())
 
 		Eventually(func(g Gomega) {
@@ -715,7 +715,7 @@ var _ = Describe("MCP Gateway Multi-Gateway", func() {
 		Expect(v.HTTPRouteNotFound(routeName, extNamespace)).To(Succeed())
 	})
 
-	It("[multi-gateway] Each MCPGatewayExtension gets its own HTTPRoute", func() {
+	It("[multi-gateway] Each MCPGatewayExtension gets its own HTTPRoute", Serial, func() {
 		const (
 			teamAExtName = "httproute-team-a"
 			teamBExtName = "httproute-team-b"
@@ -744,7 +744,6 @@ var _ = Describe("MCP Gateway Multi-Gateway", func() {
 			TargetingGateway(SharedGatewayName, GatewayNamespace).
 			WithSectionName(TeamBMCPListenerName).
 			WithPublicHost(TeamBPublicHost).
-			WithListenerPort(8081).
 			Build()
 		teamBSetup.Clean(ctx).Register(ctx)
 		defer teamBSetup.TearDown(ctx)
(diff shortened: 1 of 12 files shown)


trepel:
@coderabbitai review

Patryk-Stefanski (changes requested):
## Blocking Issues

### 1. `URLElicitationGatewayURL` produces wrong URL on OCP

**File**: `tests/e2e/commons.go` line 148

`URLElicitationGatewayURL` is declared but never referenced anywhere in the codebase. More importantly, on OCP (`e2eDomain != defaultE2EDomain`) `gatewayURLDefault` ignores its `kindDefault` argument and derives the URL from the host alone: `e2eScheme + "://" + ElicitationPublicHost + "/mcp"`. Since `ElicitationPublicHost` on OCP resolves to `"elicitation." + e2eDomain`, both `ElicitationGatewayURL` and `URLElicitationGatewayURL` end up as the same URL (`https://elicitation.<domain>/mcp`), losing the `url-elicit` listener distinction entirely.

Since the variable is currently unused this doesn't cause a test failure today, but it signals that URL elicitation on OCP is incomplete infrastructure. Either:

1. Add a dedicated `urlElicitationPublicHostDefault()` that returns `"url-elicit.mcp-gateway.local"` on Kind and `"url-elicit." + e2eDomain` on OCP, and wire it through, or
2. Delete `URLElicitationGatewayURL` until URL elicitation on OCP is actually implemented.

---

### 2. `mcps` listener in elicitation template has hardcoded Kind hostname

**File**: `config/e2e/gateway-elicitation.yaml.template` line 23

The `mcps` listener still has the hardcoded hostname `'*.elicitation.mcp.local'`. On OCP, `envsubst` correctly expands `${E2E_DOMAIN}` for the `mcp` listener and `${GATEWAY_TLS_LISTENER_HOSTNAME}` for `mcp-tls`, but `mcps` keeps a Kind-only hostname. Any test targeting this listener on OCP would silently get the wrong hostname. This was pre-existing in the original `gateway-elicitation.yaml` — the rename to `.template` was the opportunity to fix it.

Suggested fix:
```yaml
- name: mcps
  port: 8080
  protocol: HTTP
  hostname: '*.elicitation.${E2E_DOMAIN}'
```

tests/e2e/multi_gateway_test.go
```diff
@@ -420,7 +420,7 @@ var _ = Describe("MCP Gateway Multi-Gateway", func() {
 		}, TestTimeoutMedium, TestRetryInterval).To(Succeed())
 	})
 
-	It("[multi-gateway] Shared Gateway with team isolation via sectionName", func() {
+	It("[multi-gateway] Shared Gateway with team isolation via sectionName", Serial, func() {
```
david-martin, line 423: pre-existing conflict here between these tests?

david-martin (dismissed):
Generally, good to merge if existing CI passes.
I'll trust that the changes satisfy the OCP testing requirements in the separate test pipeline infra.

1 small clarifying question

tests/e2e/multi_gateway_test.go
trepel: I believe so. This test and the `Each MCPGatewayExtension...` one use same namespaces and both use the `shared-gateway` gateway.

trepel:
@Patryk-Stefanski Thank you so much for the review.

Speaking about 'URLElicitationGatewayURL' bloking issue, my intention was to mimic the KIND setup (using different hostnames for elicitation tests and url-elicitation tets) but I somehow managed to stay half way. I will fix it in 2nd commit. Good catch!

Speaking about `mcps` Listener's hostname, my understanding was that this listener is not public facing one, it is used for internal routing only so it can be arbitrary: https://github.com/Kuadrant/mcp-gateway/blob/main/config/istio/gateway/gateway.yaml#L20-L26

Patryk-Stefanski (approved):

PR #1302, opened 2026-07-24
Design for integrating guard rails into the MCP Gateway

maleck13:
## Adds support for Nemo based Guard Rails integrations

- allows for future expansion and changes to deployment model
- integrates nemo into the router 
- allows config and extension from gateway to mcpserverregistration

## Summary by CodeRabbit

* **Documentation**
  * Added a router-level Guardrails design for inspecting MCP tool calls and elicitation requests and responses.
  * Documented gateway-wide and per-server policies, fail modes, TLS trust, Secret lifecycle behavior, size limits, and streamed response handling.
  * Added documentation plans covering configuration, security, authorization, and operational constraints.

* **Tests**
  * Added test plans for end-to-end, controller, and unit coverage, including policy evaluation, failure handling, response blocking, configuration updates, and Secret deletion or recovery.

### DIFF
--- a/docs/design/guardrails/guardrails-design.md
+++ b/docs/design/guardrails/guardrails-design.md
@@ -1,0 +1,424 @@
+# Guardrails Integration
+
+## Problem
+
+> **Note:** examples reference NeMo Guardrails as the initial target.
+
+Guardrails systems don't support MCP natively. Their endpoints (`v1/chat/completions`, `v1/guardrail/checks`) are incompatible with MCP JSON-RPC. Integrating at the router level provides centralized enforcement and consistent protection across all clients.
+
+## Summary
+
+Translation layer in the router that intercepts `tools/call` requests, translates them to a guardrails check, and either proceeds or rejects. Configuration via Kubernetes Secret referenced from `MCPGatewayExtension` (global) and/or `MCPServerRegistration` (per-server). TLS reuses the gateway's existing CA bundle.
+
+## Goals
+
+- Translate MCP `tools/call` and elicitation responses to guardrails checks
+- Support global and per-server guardrails
+- Fail closed by default, configurable to allow
+- Reuse gateway CA trust pool
+- Target NeMo Guardrails initially
+
+## Non-Goals
+
+- Response guardrailing (deferred)
+- Adding MCP awareness to NeMo itself
+
+## Job Stories
+
+### When I want to enforce safety policies on tool calls
+
+When a platform engineer deploys MCP servers that interact with sensitive systems, they want tool calls checked against guardrails before execution so that dangerous calls are rejected at the gateway.
+
+### When I want guardrails on all servers by default
+
+When a platform engineer wants all tool calls checked, they want to configure guardrails once at the gateway level so every server inherits the policy.
+
+### When I need different guardrails policies per server
+
+When servers have different risk profiles, the platform engineer wants different config IDs per server so policies match capabilities.
+
+### When I want additional guardrails on a specific server
+
+When an MCP server accepts freeform input or interacts with a downstream model, the developer wants server-specific policies in addition to global ones.
+
+### When the guardrails server is down
+
+When the guardrails server is unreachable, tool calls fail closed by default. For non-critical servers, the option to fail open exists.
+
+### When the guardrails server uses a private CA
+
+When the guardrails server uses a corporate or self-signed CA, the platform engineer adds it to the gateway CA bundle.
+
+## Design
+
+### Configuration via Secret
+
+Type `guardrails/external/nemo`, following the project's Secret reference pattern (`credentialRef`, `caCertSecretRef`):
+
+```yaml
+apiVersion: v1
+kind: Secret
+metadata:
+  name: my-guardrails-config
+  namespace: mcp-system
+  labels:
+    mcp.kuadrant.io/secret: "true"
+type: guardrails/external/nemo
+stringData:
+  config.yaml: |
+    url: https://nemo-guardrails.internal:8080
+    configIDs:
+      - "tool-safety-v1"
+      - "input_checking"
+    model: "meta/llama-3.1-8b-instruct"
+    failMode: deny        # deny (default) | allow
+```
+
+| Field | Required | Description |
+|-------|----------|-------------|
+| `url` | yes | Guardrails server endpoint |
+| `configIDs` | no | Default config IDs applied to all servers. Empty if per-server `guardrailsConfigIDs` supply policies |
+| `model` | yes | Model identifier for the guardrails check |
+| `failMode` | no | `deny` (default): reject on failure. `allow`: proceed |
+
+> **Note:** TLS uses the gateway's existing CA bundle (`caCertBundleRef`). Add private CAs there.
+
+### API Changes
+
+#### MCPGatewayExtension
+
+```go
+type MCPGatewayExtensionSpec struct {
+    // ... existing fields ...
+
+    // guardrailsRef references a Secret of type guardrails/external/nemo.
+    // When set, all tools/call requests are checked before routing.
+    // +optional
+    GuardrailsRef *SecretReference `json:"guardrailsRef,omitempty"`
+}
+```
+
+```yaml
+apiVersion: mcp.kuadrant.io/v1alpha1
+kind: MCPGatewayExtension
+metadata:
+  name: mcp-gateway
+spec:
+  targetRef:
+    name: mcp-gateway
+    sectionName: mcp
+  guardrailsRef:
+    name: my-guardrails-config
+```
+
+#### MCPServerRegistration
+
+```go
+type MCPServerRegistrationSpec struct {
+    // ... existing fields ...
+
+    // guardrailsConfigIDs specifies additional config IDs merged with gateway
+    // defaults into a single check request. Requires guardrailsRef on the
+    // MCPGatewayExtension — without it, the server is set to NotReady.
+    // +optional
+    GuardrailsConfigIDs []string `json:"guardrailsConfigIDs,omitempty"`
+}
+```
+
+```yaml
+apiVersion: mcp.kuadrant.io/v1alpha1
+kind: MCPServerRegistration
+metadata:
+  name: dangerous-server
+spec:
+  targetRef:
+    name: dangerous-server-route
+  prefix: dangerous_
+  guardrailsConfigIDs:
+    - "strict-input-checking"
+    - "pii-detection"
+```
+
+#### Validation
+
+If `guardrailsConfigIDs` is set but no `guardrailsRef` exists on the MCPGatewayExtension:
+
+1. MCPServerRegistration set to `NotReady` with reason `GuardrailsNotConfigured`
+2. Server removed from gateway config (no tools/list, tool calls fail)
+3. Re-reconciled when `guardrailsRef` is added
+
+#### Guardrails Secret Deletion
+
+If the guardrails Secret is deleted:
+
+1. MCPGatewayExtension gets condition `GuardrailsSecretNotFound`
+2. The controller retains the last valid guardrails config in `mcp-gateway-config` — `GlobalGuardrails` is not cleared. The router continues to attempt guardrails checks against the configured URL
+3. All MCPServerRegistrations with `guardrailsConfigIDs` are set to `NotReady` and removed from the gateway
+4. For servers without `guardrailsConfigIDs` (covered by global guardrails only), the guardrails server is now unreachable and `failMode` applies: `deny` (default) rejects tool calls, `allow` proceeds without checks
+5. To disable guardrails entirely, the platform engineer must remove `guardrailsRef` from the MCPGatewayExtension — Secret deletion alone does not disable enforcement
+
+This ensures Secret deletion does not silently fail open. The platform engineer's intent (guardrails enabled via `guardrailsRef`) is preserved until explicitly changed. Accidental Secret deletion with `failMode: deny` (the default) blocks tool calls rather than letting unverified requests through.
+
+#### Resolution Order
+
+One guardrails server per gateway. Per-server config IDs are additive.
+
+| Gateway `guardrailsRef` | Server `guardrailsConfigIDs` | Effective Behavior |
+|-------------------------|------------------------------|--------------------|
+| Not set | Not set | No guardrails |
+| Set (with default IDs) | Not set | Gateway defaults applied |
+| Set (empty default IDs) | Not set | No check — per-server-only model |
+| Set (with default IDs) | Set | Gateway defaults + server IDs merged |
+| Set (no default IDs) | Set | Server IDs only |
+| Not set | Set | **NotReady** — server removed |
+
+### Request Flow
+
+Guardrails runs in the router after backend resolution, before the Decision is returned. Applies to `tools/call` and elicitation `accept` responses. If server resolution fails, existing error handling applies — no guardrails check.
+
+**No authorization header forwarding:** The router does not forward the client's `Authorization` header to the guardrails server. NeMo Guardrails does not require client auth — it authenticates to its LLM backend via its own `OPENAI_API_KEY` env var, configured independently on the NeMo server. The guardrails server is deployed without auth enabled (the default) and accessed over in-cluster networking. This avoids exposing client bearer tokens to infrastructure services and keeps credential flows clean: client tokens are for the gateway and upstream MCP servers, not for guardrails.
+
+```mermaid
+sequenceDiagram
+    participant Client
+    participant Envoy
+    participant Router
+    participant Guardrails as Guardrails Server
+    participant Backend as MCP Server
+
+    Client->>Envoy: tools/call (JSON-RPC)
+    Envoy->>Router: ext_proc body phase
+    Router->>Router: parse JSON-RPC, resolve backend
+    Router->>Router: check guardrails config for server
+
+    alt guardrails enabled
+        Router->>Guardrails: POST /v1/guardrail/checks
+        Guardrails-->>Router: response
+
+        alt status: blocked
+            Router-->>Envoy: Decision.Error (403)
+            Envoy-->>Client: JSON-RPC error
+        else status: success
+            Router-->>Envoy: Decision (route to backend)
+            Envoy->>Backend: tools/call
+            Backend-->>Envoy: result
+            Envoy-->>Client: result
+        else non-2xx / timeout
+            alt failMode: deny
+                Router-->>Envoy: Decision.Error (503)
+                Envoy-->>Client: JSON-RPC error
+            else failMode: allow
+                Router-->>Envoy: Decision (route to backend)
+                Envoy->>Backend: tools/call
+            end
+        end
+    else guardrails not enabled
+        Router-->>Envoy: Decision (route to backend)
+        Envoy->>Backend: tools/call
+    end
+```
+
+### Translation Mapping (NeMo)
+
+#### MCP `tools/call` to NeMo `v1/guardrail/checks`
+
+| MCP field | NeMo field |
+|-----------|------------|
+| `params.name` | `messages[0].name` |
+| `params.arguments` (JSON-encoded) | `messages[0].content` |
+| N/A | `messages[0].role = "user"` |
+| N/A | `messages[0].config = "tool"` |
+| from config | `guardrails.config_ids` |
+| from config | `model` |
+
+Example:
+
+```json
+{
+  "model": "meta/llama-3.1-8b-instruct",
+  "messages": [
+    {
+      "role": "user",
+      "name": "execute_sql",
+      "content": "{\"query\": \"DROP TABLE users\"}",
+      "config": "tool"
+    }
+  ],
+  "guardrails": {
+    "config_ids": ["tool-safety-v1", "input_checking"]
+  }
+}
+```
+
+#### Elicitation Response to NeMo
+
+Only `accept` responses are checked (`decline`/`cancel` carry no user content).
+
+| Elicitation field | NeMo field |
+|-------------------|------------|
+| `result.action` | `messages[0].name` |
+| `result` minus `action` (JSON-encoded) | `messages[0].content` |
+| N/A | `messages[0].role = "user"` |
+| N/A | `messages[0].config = "tool"` |
+| from config | `guardrails.config_ids` |
+| from config | `model` |
+
+#### Response to Decision Mapping
+
+| Outcome | Router action |
+|---------|---------------|
+| Translation failure | Always deny with 400. `failMode` does not apply |
+| `status: "success"` | Proceed with routing |
+| `status: "blocked"` | 403 with rails message |
+| Non-2xx HTTP | Apply `failMode`: deny → 503, allow → proceed |
+| Timeout | Apply `failMode`: deny → 503, allow → proceed |
+| Malformed response body | Apply `failMode`: deny → 502, allow → proceed |
+
+Translation failures (e.g. `params.arguments` not valid JSON) are always denied — `failMode: allow` applies only to guardrails server availability, not untranslatable requests.
+
+### TLS Trust
+
+Reuses the gateway's CA trust pool (system roots + `caCertBundleRef`).
+
+### Config Propagation
+
+Controller writes resolved guardrails config into `mcp-gateway-config` Secret. Per-server `guardrailsConfigIDs` go into each server entry:
+
+```go
+type BrokerConfig struct {
+    Servers          []MCPServer           `json:"servers"          yaml:"servers"`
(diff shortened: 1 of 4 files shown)


docs/design/guardrails/guardrails-design.md
```diff
@@ -0,0 +1,423 @@
...
+| Field | Required | Description |
+|-------|----------|-------------|
+| `url` | yes | Guardrails server endpoint |
+| `configIDs` | no | Default config IDs applied to all servers. Empty if per-server `guardrailsConfigIDs` supply policies |
+| `model` | yes | Model identifier for the guardrails check |
+| `failMode` | no | `deny` (default): reject on failure. `allow`: proceed |
+
+> **Note:** TLS uses the gateway's existing CA bundle (`caCertBundleRef`). Add private CAs there.
+
+### API Changes
```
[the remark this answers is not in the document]
maleck13: Not needed in this instance

docs/design/guardrails/guardrails-design.md
```diff
@@ -0,0 +1,423 @@
...
+}
+```
+
+The router constructs a single `*http.Client` for the guardrails server, reused across requests:
+
+- Keep-alive enabled
+- `MaxIdleConnsPerHost` tuned for expected concurrency to avoid connection churn
+- Timeout on transport (`http.Transport.ResponseHeaderTimeout`), not just request context
+
+Client recreated when guardrails config changes.
```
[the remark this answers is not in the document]
maleck13: this is updated

docs/design/guardrails/tasks/documentation.md
```diff
@@ -0,0 +1,76 @@
...
+- Requirement: `guardrailsRef` must exist on MCPGatewayExtension — without it, server is NotReady
+- YAML examples showing combined configuration
+
+### When I want to understand fail modes
+
+When the guardrails server may become unavailable, a platform engineer configuring gateway resilience wants to understand the trade-offs between fail-closed and fail-open so that they can choose the right policy for their risk tolerance.
+
+**Cover:**
+- `failMode: deny` (default) — tool calls rejected when guardrails unreachable
+- `failMode: allow` — tool calls proceed without guardrails check
```
christinaexyou, line 33: a user can optionally configure guardrails to modify/redact unsafe content instead of completely blocking the request. for example, replacing PII with `[MASK]` tokens. thus, we should also include a `failMode: modify`

docs/design/guardrails/tasks/documentation.md
```diff
@@ -0,0 +1,76 @@
...
+- Configuring `configIDs`, `model`, `failMode`
+- Verifying guardrails are active (test a blocked call)
+
+### When I want per-server guardrails policies
+
+When servers have different risk profiles, a platform engineer or MCP server developer managing multiple servers wants to add server-level config IDs alongside the global policy so that each server's guardrails match its capabilities.
+
+**Cover:**
+- Setting `guardrailsConfigIDs` on MCPServerRegistration (not `guardrailsRef`)
+- Additive behavior: global + per-server config IDs merged into a single request
```
christinaexyou: ~this would be difficult to orchestrate from the NeMo Guardrails side. we can either have:~
~* 1 global NeMo server + `x` NeMo servers for each server~
~* 1 NeMo server with both the global and per-server config IDs~

~but both of these options require complex routing logic - for the additive behavior, would we auto-discover the NeMo servers ?~

ignore the above comment, resolved by 1 guardrail server + 1 gateway paradigm. however, an implementation detail here, what order should the config IDs be listed ? for NeMo if a single rail from the configs is trips the execution stops. i think we always ensure that the global config is listed first for a fail-first approach

docs/design/guardrails/tasks/test_cases.md
```diff
@@ -0,0 +1,86 @@
+## Guardrails Integration Test Cases
+
+---
+test_suite: guardrails_test.go
+tags: Happy,Guardrails
+---
+
+> **Note:** E2E tests require a mock guardrails server (deployed in `tests/servers/`) implementing `v1/guardrail/checks` with configurable pass/block responses.
```
christinaexyou, line 8: once support gets added for other endpoints i.e v1/messages, v1/responses we should also add tests for those

docs/design/guardrails/tasks/test_cases.md
```diff
@@ -0,0 +1,86 @@
...
+
+- When an MCPGatewayExtension is configured with `guardrailsRef` and a client makes a `tools/call` with arguments that trigger a block, the gateway should return a JSON-RPC error with a 403 status. The request should never reach the backend MCP server.
+
+### [Happy,Guardrails] Global guardrails allows a safe tool call
+
+- When an MCPGatewayExtension is configured with `guardrailsRef` and a client makes a `tools/call` with arguments that pass the guardrails check, the request should be routed to the backend and the tool result returned.
+
+### [Happy,Guardrails] Per-server guardrailsConfigIDs merged with global defaults
+
+- When an MCPGatewayExtension has `guardrailsRef` with default config IDs and an MCPServerRegistration specifies additional `guardrailsConfigIDs`, tool calls to that server should be checked with all config IDs merged into a single request. Tool calls to other servers should use only gateway defaults.
```
christinaexyou, line 22: this comment is moreso that i remember: we should do the merge at deploy time rather than runtime to avoid latency

docs/design/guardrails/tasks/test_cases.md
```diff
@@ -0,0 +1,86 @@
...
+### Guardrails Secret update triggers config reload
+
+- When the guardrails Secret is updated (e.g. changing `configIDs` or `url`), the controller should detect the change, re-validate, and write the updated config to `mcp-gateway-config`.
+
+### Guardrails Secret deletion clears config and sets all registrations NotReady
+
+- When the guardrails Secret is deleted, the controller should clear `GlobalGuardrails` from `mcp-gateway-config`, set MCPGatewayExtension condition to `GuardrailsSecretNotFound`, and set all MCPServerRegistrations to NotReady. Recreating the Secret should restore normal operation.
+
+## Unit Tests
+
```
christinaexyou, line 111: we should also test api translation between MCP JSON-RPC and NeMo supported formats

docs/design/guardrails/tasks/documentation.md
```diff
@@ -0,0 +1,76 @@
...
+
+## API Reference Updates
+
+### `docs/reference/mcpgatewayextension.md`
+
+**Cover:**
+- `guardrailsRef` field (optional, SecretReference)
+- Secret requirements: type `guardrails/external/nemo`, label `mcp.kuadrant.io/secret=true`
+- Secret key: `config.yaml` (url, configIDs, model, failMode)
+- `configIDs` is optional — may be empty for per-server-only policies
```
christinaexyou: i wonder if we should auto-discover config IDs or at the v least, add some sort of validation/correction bc potentially the user could misspell the config ID

docs/design/guardrails/tasks/documentation.md
```diff
@@ -0,0 +1,76 @@
...
+- Adding the guardrails server CA to the gateway CA bundle (`caCertBundleRef`)
+- No separate CA configuration needed
+- Non-TLS acceptable for in-cluster communication (log warning emitted)
+
+## API Reference Updates
+
+### `docs/reference/mcpgatewayextension.md`
+
+**Cover:**
+- `guardrailsRef` field (optional, SecretReference)
```
christinaexyou: i don't know if a Secret is the right choice of resource bc it prevents us from doing `+kubebuilder` validation checks

docs/design/guardrails/tasks/documentation.md
maleck13: hmm. I had thought of fail mode as "if the service is unreachable what should I do". So perhaps we are both thinking of different things?

docs/design/guardrails/tasks/documentation.md
maleck13: Yes agree, global -> specific order makes sense to me

docs/design/guardrails/tasks/documentation.md
maleck13: We can still validate in the controller based on the type and report issues via the status of the resource we a referencing from. I used a secret initially because I thought we needed auth, but per discussion that isn't the case so open to other options?

docs/design/guardrails/tasks/documentation.md
maleck13: What would that look like? If we sent an invalid configid to nemo would it respond with something useful we could log for visbility? Would this fall under a 4xx response and so trigger the failure mode logic?

docs/design/guardrails/tasks/test_cases.md
maleck13: would the gateway need to call different endpoints or does it just need to call the check endpoint? This piece refers to a mock server for teting to avoid having to deploy the nemo server in CI

docs/design/guardrails/tasks/documentation.md
christinaexyou: oh yes - ignore my comment

docs/design/guardrails/tasks/documentation.md
christinaexyou: yes if a request to the `v1/checks` endpoint contains a `config_id` that doesn't exist / that the system can't locate, it returns status 500 with the reason in detail

christinaexyou: afaik, the only way for the MCPGateway operator to auto-discover config IDs is to grant it permission to view NeMoGuardrails CRs, read the list of NeMo config maps, and do a validation check at runtime to make sure that the `config_id(s)` listed in the request are valid

docs/design/guardrails/tasks/test_cases.md
christinaexyou: the gateway would need to call different endpoints. the `v1/checks` endpoint sits on top of the `v1/chat/completions` endpoint. 

Rob had the idea of making the `v1/checks` endpoint abstract so that it routes requests to different endpoints according to the model provider but I think this is out of scope for 3.6

docs/design/guardrails/tasks/documentation.md
christinaexyou: ack, good point. i was initially thinking that we add type `guardrailsRef` directly to the `MCPGatewayExtension` API where `guardrailsRef` stored the NeMo server url and a reference to the CABundle:

```
type guardrailsRef {
   URL:
   CABundle:
}
``` 

by having `guardrailsRef.url` as a part of the API, this makes it easier us to validate it i.e have the controller check that it is a valid Route. i suppose we could do the same with a Secret resource but imo it seems more convoluted

docs/design/guardrails/guardrails-design.md
```diff
@@ -0,0 +1,423 @@
...
+
+A provider like Praxis could offer guardrails natively. This design allows for that. The controller would translate `guardrailsRef` into provider-specific resources instead of the router performing translation. CRD field stays the same, Secret type differs per provider (example `type: guardrails/internal/praxis` ).
+
+## Open Questions
+
+### Request-side value for schema-constrained tools
+
+Most MCP tools have typed input schemas. AuthPolicy controls access, input validation handles correctness. Guardrails add clear value for freeform text or arguments passed to downstream models. For rigid schemas (`{"namespace": "prod", "replicas": 3}`), the latency cost may not be justified.
+
+### Should guardrails be on the request path, response path, or both?
```
christinaexyou: guardrails should be on both the request and response paths

docs/design/guardrails/guardrails-design.md
```diff
@@ -0,0 +1,423 @@
...
+            Envoy->>Backend: tools/call
+            Backend-->>Envoy: result
+            Envoy-->>Client: result
+        else non-2xx / timeout
+            alt failMode: deny
+                Router-->>Envoy: Decision.Error (503)
+                Envoy-->>Client: JSON-RPC error
+            else failMode: allow
+                Router-->>Envoy: Decision (route to backend)
+                Envoy->>Backend: tools/call
```
christinaexyou, line 207: we also should consider the case where the guardrails status is `modified` but i assume that will follow the same request path as `success`

docs/design/guardrails/guardrails-design.md
maleck13: ok this was an open question in the doc you shared. So it remained that way here for now. Will add some details about how we might be able to achieve this. Likely will be based on https://github.com/Kuadrant/mcp-gateway/pull/1328 where we can configure the response mode if the guadrails config is present

docs/design/guardrails/guardrails-design.md
```diff
@@ -0,0 +1,509 @@
...
+
+## Open Questions
+
+### Request-side value for schema-constrained tools
+
+Guardrails add clear value for freeform text. For rigid schemas (`{"namespace": "prod", "replicas": 3}`), the latency cost may not be justified — AuthPolicy and input validation already cover access and correctness.
+
+### Modified response handling
+
+NeMo can return `status: "modified"` with altered content (e.g. PII redacted). Should the router replace the original body with the modified content, or release the original?
```
david-martin: why would you *not* replace the response body if something was redacted?

docs/design/guardrails/guardrails-design.md
```diff
@@ -0,0 +1,509 @@
...
+
+Single `*http.Client` per guardrails server, reused across requests. Follows the `HairpinClientPool` pattern. Recreated on config change (URL or TLS).
+
+| Setting | Value | Rationale |
+|---------|-------|-----------|
+| Keep-alive | enabled | avoid TCP/TLS per check |
+| `MaxIdleConnsPerHost` | ext_proc stream count | match concurrency |
+| `http.Client.Timeout` | none | per-request `context.WithTimeout` instead |
+| `DialContext` timeout | 1s | fast-fail DNS/TCP hangs |
+| Per-request deadline | 3s | guardrails budget within 10s `message_timeout` |
```
david-martin: is 3s enough if guardrails are potentially using an llm?

shaneutt (dismissed):
Please hold for me.

docs/design/guardrails/guardrails-design.md
maleck13: I am not sure right now what nemo sends back in this case, so it is an open question

docs/design/guardrails/tasks/test_cases.md
RobGeada: @christinaexyou I don't think this is relevant here - we entirely control the caller, meaning we can pass the tool response data in whatever API we want.

maleck13:
Moved to annotation based API to avoid the need for spec / v1 api changes to support something that may / hopefully will ultimately move out of the gateway and into a filter chain

docs/design/guardrails/guardrails-design.md
maleck13: I have a question on the modified response. Is it expected the router will do something with the response body  based on what is returned from Nemo in this case and then send it to the client? I haven't included that in this design yet there is an open question on that below

docs/design/guardrails/guardrails-design.md
```diff
@@ -0,0 +1,509 @@
...
+
+#### Response to Decision Mapping
+
+| Outcome | Router action |
+|---------|---------------|
+| `status: "success"` | Release original event/body to client |
+| `status: "modified"` | **Unresolved** — see Open Questions |
+| `status: "blocked"` | Error response, discard buffer, close stream |
+| Event/body exceeds `maxBodyBytes` | 413 error — general resource limit, not guardrails-specific. `failMode` does not apply |
+| Non-2xx / timeout | Apply `failMode`: deny → error, allow → release |
```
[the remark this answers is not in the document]
maleck13: It is down as an open question for now

docs/design/guardrails/guardrails-design.md
```diff
@@ -0,0 +1,498 @@
...
+
+Single `*http.Client` per guardrails server, following the `HairpinClientPool` pattern. Recreated on config change.
+
+| Setting | Value | Rationale |
+|---------|-------|-----------|
+| Keep-alive | enabled | avoid TCP/TLS per check |
+| `MaxIdleConnsPerHost` | ext_proc stream count | match concurrency |
+| `http.Client.Timeout` | none | per-request `context.WithTimeout` |
+| `DialContext` timeout | 1s | fast-fail DNS/TCP |
+| Per-request deadline | 3s | within 10s `message_timeout` |
```
maleck13, line 389: 3s is a bit arbitrary at the moment. It reflects the need to constrain the timeout within the larger timeout context

docs/design/guardrails/tasks/documentation.md
```diff
@@ -0,0 +1,89 @@
+# Guardrails Integration — Documentation Plan
+
+Documentation for guardrails integration, organized by user goals.
+
+## User-Facing Guide (`docs/guides/guardrails.md`)
+
+### When I want to protect tool calls with guardrails
```
christinaexyou, line 7: should we refer to guardrails that are applied at the Gateway level this as "Global Guardrails Policies" and the per-server ones "Per-Server Guardrails Policies"? 

this doesn't need to be addressed immediately but i think having a standard terminology across all docs will be helpful

docs/design/guardrails/tasks/test_cases.md
```diff
@@ -0,0 +1,94 @@
...
+tags: Happy,Guardrails
+---
+
+> **Note:** E2E tests require a mock guardrails server (deployed in `tests/servers/`) implementing `v1/guardrail/checks` with configurable pass/block responses.
+
+## E2E Tests
+
+### [Happy,Guardrails] Global guardrails blocks a dangerous tool call
+
+- When an MCPGatewayExtension is configured with `guardrailsRef` and a client makes a `tools/call` with arguments that trigger a block, the gateway should return a JSON-RPC error with a 403 status. The request should never reach the backend MCP server.
```
christinaexyou: nit here - update to "When an MCPGatewayExtension is annotated with `mcp.kuadrant.io/guardrails-ref`"

docs/design/guardrails/guardrails-design.md
```diff
@@ -0,0 +1,498 @@
...
+
+`mcp.kuadrant.io/guardrails-config-ids` — comma-separated config IDs merged with gateway defaults. Requires `mcp.kuadrant.io/guardrails-ref` on the MCPGatewayExtension.
+
+```yaml
+apiVersion: mcp.kuadrant.io/v1alpha1
+kind: MCPServerRegistration
+metadata:
+  name: dangerous-server
+  annotations:
+    mcp.kuadrant.io/guardrails-config-ids: "strict-input-checking,pii-detection"
```
christinaexyou, line 132: ~let me look into a better way of referencing all the configIDs~ i think having everything in a single string is 1) tedious for the user to configure and 2) requires us to write failsafes if the string is malformed

docs/design/guardrails/guardrails-design.md
```diff
@@ -0,0 +1,498 @@
...
+
+#### Resolution Order
+
+One guardrails server per gateway. Per-server config IDs are additive — they cannot remove gateway defaults. This lets gateway admins enforce org-wide policies via the Secret's `configIDs` while teams add domain-specific rails for their servers.
+
+| Gateway `guardrails-ref` | Server `guardrails-config-ids` | Effective Behavior |
+|-------------------------|------------------------------|--------------------|
+| Not set | Not set | No guardrails |
+| Set (Secret has default IDs) | Not set | Gateway defaults applied |
+| Set (Secret has empty IDs) | Not set | No check — per-server-only model |
```
christinaexyou: can we remove the `per-server-only model` addendum ? i found it a bit confusing since the effective behaviour here is that the no guardrails policies would be applied

docs/design/guardrails/guardrails-design.md
christinaexyou: no, the router shouldn't behave differently - it should return the modified request/response as is

docs/design/guardrails/guardrails-design.md
```diff
@@ -0,0 +1,498 @@
...
+
+## Open Questions
+
+### Request-side value for schema-constrained tools
+
+For rigid schemas (`{"namespace": "prod", "replicas": 3}`), the latency cost may not be justified — AuthPolicy and input validation already cover access and correctness.
+
+### Modified response handling
+
+NeMo can return `status: "modified"` with altered content (e.g. PII redacted). Should the router replace the original body or release the original?
```
christinaexyou: i wonder if the router can return both i.e it forwards the modified content to the servers but we log the original

docs/design/guardrails/guardrails-design.md
christinaexyou: ideally i'd prefer it in the spec but can we encode it in JSON instead of just plain CSV, i.e: 
```
annotations:
  mcp.kuadrant.io/guardrails-config-ids: '["strict-input-checking","pii-detection"]'
```

pros: easier parsing/validation

docs/design/guardrails/guardrails-design.md
christinaexyou: addressed below !

docs/design/guardrails/guardrails-design.md
christinaexyou: but if not, i think the router should forward along the modified content - main reason being that the NeMo Guardrails PII detection is the only config with modification built-in to it. if users want to use other guardrails types with content modification they would have to write a custom flow. i imagine the user would be pretty miffed if their flow was just completely ignored by the router

docs/design/guardrails/guardrails-design.md
maleck13: Sure we can go with this, it seems less friendly to the user though but I am ok with either of these options

docs/design/guardrails/tasks/documentation.md
maleck13: makes sense

docs/design/guardrails/tasks/test_cases.md
```diff
@@ -0,0 +1,94 @@
...
+
+- When guardrails are configured and a backend MCP server returns a `tools/call` result containing text content that triggers a block, the gateway should return a JSON-RPC error to the client instead of the tool result. The original response body should not reach the client.
+
+### [Guardrails] Response body exceeding maxBodyBytes rejected with 413
+
+- When guardrails are configured and a backend MCP server returns a response body larger than the configured `maxBodyBytes`, the gateway should reject with a 413 JSON-RPC error regardless of `failMode`.
+
+### [Guardrails,Security] Guardrails Secret deletion fails closed
+
+- When an MCPGatewayExtension has `guardrailsRef` and the guardrails Secret is deleted, `GlobalGuardrails` should be cleared from `mcp-gateway-config`, all MCPServerRegistrations should be set to NotReady and removed from the gateway config. Tool calls should fail because no tools are available.
```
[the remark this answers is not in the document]
maleck13: There is no need to do that as the router will translate into the guardrails format for each / any MCP request

docs/design/guardrails/guardrails-design.md
maleck13: So how about the router logs the status it received (so a modified status in this case) from guardrails and forwards the modified response to the caller but not log the actual body? In my mind logging the original and modified body is a concern of Nemo's and should be something that can be discovered in the Nemo logs rather than the routers logs. WDYT

christinaexyou: yeah that makes sense !

christinaexyou (approved):

docs/design/guardrails/tasks/documentation.md
maleck13: ok not sure we want to couple them in that way so would probably just go with the error logging and failure mode to cover this

david-martin (dismissed):

docs/design/guardrails/guardrails-design.md
```diff
@@ -0,0 +1,500 @@
...
+
+### Non-text response content
+
+Only `text` content items are inspected. `image`/`audio` (base64) and `resource` (URI reference) bypass guardrails.
+
+## Execution
+
+See:
+- [tasks/tasks.md](tasks/tasks.md) for the implementation plan (TODO)
```
david-martin: ci failing on this missing file being linked.

docs/design/guardrails/guardrails-design.md
maleck13: updated the design to call this out

docs/design/guardrails/guardrails-design.md
maleck13: ok will remove for now

maleck13:
@christinaexyou Made some small changes (largely all the same content) 

- Added to more specific 2026 protocol test cases
- Called out the modify path we discussed
- Emphasised the need to isolate the guardrails logic into its own package
- removed the 404 doc link


cc @david-martin will need a fresh approval

david-martin (approved):

PR #1310, opened 2026-07-27
feat(resources-federation): add upstream plumbing + broker federation

kkukadia:
## Tasks

- **Task 1**: Upstream resource listing plumbing
  - Add `SupportsResources()` and `ListResources()` to MCP interface
  - Implement live fetching (no caching, unlike tools/prompts)
  
- **Task 2**: Broker `resources/list` federation  
  - Fan out to all upstreams concurrently
  - Merge results with prefix injection into URI authority
  - Skip upstreams that error/timeout/have no prefix

## Implementation details

- Resources are fetched live per request (not cached)
- Prefix re-validation against `[a-z0-9_]+` allowlist
- `GetServerInfoByResource()` for longest-prefix-match resolution
- Pagination logged but not followed (observable partial lists)

## Dependencies

Blocks: #1356, #1357

## Testing

- Unit tests: existing patterns from `ListTools`/`ListPrompts`
- E2E: resources/list merges from multiple upstreams


maleck13:
There was a missing image in CI so rebasing should solve the build issue

kkukadia:
> There was a missing image in CI so rebasing should solve the build issue

thank you 🙏🏽

internal/broker/broker.go
```diff
@@ -747,6 +812,119 @@ func (m *mcpBrokerImpl) ValidateAllServers() StatusResponse {
...
+			continue
+		}
+		if !resourcePrefixAllowlist.MatchString(cfg.Prefix) {
+			m.logger.Error("skipping resources/list for server with invalid prefix", "server", srv.MCPName(), "prefix", cfg.Prefix)
+			continue
+		}
+
+		g.Go(func() error {
+			resources, err := m.fetchResourcesFromServer(gCtx, srv, cfg.Prefix)
+			if err != nil {
```
Patryk-Stefanski, line 904: **[blocking] SDK resource cache corrupted by in-place URI mutation**

`srv.ListResources()` delegates to the go-sdk `ClientSession.ListResources()`, which (in v1.7.0, `client.go:1320`) unconditionally stores the returned `*ListResourcesResult` in a per-session TTL cache and returns the exact same pointer on a cache hit when the upstream sets `ttlMs > 0`. Because `result.Resources` is `[]*mcp.Resource`, each `r` in the `for _, r := range` loop points into the cached struct, and `r.URI = rewriteResourceURI(...)` mutates it in place. The next call returns the already-prefixed URI, the prefix is injected again (`ui://pfx_pfx_template.html`), and every subsequent cached fetch is wrong.

Fix — copy before rewriting:

```go
for i, r := range result.Resources {
    if r == nil {
        continue
    }
    copied := *r
    copied.URI = rewriteResourceURI(r.URI, prefix)
    result.Resources[i] = &copied
}
```

internal/broker/broker.go
```diff
@@ -747,6 +812,119 @@ func (m *mcpBrokerImpl) ValidateAllServers() StatusResponse {
...
+				m.logger.Error("failed to fetch resources", "server", srv.MCPName(), "error", err)
+				return nil // graceful degradation: log and exclude, don't fail the request
```
Patryk-Stefanski, line 906: **[blocking] Upstream credential leakage via Userinfo in ui:// URIs**

`url.Parse` preserves the `Userinfo` component; `u.String()` re-serialises it. An upstream returning `ui://secret@host/template.html` produces `ui://secret@pfx_host/template.html` in the `resources/list` response — visible to every client.

```go
// verified:
u, _ := url.Parse("ui://secret@somehost/template.html")
u.Host = "pfx_" + u.Host
u.String() // => "ui://secret@pfx_somehost/template.html"
```

Fix — strip before rewriting:

```go
func rewriteResourceURI(uri, prefix string) string {
    u, err := url.Parse(uri)
    if err != nil || u.Scheme != "ui" {
        return uri
    }
    u.User = nil // never forward upstream credentials to clients
    u.Host = prefix + u.Host
    return u.String()
}
```

Patryk-Stefanski, line 906: **[blocking] Upstream can spoof another server's resource namespace**

`rewriteResourceURI` concatenates the prefix directly onto `u.Host` with no separator enforcement. An upstream with prefix `a_` can return `ui://b_victim.html`; after rewrite this becomes `ui://a_b_victim.html`. If a server with prefix `a_b_` exists, `GetServerInfoByResource`'s longest-prefix match attributes the URI to that server, letting upstream `a_` route `resources/read` calls to a different backend.

In practice, all prefixes in the codebase end with `_`, which prevents this — but `resourcePrefixAllowlist` (`^[a-z0-9_]+$`) does not enforce the trailing underscore.

Fix: require the trailing separator in the allowlist so no two prefixes can form an ambiguous concatenation:

```go
var resourcePrefixAllowlist = regexp.MustCompile(`^[a-z0-9]+_$`)
```

Patryk-Stefanski (commented):
Reviewed in three independent passes (correctness, security, test coverage). Seven findings below — four blocking.

All blocking issues are in `broker.go`; the first three are security/correctness bugs in `rewriteResourceURI` and the fourth is a test that does not exercise the code path it documents.

internal/broker/resources_test.go
```diff
@@ -0,0 +1,237 @@
...
+func (h *recordingHandler) Handle(_ context.Context, r slog.Record) error {
+	h.messages = append(h.messages, r.Message)
+	return nil
+}
+func (h *recordingHandler) WithAttrs(_ []slog.Attr) slog.Handler { return h }
+func (h *recordingHandler) WithGroup(_ string) slog.Handler      { return h }
+
+func (h *recordingHandler) hasMessage(substr string) bool {
+	for _, m := range h.messages {
+		if strings.Contains(m, substr) {
```
Patryk-Stefanski, line 236: **[blocking] Test does not exercise the nil-result guard it documents**

`mockActiveServer.ListResources` (in `version_test.go`) falls through to `return &mcp.ListResourcesResult{}, nil` when both `listResourcesResult == nil` and `listResourcesErr == nil` — it never returns `(nil, nil)`. So `TestFetchResourcesFromServer_NilResult` goes through the `if result == nil` guard with a non-nil empty struct, meaning the panic-prevention code added in commit 2 is never actually exercised by this test.

Fix — use a dedicated mock (or a flag) that returns `(nil, nil)`:

```go
type nilResultServer struct{ mockActiveServer }
func (n *nilResultServer) ListResources(context.Context) (*mcp.ListResourcesResult, error) {
    return nil, nil
}

func TestFetchResourcesFromServer_NilResult(t *testing.T) {
    b := newResourcesTestBroker(5 * time.Second)
    resources, err := b.fetchResourcesFromServer(context.Background(), &nilResultServer{}, "pfx_")
    require.NoError(t, err)
    assert.Nil(t, resources)
}
```

internal/broker/broker.go
```diff
@@ -747,6 +812,119 @@ func (m *mcpBrokerImpl) ValidateAllServers() StatusResponse {
...
 }
 
+// FetchResources populates result with resources fetched live from every
+// upstream that supports them. Unlike tools/prompts, nothing is
+// pre-registered on the gateway server, so this builds the entire result
+// from scratch on every call rather than augmenting/filtering an existing
+// one. Upstreams are fanned out to concurrently (mirroring
+// FetchUserSpecificTools in user_specific_tools.go): one goroutine per
+// upstream, each with its own timeout, so one slow or down server doesn't
+// add its latency to every other server's contribution. A failing upstream
```
Patryk-Stefanski, line 866: **[nit] `FetchResources` godoc contradicts the implementation**

The comment says "builds the entire result from scratch" but the function does `result.Resources = append(result.Resources, allResources...)`. Currently safe because the SDK always passes an empty slice, but the comment will mislead any future caller who passes a pre-seeded result (silently duplicated resources, no warning).

Suggest: "Caller must pass a result with an empty Resources slice; this function appends to it." Or use assignment instead of append since the intent is to populate from scratch:

```go
result.Resources = allResources
```

internal/broker/broker.go
```diff
@@ -657,6 +677,51 @@ func (m *mcpBrokerImpl) GetServerInfoByPrompt(prompt string) (*config.MCPServer,
...
+	m.mcpLock.RLock()
+	defer m.mcpLock.RUnlock()
+
+	authority := resourceAuthority(uri)
+
+	var bestMatch config.MCPServer
+	var found bool
+	for _, upstream := range m.mcpServers {
+		cfg := upstream.Config()
+		if cfg.Prefix != "" && strings.HasPrefix(authority, cfg.Prefix) {
```
Patryk-Stefanski, line 740: **[nit] `GetServerInfoByResource` always fails for non-`ui://` resources — undocumented**

`resourceAuthority("https://example.com/doc.html")` returns `"example.com"`, which never matches any `[a-z0-9_]+_` prefix. Non-`ui://` resources appear in `resources/list` (passed through unrewritten by `rewriteResourceURI`) but `GetServerInfoByResource` always returns an error for them. When `resources/read` routing lands in the next PR, this will silently fail for every non-`ui://` resource with no explanation.

Suggest adding to the godoc: *"Only `ui://` URIs can be resolved; non-`ui://` URIs are returned by FetchResources unrewritten and will always produce an error here."*

internal/broker/broker.go
```diff
@@ -747,6 +812,119 @@ func (m *mcpBrokerImpl) ValidateAllServers() StatusResponse {
...
+	servers := make([]upstream.ActiveMCPServer, 0, len(m.mcpServers))
+	for _, srv := range m.mcpServers {
+		servers = append(servers, srv)
+	}
+	m.mcpLock.RUnlock()
+
+	ctx, span := brokerTracer().Start(ctx, "broker.resources.fetch-all",
+		trace.WithAttributes(
+			attribute.Int("mcp.resources.server_count", len(servers)),
+		),
```
Patryk-Stefanski, line 879: **[nit] Missing comment on why `resources/list` skips header extraction**

`tools/list` and `prompts/list` both extract `req.GetExtra().Header` for per-user filtering. The `resources/list` case silently omits this. The omission is intentional (resources are not user-specific in this phase), but there is no comment to explain it. A future contributor wiring per-user resource filtering will not know whether this was deliberate or an oversight.

Suggest adding:

```go
case "resources/list":
    // no header extraction — resources are not user-specific in this phase
    m.FetchResources(ctx, resourcesResult)
```

kkukadia:
> Reviewed in three independent passes (correctness, security, test coverage). Seven findings below — four blocking.
> 
> All blocking issues are in `broker.go`; the first three are security/correctness bugs in `rewriteResourceURI` and the fourth is a test that does not exercise the code path it documents.

 Thank you for the review! Applied all 7 fixes using your exact solutions.

internal/broker/broker.go
```diff
@@ -747,6 +815,123 @@ func (m *mcpBrokerImpl) ValidateAllServers() StatusResponse {
...
+			allResources = append(allResources, resources...)
+			mu.Unlock()
+			return nil
+		})
+	}
+	_ = g.Wait()
+
+	span.SetAttributes(attribute.Int("mcp.resources.resources_fetched", len(allResources)))
+	result.Resources = allResources
+}
```
Patryk-Stefanski, line 918: **Bug: null resources array when no upstreams qualify**

`allResources` is declared as `var allResources []*mcp.Resource`, which initialises it to `nil` in Go. If every upstream is skipped (no prefix, invalid prefix, none supports resources), this line assigns `nil` into `result.Resources`, overwriting the empty slice the SDK had already set there to prevent this exact case. The result serialises as `{"resources":null}` instead of `{"resources":[]}`, which violates the MCP spec and will confuse clients that distinguish the two.

Fix: initialise as a non-nil slice before the fan-out:
```go
allResources := make([]*mcp.Resource, 0)
```

internal/broker/broker.go
```diff
@@ -747,6 +815,123 @@ func (m *mcpBrokerImpl) ValidateAllServers() StatusResponse {
...
+	}
+
+	for i, r := range result.Resources {
+		if r == nil {
+			continue
+		}
+		copied := *r
+		copied.URI = rewriteResourceURI(r.URI, prefix)
+		result.Resources[i] = &copied
+	}
```
Patryk-Stefanski, line 958: **Bug: mutates cached slice, causing double-prefix on repeated calls**

`srv.ListResources` caches its result by TTL. When TTL > 0, a second call returns the **same `*ListResourcesResult` pointer**. Writing `result.Resources[i] = &copied` mutates the slice inside that cached object, so the next request finds already-prefixed URIs and prefixes them again: `ui://pfx_pfx_resource.html`.

The struct copy (`copied := *r`) avoids mutating the individual `*Resource` values but does not protect the slice itself. Fix by building a new output slice instead of writing back into `result.Resources`:
```go
out := make([]*mcp.Resource, 0, len(result.Resources))
for _, r := range result.Resources {
    if r == nil {
        continue
    }
    copied := *r
    copied.URI = rewriteResourceURI(r.URI, prefix)
    out = append(out, &copied)
}
return out, nil
```

Patryk-Stefanski:
@david-martin  — question on the CRD prefix validation pattern.

Currently `MCPServerRegistrationSpec.prefix` has:

\`\`\`
// +kubebuilder:validation:Pattern=`^[a-z0-9][a-z0-9_]*$`
\`\`\`

This accepts values without a trailing underscore (e.g. \`fast_slow\`). However, the broker requires a trailing underscore because it concatenates the prefix directly onto the resource name/URI with no separator:

\`\`\`go
// rewriteResourceURI — broker.go
u.Host = prefix + u.Host  // ui://template.html -> ui://<prefix>template.html
\`\`\`

Without the trailing \`_\`, a prefix like \`fast_slow\` would produce \`ui://fast_slowtemplate.html\` — the prefix bleeds into the resource name with no delimiter, making the URI ambiguous and breaking \`ResolveResourceServer\` (which uses \`strings.HasPrefix(authority, cfg.Prefix)\` to route resources back to the correct upstream).

Two options:

1. **Tighten the CRD pattern** to \`^[a-z0-9][a-z0-9_]*_$\` — enforces trailing underscore at the API boundary. Single place to enforce the invariant.
2. **Leave the CRD as-is, inject the separator in the broker** — change every concatenation site to \`prefix + "_" + name\`. Works but the invariant is now distributed across multiple callsites and easy to miss in future changes.

Option 1 seems cleaner. What do you think?

internal/broker/broker.go
kkukadia: applied suggested fix 🙏

internal/broker/broker.go
kkukadia: thank you for catching this, applied the fix 🙏

david-martin:
I'm inclined to keep the CRD validation as is (backwards compatible).
We could have a different validation for resources, but then it differs between tools/prompts and resources.

This pushes the decision about prefixing back on the user as they may not necessarily want an underscore

internal/broker/broker.go
```diff
@@ -698,6 +721,42 @@ func (m *mcpBrokerImpl) GetServerInfoByPrompt(prompt string) (*config.MCPServer,
...
+	defer m.mcpLock.RUnlock()
+
+	authority := routing.ResourceAuthority(uri)
+
+	var bestMatch config.MCPServer
+	var found bool
+	for _, upstream := range m.mcpServers {
+		cfg := upstream.Config()
+		if cfg.Prefix != "" && strings.HasPrefix(authority, cfg.Prefix) {
+			if !found || len(cfg.Prefix) > len(bestMatch.Prefix) {
```
Patryk-Stefanski, line 743: **GetServerInfoByResource does not apply ensureSeparator — mismatch with rewriteResourceURI**

`rewriteResourceURI` builds URIs using `ensureSeparator(prefix) + u.Host`, so a prefix `my` rewrites `ui://admin_thing.html` to `ui://my_admin_thing.html`. But this lookup checks `strings.HasPrefix(authority, cfg.Prefix)` against the raw prefix — no separator applied.

Concrete failure (once resources/read is wired in #1339): server with prefix `my` returns `ui://admin_thing.html`, broker rewrites it to `ui://my_admin_thing.html`. A client calls `resources/read` for that URI. `GetServerInfoByResource` checks `HasPrefix("my_admin_thing.html", "my")` (true, len 2) and `HasPrefix("my_admin_thing.html", "my_admin_")` (true, len 9 — wins). Resource attributed to the wrong server.

No impact today — `resources/read` is still blocked with HTTP 400. But this will be a live routing bug when #1339 lands.

Fix is one line:
```go
if cfg.Prefix != "" && strings.HasPrefix(authority, ensureSeparator(cfg.Prefix)) {
```

Patryk-Stefanski:
## E2E failure is caused by `checkPrefixUniqueness()` introduced in this PR

The failing test is `[Happy] should become Ready even when tool conflicts exist from same prefix` (`happy_path_test.go:1039`). This is not pre-existing infrastructure flakiness — it's deterministically broken by `checkPrefixUniqueness()` added to the controller reconciler in this PR.

**What the test asserts:**

Two `MCPServerRegistration`s with the same prefix (`conflict_`) should both reach `Ready=True`. The test comment says it explicitly: _"conflicts are detected by the broker at runtime, not by the controller"_.

**What `checkPrefixUniqueness()` does:**

When the second registration arrives, the controller sees the prefix is already in use and sets `Ready=False` with a `PrefixConflict` condition. The test waits 300s for `Ready=True` and times out.

**Why we should remove `checkPrefixUniqueness()`:**

The broker already handles this correctly. `findToolConflicts()` and `findPromptConflicts()` detect actual name collisions at registration time and surface them in the `/status` endpoint. Both registrations still reach `Ready=True` in Kubernetes — which is correct, because the *registration* is valid. The conflict is a runtime broker concern, not a controller concern.

The controller check is also the wrong abstraction: two servers with the same prefix only actually conflict if they expose overlapping tool names. The broker knows that. The controller doesn't — it's rejecting registrations based on a proxy signal (same prefix string) rather than the real condition (conflicting tool names).

**The fix:**

Remove `checkPrefixUniqueness()` and its call site from the controller. The e2e test passes again without modification, and conflict detection stays where it belongs — in the broker.

Patryk-Stefanski (approved):

PR #1332, opened 2026-07-31
fix: propagate upstream 4xx status codes during lazy init

### LINKED ISSUE #1331: Lazy init wraps upstream 4xx responses as 500
When a hairpin initialize request during lazy init receives a 4xx from upstream (e.g. Authorino returning 403 for unauthorized tool access), the router wraps it as a 500. The original status code is lost.

The chain:

1. The MCP SDK's streamable transport wraps any 4xx as a generic `"transport error: server returned 4xx"` string
2. `internal/clients/clients.go` wraps that as `"failed to create client"`
3. `internal/routing/router_202511.go:524` unconditionally wraps all `InitForClient` errors as 500

Clients see HTTP 500 with `"failed to create session for mcp server"` when they should see the original status code and upstream error message (e.g. 403 with the Authorino denial reason).

david-martin:
## Summary

- Adds `StatusCapturingRoundTripper` to intercept non-2xx HTTP responses before the MCP SDK wraps them as untyped strings, preserving the status code in a typed `HTTPStatusError`
- Router now extracts 4xx status codes from `InitForClient` errors and propagates them to the client instead of always returning 500
- 5xx and transport-level errors (connection refused, timeouts) still become 500

## Context

When Authorino denies a hairpin initialize request (e.g. user lacks tool access), the 403 and its denial message were lost through three layers of error wrapping (SDK transport, `clients.Initialize`, router). Clients saw HTTP 500 with a generic "failed to create session" message instead of the 403 with the actual denial reason.

Fixes #1331

## Test plan

- [x] Unit tests for `StatusCapturingRoundTripper` (2xx passthrough, 4xx/5xx typed error)
- [x] Router test covering 403, 401, 400 propagation and 502/connection-refused wrapping as 500
- [x] Full unit test suite passes
- [x] Lint clean (no new issues)

## Summary by CodeRabbit

* **Bug Fixes**
  * Preserved upstream HTTP 4xx status codes during backend session initialization, improving error accuracy.
  * Included relevant upstream response details in reported errors, with response content safely limited for reliability.
  * Continued mapping server-side and connection failures to a generic 500 status.
  * Improved retry handling for custom TLS connections by reliably recreating clients after failed initialization attempts.

### DIFF
--- a/internal/routing/router_202511.go
+++ b/internal/routing/router_202511.go
@@ -524,7 +524,7 @@ func (r *Router202511) initializeMCPServerSession(ctx context.Context, mcpReq *M
 			mcpotel.SpanError(initSpan, err, "failed to initialize backend session")
 			var httpErr *transport.HTTPStatusError
 			if errors.As(err, &httpErr) && httpErr.Code >= 400 && httpErr.Code < 500 {
-				return "", NewRouterErrorf(int32(httpErr.Code), "failed to create session for mcp server: %w", err)
+				return "", NewRouterError(int32(httpErr.Code), fmt.Errorf("failed to create session for mcp server: %w", err)) //nolint:gosec,nolintlint // code bounded to [400,499] by check above
 			}
 			return "", NewRouterErrorf(500, "failed to create session for mcp server: %w", err)
 		}
--- a/internal/routing/router_202511.go
+++ b/internal/routing/router_202511.go
@@ -18,6 +18,7 @@ import (
 	internaljwt "github.com/Kuadrant/mcp-gateway/internal/jwt"
 	mcpotel "github.com/Kuadrant/mcp-gateway/internal/otel"
 	"github.com/Kuadrant/mcp-gateway/internal/session"
+	"github.com/Kuadrant/mcp-gateway/internal/transport"
 	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/codes"
 	"go.opentelemetry.io/otel/trace"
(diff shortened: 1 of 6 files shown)


david-martin:
Tested live in an OCP cluster as well, swapping in a custom image build with this fix.

internal/routing/router_202511.go
```diff
@@ -521,6 +522,10 @@ func (r *Router202511) initializeMCPServerSession(ctx context.Context, mcpReq *M
 		if err != nil {
 			r.Logger.ErrorContext(ctx, "failed to get remote session ", "error", err)
 			mcpotel.SpanError(initSpan, err, "failed to initialize backend session")
+			var httpErr *transport.HTTPStatusError
+			if errors.As(err, &httpErr) && httpErr.Code >= 400 && httpErr.Code < 500 {
+				return "", NewRouterErrorf(int32(httpErr.Code), "failed to create session for mcp server: %w", err)
+			}
```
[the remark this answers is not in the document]
david-martin: Re: integration test through real SDK error chain (router_202511.go comment): verified the full error chain end-to-end on an OCP cluster. StatusCapturingRoundTripper returns the typed error, the SDK wraps it with %w, and errors.As extracts it in the router. An httptest.Server-based test would be a useful canary if the SDK ever changes how it wraps errors, but it is coupling to SDK internals rather than testing our code. Skipping for now.

internal/routing/router_202511.go
```diff
@@ -521,6 +522,10 @@ func (r *Router202511) initializeMCPServerSession(ctx context.Context, mcpReq *M
 		if err != nil {
 			r.Logger.ErrorContext(ctx, "failed to get remote session ", "error", err)
 			mcpotel.SpanError(initSpan, err, "failed to initialize backend session")
+			var httpErr *transport.HTTPStatusError
+			if errors.As(err, &httpErr) && httpErr.Code >= 400 && httpErr.Code < 500 {
```
maleck13, line 526: nit: might be nice to have options on HTTPStatusError like: ```httpErr.Is4xx()``` just an option though

david-martin: Single call site for the range check, so I'd leave it inline for now. If we add more status classification logic later we can pull it out then.

maleck13 (approved):

PR #1347, opened 2026-08-06
feat(router): structured audit log for tool calls

Patryk-Stefanski:
## Summary

- Adds structured `Info`-level audit log entries in the ext_proc adapter for `tools/call` requests
- Fields on every entry: `audit` (always `true`), `user` (JWT sub from Authorization header), `tool`, `server`, `status`, `request_id`, `session`
- No new packages, types, or config

## Motivation

The existing Istio-based auditing approach (docs/guides/auditing.md) patches the Istio CR with a custom `extensionProvider`. On OpenShift 4.19–4.21 the Cluster Ingress Operator overwrites that CR; on 4.22+ the Istio CR is absent entirely (Istio installed via Helm through OSSM). This provides a self-contained audit trail in router logs independent of the Envoy/Istio pipeline.

Closes CONNLINK-1385

## What gets logged and when

The router emits a `level=INFO msg="tool call" audit=true` entry at two sites:

**1. Response headers phase** — tool call reached the backend and got a response:
```
level=INFO msg="tool call" audit=true user=alice@example.com tool=echo server=mcp-test/my-server status=200 request_id=<uuid> session=jti:...
```

**2. Request body phase** — router returned an error before forwarding (e.g. session init failure, unknown server):
```
level=INFO msg="tool call" audit=true user=alice@example.com tool=echo server="" status=401 request_id=<uuid> session=sha256:...
```

**Out of scope** (not logged — no tool identity available or not the router's responsibility):
- AuthPolicy denials — Envoy rejects before ext_proc is involved
- Invalid JSON-RPC / unparseable body — rejected before `mcpRequest` is populated
- Request body too large — rejected before parse
- `tools/list`, `initialize`, and all other non-`tools/call` methods — intentional

## How to review

1. `internal/mcp-router/ext_proc_adapter.go` — two `logger.InfoContext` blocks with `audit=true`:
   - After `respDecision := respHandler.HandleResponse(...)` in `ProcessingRequest_ResponseHeaders` (successful path)
   - After `decision := router.RouteRequest(...)` in `ProcessingRequest_RequestBody` when `decision.Error != nil` (error path)
2. `internal/mcp-router/elicitation.go` — elicitation rewrite log changed from Info to Debug
3. `internal/mcp-router/ext_proc_adapter_test.go` — `TestProcess_ToolCallAuditLog` (success path) and `TestProcess_ToolCallAuditLog_RouterError` (error path)
4. `docs/guides/auditing.md` — router-native approach now leads; Istio Telemetry approach moved to a secondary section with OpenShift compatibility caveat

## Test plan

- [x] `make test-unit` passes
- [x] `make lint` passes
- [ ] Manual verification on Kind cluster

## Summary by CodeRabbit

* **New Features**
  * Added structured audit logs for tool-call routing and responses, including user identity, tool, server, status, request ID, and session details.
  * Audit records now cover both successful requests and routing errors.

* **Documentation**
  * Updated the auditing guide with structured log fields, phases, exclusions, querying, production shipping, and platform limitations.
  * Simplified audit verification and configuration guidance.

* **Improvements**
  * Reduced the verbosity of elicitation ID rewrite messages by logging them at debug level.


david-martin:
few small things:

1. The PR test plan in the description references a scenario in tests/manual-testcases/0.7.0.md (even though that file is removed). For testing, I think we could defer to....
2. Docs: Update docs/guides/auditing.md. The motivation is that the existing Istio-based approach breaks on OpenShift 4.19+. lets update the guide to use this router-native alternative (and downgrade/relegate the istio cr approach in that guide as an alternative)
3. Can we add a key (e.g. audit=true) to distinguish audit entries from other Info logs. Filtering by msg="tool call" is fragile if more Info lines are added later.
4. and related to that, looks like everywhere except [internal/mcp-router/elicitation.go](https://github.com/Kuadrant/mcp-gateway/blob/main/internal/mcp-router/elicitation.go#L108) uses debug. Let's change this to debug as well so that info level is the defacto for audit logs.

david-martin (approved):
Thanks, all my feedback has been addressed. LGTM

PR #1356, opened 2026-08-10
feat(resources-federation): add router resources/read routing

kkukadia:
## Task 4: Router `resources/read` routing

Routes `resources/read` requests to the upstream that owns the requested URI via longest-prefix-match against registered resource-federated servers.

## Implementation

- `HandleResourceRead()` in router request handlers (mirrors `HandleToolCall` pattern)
- Reuses existing `validateSession()` session validation
- `GetServerInfoByResource()` from #1310 for prefix resolution
- Strips prefix before forwarding to upstream
- Sets `x-mcp-servername` for AuthPolicy enforcement

## Changes

- `internal/routing/router_202511.go`: 
  - Add `routeResourceRead()` function
  - Add `resources/read` case to `RouteRequest()`
  - Helper: `resourceAuthority()`, `stripResourcePrefix()`
- Unit tests with session validation, prefix matching, error cases
- E2E tests for round-trip, prefix matching, unknown prefix errors

## Dependencies

- Depends on: #1310 (Task 2: `GetServerInfoByResource`)
- Independent of: #1357 (Task 3: auth filtering operates at broker, this is router)

## Status

Ready for review once #1310 merges to main (will rebase at that time)


kkukadia:
@Patryk-Stefanski thanks for helping me to figure out what's wrong, such a great learning experience :)

Patryk-Stefanski (dismissed):
Two blocking bugs found — see inline comments.

Patryk-Stefanski (commented):
Reviewed the delta from #1310. The core implementation is correct:

- `Router202511.routeResourceRead` correctly calls `LookupResourcePrefix` (not `LookupResource`) and sets `ResourceHeader = upstreamURI` (stripped form)
- The data race on `SupportsResources` is fixed here with `clientMu.RLock()`

One nit inline on `StripPrefixFromResourceURI` — no scheme guard. Everything else looks good.

internal/routing/router_202511.go
```diff
@@ -261,6 +265,83 @@ func (r *Router202511) routePromptGet(ctx context.Context, table RoutingTable, m
...
+
+// stripResourcePrefix removes prefix from a ui:// URI's authority segment,
+// reconstructing the original upstream URI - the inverse of the broker's
+// rewriteResourceURI. Non-ui:// and malformed URIs are returned unchanged.
+func stripResourcePrefix(uri, prefix string) string {
+	u, err := url.Parse(uri)
+	if err != nil || u.Scheme != "ui" {
+		return uri
+	}
+	u.Host = strings.TrimPrefix(u.Host, prefix)
```
Patryk-Stefanski: ## Bug — `resources/read` corrupts the upstream URI for any prefix without a trailing `_`

`stripResourcePrefix` strips the **raw** prefix, but the broker injects the prefix **with a separator** via `ensureSeparator`. The two sides are asymmetric.

**Broker write side** (`internal/broker/broker.go:978`):
```go
u.Host = ensureSeparator(prefix) + u.Host   // prefix "foo" -> ui://foo_file.html
```
The broker's own read/match side agrees (`broker.go:742`): `strings.HasPrefix(authority, ensureSeparator(cfg.Prefix))`.

**Router strip side** (here, `router_202511.go:341`):
```go
u.Host = strings.TrimPrefix(u.Host, prefix) // strips only "foo" -> ui://_file.html
```

**Failure:** a server registered with prefix `foo` (valid per the CRD pattern `^[a-z0-9][a-z0-9_]*$`) makes the broker emit `ui://foo_file.html` to the client. On `resources/read`, the router does `TrimPrefix("foo_file.html", "foo")` = `_file.html`, forwarding `ui://_file.html` upstream — a resource the server doesn't own. The read fails. Tools/prompts are unaffected because only resources go through `ensureSeparator`. This breaks every prefix that doesn't already end in `_`.

**Fix:** strip `ensureSeparator(prefix)`, not the raw prefix. Since `ensureSeparator` is broker-only today, either lift it to a shared helper or replicate the one-liner router-side.

**Test gap:** `TestHandleResourceRead_SingleCharPrefix` (`internal/routing/router_test.go:1940`) masks this. With `Prefix: "a"` it hard-codes the client URI as `ui://afile.html` and asserts strip to `ui://file.html` — but the broker with prefix `a` actually emits `ui://a_file.html` (`ensureSeparator("a")` = `"a_"`). Feed the real broker output through the router and you get `ui://_file.html`. The other passing tests use prefixes like `s_`/`gh_` where the separator is already inside the prefix string, so none exercise the gap. The test should round-trip against `rewriteResourceURI`'s real output.

Patryk-Stefanski:
I think we can remove the [‎repro](https://github.com/Kuadrant/mcp-gateway/pull/1356/changes#diff-564d1f19f8bb6fa6bbc659d9dba111b939456860d46f477202b5d4f53f577b03) folder from this PR

internal/routing/mcp_request.go
```diff
@@ -289,3 +320,15 @@ func ResourceAuthority(uri string) string {
 	}
 	return u.Host
 }
+
+// ensureSeparator returns prefix with trailing underscore, adding one if needed.
+// Empty prefix is returned unchanged.
+func ensureSeparator(prefix string) string {
```
Patryk-Stefanski: Two non-blocking follow-ups on the resources/read routing (nice work fixing the separator round-trip 👍):

**1. `ensureSeparator` is now duplicated.** This copy is identical to the one in `internal/broker/broker.go:958`. The broker already imports `internal/routing` (it calls `routing.ResourceAuthority` at `broker.go:736`), so the two can silently diverge — if the separator convention ever changes, one side gets updated and the broker-rewrite/router-strip round-trip breaks. Suggest exporting this as `routing.EnsureSeparator` and having `broker.go` reuse it, mirroring how `ResourceAuthority` is already shared.

**2. No e2e coverage for `resources/read`.** The PR adds only unit tests. `.claude/rules/ext-proc-handlers.md` asks for e2e coverage on router/ext_proc routing changes, and the round-trip through Envoy (session hairpin init, `:authority` rewrite, AuthPolicy via `x-mcp-resourceuri`) isn't exercised end-to-end — a regression in the Envoy hop or AuthPolicy interaction would pass all unit tests. Either add an e2e test alongside the existing tool/prompt routing tests, or update the PR description if e2e is intentionally deferred.

kkukadia: thank you! on the follow-ups right away.

Patryk-Stefanski (approved):

PR #1362, opened 2026-08-11
fix: handle SIGTERM so graceful shutdown runs in Kubernetes

### LINKED ISSUE #1351: broker/router: SIGTERM is not handled, graceful shutdown never runs under Kubernetes
## Description

`run()` registers only `os.Interrupt` for shutdown (`cmd/mcp-broker-router/main.go:276`):

```go
stop := make(chan os.Signal, 1)
signal.Notify(stop, os.Interrupt)
```

`os.Interrupt` is SIGINT. Kubelet sends SIGTERM on pod termination, and SIGTERM is not registered, so the process dies by default signal disposition and the entire shutdown block below `<-stop` never executes:

- `a.otelShutdown(shutdownCtx)` : pending spans and metrics are dropped rather than flushed
- `a.brokerServer.Shutdown(shutdownCtx)` : in-flight HTTP requests and SSE streams are cut at the socket instead of being allowed to finish
- `a.metricsServer.Shutdown(shutdownCtx)`
- `a.mcpBroker.Shutdown(shutdownCtx)` : upstream managers are never stopped
- `a.grpcServer.GracefulStop()` : active ext_proc streams are severed mid-request
- `a.redisClient.Close()`

The control-plane binary is not affected: `cmd/main.go:82` uses `ctrl.SetupSignalHandler()`, which handles both signals. The repo's own test servers already use the correct pattern (`tests/servers/server2/main.go:34`, `tests/servers/user-specific-server/main.go:31`); the data-plane binary is the outlier.

## Impact

Every pod termination is effectively a hard kill: rollouts, scale-down, node drain, and eviction. Clients with an in-flight `tools/call` see a transport-level reset rather than a clean close, and telemetry for the terminating pod is lost, which makes shutdown-adjacent incidents hard to reconstruct after the fact.

The shutdown code has been present and exercised in local runs (where Ctrl-C does deliver SIGINT), which is likely why this has gone unnoticed... it works everywhere except the environment it was written for.

## Proposed Fix

Register SIGTERM alongside `os.Interrupt`:

```go
signal.Notify(stop, os.Interrupt, syscall.SIGTERM)
```

One line plus the `syscall` import, matching the pattern already used by the test servers.

Worth being explicit about the boundary: this only makes the existing shutdown path reachable. It does not add a drain window, so work still in flight when the handler fires is bounded by the 10s `shutdownCtx` and by whatever the pod's grace period happens to be, and backend sessions whose cleanup timer lived in the dying process are still not reclaimed. Those are real gaps but they are a design question rather than a defect, and I would rather not conflate them with a one-line fix.., happy to open something separate on graceful draining and session ownership if there is appetite for it.

## Review Guidance

- `cmd/mcp-broker-router/main.go:276` : the registration
- `cmd/mcp-broker-router/main.go:314-350` : the shutdown block that is currently unreachable in-cluster
- `cmd/main.go:82` : control plane, already correct via `ctrl.SetupSignalHandler()`

## Verification

Built the binary before and after the change and sent a real SIGTERM to each, running as PID 1 in a container to match how the image actually executes (`Dockerfile:28` is exec-form `CMD` with no init wrapper):

| binary | runs as | `docker stop` result | cleanup ran |
| --- | --- | --- | --- |
| unpatched | PID 1 | exit 2, no shutdown log | no |
| unpatched | not PID 1 (via `--init`) | exit 143, no shutdown log | no |
| patched | PID 1 | exit 0, `msg="shutting down MCP Broker and MCP Router"` | yes |

Worth recording the PID 1 detail, because it is the reproduction most people will hit and the exit code is not the obvious one. The Go runtime does install a SIGTERM handler, so the signal is delivered even at PID 1 and the container does not hang until the grace period expires. With no `signal.Notify` registration the runtime calls `dieFromSignal`, which restores `SIG_DFL` and re-raises; at PID 1 the kernel discards that re-raise, so the runtime falls through to `exit(2)`. Outside PID 1 the re-raise lands normally and you get the expected 143. Both paths skip the handler entirely, so the observable defect is identical either way.

Container stop was effectively instant in every case, which rules out the alternative theory that the signal is simply ignored until SIGKILL.

`make lint` and `make test-unit` pass. No unit test is included: `cmd/mcp-broker-router` has no test files today and `run()` binds listeners and blocks on the signal channel, so covering this in isolation would mean refactoring the signal set for injection; disproportionate for a one-line change. A rollout-restart e2e would cover it naturally if draining is picked up later.

Aman-Cool:
Fixes: #1351 
Kubelet sends SIGTERM on pod termination, but run() registered only os.Interrupt, so nothing below <-stop in cmd/mcp-broker-router/main.go ever ran in a cluster ; this registers SIGTERM alongside it, one line plus the import.

What was being skipped on every pod termination: pending telemetry dropped rather than flushed, in-flight HTTP requests and SSE streams cut at the socket, upstream managers never stopped, active ext_proc streams severed mid-request, Redis client never closed. Rollouts, scale-down, node drain and eviction were all effectively hard kills. It went unnoticed because Ctrl-C locally does deliver SIGINT, so the path works everywhere except the environment it was written for. The control-plane binary was never affected, cmd/main.go uses ctrl.SetupSignalHandler() which covers both, and the repo's own test servers already used the correct pattern, so the data-plane binary was the outlier.

Verified by building the binary before and after and sending a real SIGTERM to each, running as PID 1 in a container to match how the image actually executes. Unpatched at PID 1 exits 2 with no shutdown log, unpatched outside PID 1 exits 143 with no shutdown log, patched at PID 1 exits 0 and logs "shutting down MCP Broker and MCP Router". The PID 1 detail is worth flagging for anyone reproducing this, since the exit code is not the obvious one: the Go runtime does install a SIGTERM handler, so the signal is delivered and the container does not stall until the grace period, but with no signal.Notify registration the runtime calls dieFromSignal, restores SIG_DFL and re-raises, and at PID 1 the kernel discards that re-raise so the runtime falls through to exit(2). Both paths skip the handler, so the defect is identical either way. make lint and make test-unit pass.

One thing I want to flag rather than quietly ship, because it is a real behavioural change and not obviously in scope for a one-liner. Making the path reachable means grpcServer.GracefulStop() now actually runs, and it takes no context, so it blocks until pending RPCs finish ; the 10s shutdownCtx above it does not apply to it. ext_proc streams are pending RPCs, so on a busy pod termination goes from instant to bounded only by kubelet's SIGKILL at the grace period. That is more correct than what we had, and it is the point of graceful shutdown, but it is slower, and I have confirmed the mechanism from the gRPC semantics rather than measured the magnitude under real ext_proc load. Happy to add a bounded fallback here, Stop() after a deadline, either in this PR or a follow-up, whichever you prefer.

No unit test: cmd/mcp-broker-router has no test files today and run() binds listeners and blocks on the signal channel, so covering this in isolation means refactoring the signal set for injection, which felt disproportionate against a one-line change. A rollout-restart e2e would cover it naturally, and that is where I would put it if the drain gap gets picked up separately, since this only makes the existing shutdown path reachable, it does not add a drain window or reclaim backend sessions whose cleanup timer died with the process.

### DIFF
--- a/cmd/mcp-broker-router/main.go
+++ b/cmd/mcp-broker-router/main.go
@@ -12,6 +12,7 @@ import (
 	"os"
 	"os/signal"
 	"sync"
+	"syscall"
 	"time"
 
 	"github.com/Kuadrant/mcp-gateway/internal/broker"
@@ -288,7 +289,10 @@ func (a *app) loadAndWatchConfig(ctx context.Context) {
 
 func (a *app) run(ctx context.Context) {
 	stop := make(chan os.Signal, 1)
-	signal.Notify(stop, os.Interrupt)
+	// SIGTERM is what kubelet sends on pod termination; without it the graceful
+	// shutdown below never runs in a cluster and the process dies by default
+	// signal disposition. os.Interrupt covers local runs.
+	signal.Notify(stop, os.Interrupt, syscall.SIGTERM)
 
 	lc := net.ListenConfig{}
 	lis, err := lc.Listen(ctx, "tcp", a.routerCfg.addr)
--- a/cmd/mcp-broker-router/main.go
+++ b/cmd/mcp-broker-router/main.go
@@ -289,9 +289,7 @@ func (a *app) loadAndWatchConfig(ctx context.Context) {
 
 func (a *app) run(ctx context.Context) {
 	stop := make(chan os.Signal, 1)
-	// SIGTERM is what kubelet sends on pod termination; without it the graceful
-	// shutdown below never runs in a cluster and the process dies by default
-	// signal disposition. os.Interrupt covers local runs.
+	// handle local interrupts and SIGTERM from kubernetes
 	signal.Notify(stop, os.Interrupt, syscall.SIGTERM)
 
 	lc := net.ListenConfig{}


david-martin (approved):
Thanks @Aman-Cool

PR #1379, opened 2026-08-15
remaining resources federation tasks 

### LINKED ISSUE #1378: Add resource authorization filtering, response rewriting, and prefix validation
**Part of**: #1356 

## Summary

Complete the MCP resources federation feature with JWT-based authorization filtering, response URI rewriting, and controller-level prefix validation. Resources federation allows MCP servers to expose resources (files, documents, etc.) through the gateway with proper authorization, prefix isolation, and routing.

### Resource Authorization Filtering
Implement JWT-based authorization filtering for resources via `x-mcp-authorized` header:
- Parse JWT claim structure: `{"resources": {"server-name": ["authority1", "authority2"]}}`
- Filter resources array in broker before returning to clients
- Strip server prefix from authority for JWT comparison (e.g., `app_example.com` → `example.com`)
- Support enforcement mode: require JWT claim to list any resources
- Deny resources with unauthorized authorities

### Response URI Rewriting
Rewrite resource URIs in tool/call responses:
- Strip server prefix from `_meta.ui.resourceUri` field in tool response content
- Ensure URLs sent back to clients match original authority (without gateway prefix)
- Clients never see the internal prefix format

### Controller Prefix Validation
Add cluster-wide prefix uniqueness validation in MCPServerRegistration controller:
- Reject exact prefix duplicates across active registrations
- Allow substring collisions (e.g., `app_` and `app_admin_`) — resolved via longest-prefix-match in broker
- Validate early in reconcile, before HTTPRoute lookup
- Set appropriate status conditions for clear user feedback

### Unit and E2E Tests
- Unit tests for JWT filtering, prefix stripping, authority matching
- Unit tests for controller validation logic
- E2E tests verifying resources/read routing through Envoy with prefix rewriting
- Edge cases: empty claims, missing servers, malformed URIs, single-char prefixes, overlapping prefixes

kkukadia:
## Summary

Complete the MCP resources federation feature with JWT-based authorization filtering, response URI rewriting, and controller-level prefix validation. Resources federation enables MCP servers to expose resources through the gateway with proper authorization, prefix isolation, and routing.

### Resource Authorization Filtering
JWT-based filtering for resources via x-mcp-authorized header. Clients send a JWT claim mapping server names to allowed resource
authorities. The broker filters the resources arrants, stripping the server prefix from authoritiesto match the JWT claim. Supports enforcement mode where resources require a valid JWT claim.

- Parse JWT claim structure: {"resources": {"server-name": ["authority1"]}}
- Filter resources array before returning to clien
- Strip server prefix from authority for JWT comparison (example: app_example.com becomes example.com)
- Require JWT claim to list resources in enforceme
- Deny resources with unauthorized authorities

### Response URI Rewriting
Strip server prefix from resource URIs in tool/calcall a tool that returns resources, responsecontains _meta.ui.resourceUri fields with original authorities. Prevents information leakage of internal prefix scheme to
clients.

- Strip server prefix from _meta.ui.resourceUri fi
- Ensure URLs sent back to clients match original authority
- Clients never see internal prefix format

### Controller Prefix Validation
Cluster-wide prefix uniqueness validation in MCPServerRegistration controller. Rejects exact prefix duplicates across servers. Allows substring collisions (app_ and app_admin_) x-match in broker. Validates early in reconciliation before HTTPRoute lookup.

- Reject exact prefix duplicates across active registrations
- Allow substring collisions resolved via longest-
- Validate early in reconcile, before HTTPRoute lookup
- Set appropriate status conditions for user feedback

Closes #1378

@Patryk-Stefanski @david-martin


tests/e2e/resources_read_routing_test.go
```diff
@@ -0,0 +1,169 @@
...
+
+		Expect(json.Unmarshal([]byte(respBody), &listResp)).To(Succeed())
+		Expect(len(listResp.Result.Resources)).To(BeNumerically("<=", resourceCount), "JWT filter should reduce resources")
+
+		By("Verifying all remaining resources are from example.com authority")
+		for _, resource := range listResp.Result.Resources {
+			Expect(resource.URI).To(ContainSubstring("example.com"), "only example.com resources should remain")
+		}
+	})
+})
```
Patryk-Stefanski, line 172: **Blocking — wrong JSON shape assertion (this is likely why the test is failing)**

The router's `routeResourceRead` returns `BuildSSEToolError(...)`, which wraps the error as a JSON-RPC *result* with `isError: true`:

```json
{"result":{"content":[{"type":"text","text":"MCP error -32602: Resource not found"}],"isError":true}}
```

There is no `error` key in this response, so `errResp.Error.Code` will always be `0`, and `Expect(0).To(Equal(-32603))` fails.

Also, the comment says `-32603` (internal error) but the router sends the string `"MCP error -32602: Resource not found"` — the code and comment disagree.

Fix the test to decode the actual shape:
```go
var toolErrResp struct {
    Result struct {
        Content []struct {
            Text string `json:"text"`
        } `json:"content"`
        IsError bool `json:"isError"`
    } `json:"result"`
}
Expect(json.Unmarshal([]byte(respBody), &toolErrResp)).To(Succeed())
Expect(toolErrResp.Result.IsError).To(BeTrue())
Expect(toolErrResp.Result.Content[0].Text).To(ContainSubstring("Resource not found"))
```

Patryk-Stefanski (changes requested):
Thanks for pushing through the hard parts of resources federation, Krishna! The routing logic and unit tests are solid. Found a few issues — two blocking ones that explain the test failure, and some nits. See inline comments below.

internal/mcp-router/resource_rewrite.go
```diff
@@ -0,0 +1,27 @@
...
+}
+
+// Process processes response body and rewrites resource URIs by stripping the prefix.
+func (r *resourceURIRewriter) Process(_ context.Context, body []byte) []byte {
+	if r.prefix == "" {
+		return body
+	}
+	// URI rewriting happens at the broker level via stripResourcePrefix.
+	// The ext_proc adapter just passes through for now.
+	return body
```
Patryk-Stefanski: **Blocking — this is a wired but no-op stub**

`Process` unconditionally returns `body` unchanged, and the comment explains why. But this struct is constructed and wired into the hot response path in `ext_proc_adapter.go` on every tool-call response when `mcpRequest.ServerPrefix != ` — allocating a struct and calling into it for zero effect.

The PR description says "Strip server prefix from `_meta.ui.resourceUri` fields", but that code doesn't exist here.

Please either:
- Implement the actual rewrite (scan the JSON body for `_meta.ui.resourceUri` fields and strip the prefix), or
- Remove `resourceURIRewriter`, its construction in `ext_proc_adapter.go` (lines ~740–761), and the `defer resourceRewriter.Flush` — and file a follow-up issue for the response-side rewrite

Leaving a no-op rewriter wired in is misleading and wastes allocations on every streamed tool response.

internal/broker/filtered_resources_handler.go
```diff
@@ -0,0 +1,129 @@
...
+}
+
+// resourceAuthorityFromURI extracts the authority (host) from a resource URI.
+// For malformed URIs or URIs with no host, returns empty string.
+func resourceAuthorityFromURI(uri string) string {
+	u, err := url.Parse(uri)
+	if err != nil {
+		return uri
+	}
+	return u.Host
```
Patryk-Stefanski: **Nit — duplicate of `routing.ResourceAuthority`**

This function does exactly the same thing as `routing.ResourceAuthority` in `internal/routing/mcp_request.go` (both call `url.Parse(uri).Host`, same error fallback). The broker package already imports `internal/routing` (used for `routing.EnsureSeparator` on line 114), so you can just call `routing.ResourceAuthority(resource.URI)` directly and delete this function. Reduces maintenance surface.

internal/broker/filtered_resources_handler.go
```diff
@@ -0,0 +1,129 @@
...
+	"github.com/Kuadrant/mcp-gateway/internal/routing"
+	"github.com/modelcontextprotocol/go-sdk/mcp"
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/trace"
+)
+
+// FilterResources reduces the resource set based on authorization headers.
+func (broker *mcpBrokerImpl) FilterResources(ctx context.Context, headers http.Header, mcpRes *mcp.ListResourcesResult) {
+	attrs := []attribute.KeyValue{brokerComponentAttr}
+	ctx, span := brokerTracer().Start(ctx, "mcp-broker.resources-list", trace.WithAttributes(attrs...))
```
Patryk-Stefanski: **Nit — span name reuses the listing span name**

`"mcp-broker.resources-list"` is the same name used by the resources/list path. In traces these two spans will look identical, making it hard to distinguish listing from filtering. Use `"mcp-broker.resources-filter"` here.

internal/broker/routing_table.go
```diff
@@ -33,6 +33,13 @@ func (m *mcpBrokerImpl) buildRoutingTable() *routing.Table {
 			b.AddPrefix(cfg.Prefix, route)
 		}
 
+		// resources are never pre-registered (see FetchResources), so
+		// resources/read routing resolves purely by prefix, same skip
+		// conditions FetchResources applies when fetching resources/list
+		if up.SupportsResources() && cfg.Prefix != "" && resourcePrefixAllowlist.MatchString(cfg.Prefix) {
+			b.AddResourcePrefix(cfg.Prefix, route)
```
Patryk-Stefanski: **Nit — store the separator-padded prefix as the key, not the raw prefix**

`rewriteResourceURI` in the broker stores URIs with `EnsureSeparator(prefix)` applied to the authority segment (e.g. prefix `"docs"` → authority `"docs_example.com"`). `LookupResourcePrefix` then does `strings.HasPrefix(authority, prefix)`, which will match `"docs_example.com"` against the raw key `"docs"` — but it will also spuriously match `"docsxyz_example.com"`.

Storing `EnsureSeparator(cfg.Prefix)` as the key makes the lookup exact and consistent with how URIs are constructed:
```go
b.AddResourcePrefix(routing.EnsureSeparator(cfg.Prefix), route)
```

tests/e2e/resources_read_routing_test.go
```diff
@@ -0,0 +1,169 @@
...
+		var resp struct {
+			Result struct {
+				Contents []struct {
+					URI  string `json:"uri"`
+					Text string `json:"text"`
+				} `json:"contents"`
+			} `json:"result"`
+		}
+		if err := json.Unmarshal([]byte(respBody), &resp); err != nil {
+			fmt.Printf("RESPONSE BODY: %s\n", respBody)
```
Patryk-Stefanski: **Nit — debug `fmt.Printf` left in test**

This will print to stdout during any test run where this unmarshal fails. Use `GinkgoWriter.Printf` so output is captured by the test framework and only surfaced in verbose mode, or remove it (Ginkgo already prints the failing assertion context).

PR #1391, opened 2026-08-18
feat(resources): implement resources federation with JWT authorization and prefix validation

kkukadia:
## Summary

Adds `_meta.ui.resourceUri` rewriting for `tools/call` responses, renames `sseRewriter` → `elicitationRewriter` to make room for it, and adds e2e coverage for JWT-based resource authorization filtering (implemented earlier in #1356).

Part of #1378

Prefix uniqueness enforcement (originally planned as part of this PR) has been split out to its own PR: #1398.

### Response URI Rewriting

Rewrites `_meta.ui.resourceUri` in `tools/call` responses so the URI a client gets back from a tool call matches the prefixed form `resources/list` already returns for the same resource. Only `ui://` values are rewritten; everything else passes through untouched.

`resourceURIRewriter` rewrites incrementally per complete line (not buffered until end-of-stream) because it runs chained after `elicitationRewriter` on the same response — `elicitationRewriter` must forward `elicitation/create` immediately so the client can answer it while the backend call is still open, and a downstream rewriter holding every byte until end-of-stream would swallow that message and deadlock the exchange when both are active on the same response.

- Bounded line buffering (1 MiB, matching the existing `maxSSELineBytes` convention) — an oversized unterminated line is forwarded unrewritten rather than held indefinitely
- Streaming is only enabled for tool calls to servers with a prefix, not every tool call

### Rename `sseRewriter` → `elicitationRewriter`

No behavior change. Renamed for clarity ahead of `resourceURIRewriter` (above) to avoid a naming collision.

### JWT Resource Authorization Filtering (e2e coverage only)

The broker-side filtering itself already shipped in #1356; this PR adds e2e tests for it (inclusion/exclusion filtering by JWT claim, empty-claim-denies-all) alongside the router-side work above.

## Testing

- Unit tests covering response URI rewriting (including a regression guard for the composition deadlock described above)
- E2E tests covering JWT resource claim filtering (inclusion and exclusion), elicitation rewriting post-rename, and resource URI rewriting (prefixed form, multi-server isolation, non-`ui://` passthrough, ordinary-tool-call regression guard)

Verified against a live kind cluster: rebuilt and loaded the gateway and server1 images, ran all Resources Federation e2e specs against the real broker/router/controller.

### DIFF
--- a/config/istio/gateway/gateway.yaml
+++ b/config/istio/gateway/gateway.yaml
@@ -89,7 +89,7 @@ spec:
           from: All
     - name: resources-federation
       hostname: '*.resources-federation.127-0-0-1.sslip.io'
-      port: 8080
+      port: 8082
       protocol: HTTP
       allowedRoutes:
         namespaces:
--- a/config/istio/gateway/gateway.yaml
+++ b/config/istio/gateway/gateway.yaml
@@ -87,3 +87,10 @@ spec:
       allowedRoutes:
         namespaces:
           from: All
+    - name: resources-federation
+      hostname: '*.resources-federation.127-0-0-1.sslip.io'
+      port: 8080
+      protocol: HTTP
+      allowedRoutes:
+        namespaces:
+          from: All
--- a/config/istio/gateway/nodeport.yaml
+++ b/config/istio/gateway/nodeport.yaml
@@ -39,6 +39,12 @@ spec:
     port: 8081
     protocol: TCP
     targetPort: 8081
+  - appProtocol: http
+    name: resources-federation
+    nodePort: 30084
+    port: 8082
+    protocol: TCP
+    targetPort: 8082
   selector:
     gateway.networking.k8s.io/gateway-name: mcp-gateway
   sessionAffinity: None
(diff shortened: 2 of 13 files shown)


kkukadia:
Hi @Patryk-Stefanski closed #1379 in favor of this PR. #1379 bundled several features together, which made e2e tests difficult for me to manage. This PR currently covers JWT resource authorization filtering) and sseRewriter to elicitationRewriter rename tasks. This is still a work in progress, no action needed yet, just wanted to let you why #1379 was closed.

david-martin (approved):
LGTM. URI rewriting is correct and the test coverage is thorough (SSE + plain JSON, chunk-split reassembly, oversized-line overflow, and the elicitation-composition guard).

Non-blocking follow-up: `injectResourcePrefix` duplicates the broker's `rewriteResourceURI`, and the inverse `stripResourcePrefix` is duplicated too (broker + router). The list/call consistency invariant currently holds only because the copies stay in sync by hand. Worth consolidating both directions into `internal/routing` (next to `EnsureSeparator`) as a separate cleanup, not here.

One doc nit: the PR body leads with JWT authorization filtering, which is broker-side and not in this diff. Trim it so reviewers know what they're signing off on.

kkukadia:
I am on it 🫡

PR #1398, opened 2026-08-18
feat(resources-federation): enforce prefix uniqueness per MCPGatewayExtension

kkukadia:
## Summary

Rejects an `MCPServerRegistration` when its prefix exactly duplicates another active registration's prefix within any `MCPGatewayExtension` namespace it would feed. Closes the gap called out in the resources-federation design doc's "Conflict Detection" section: exact-duplicate prefixes are not currently rejected at registration time.

Part of #1378

## Design: scoped per-extension, not per-Gateway

Two registrations only collide if their configs reach the same broker/router instance. A single Gateway can host multiple independent `MCPGatewayExtension`s, each with its own broker deployment and its own federated server list — verified directly against a running cluster (`tool-discovery-ext`, `protocol-2026-ext`, and `resources-federation-ext` all target listeners on the same shared `mcp-gateway` Gateway object, each with a separate broker pod in a separate namespace). Two registrations with the same prefix feeding *different* extensions never share a routing table and can't actually conflict — scoping the check per-Gateway would reject that as a false positive.

## Changes

- `checkPrefixConflict`, `resolveValidNamespaces`, `findOldestMCPServerRegistration` in `mcpserverregistration_controller.go`
- New `PrefixIndex` field indexer, new `PrefixConflict` status condition reason
- Oldest registration (by creation time) wins, mirroring the existing `checkNamespaceConflict`/`checkListenerConflict` pattern in `mcpgatewayextension_controller.go`
- Symmetric UID tie-break for same-second creation: `CreationTimestamp` only has second granularity, so two registrations created in the same second compare equal there. An index-based tie-break (as used by the existing `findOldestExtension`) lets each registration see *itself* as oldest when reconciling, so both could reach Ready with a real collision. This is fixed for the new code here; the same latent behavior in `findOldestExtension` is unfixed and out of scope for this PR.
- Substring-prefix collisions (`app_` vs `app_admin_`) are explicitly not rejected — resolved deterministically by longest-prefix-match elsewhere, per the design doc's accepted trade-off

### Required test updates elsewhere

Rejecting same-prefix registrations is a real behavior change, and it broke two pre-existing assumptions in the e2e suite:

- `happy_path_test.go`: a test named "should become Ready even when tool conflicts exist from same prefix" explicitly asserted the old contract (both registrations Ready, conflicts resolved broker-side). Updated to assert the new one: the older registration stays Ready, the newer is rejected with `PrefixConflict`.
- `user_specific_list_test.go`: reuses the same registration prefixes (`std_`, `uspec_`) across 7 separate specs feeding the shared default extension — a pre-existing violation of this suite's own documented "every `MCPServerRegistration` must use a unique prefix" rule, harmless before this change but now a latent cross-process scheduling race under `--procs>1`. Marked the affected specs `Serial` rather than renaming prefixes, which would have required updating 27 hardcoded literal tool-name assertions for no behavioral benefit.

## Testing

- Unit tests: `findOldestMCPServerRegistration`, including a dedicated test proving the UID tie-break is symmetric regardless of which registration is reconciling
- Integration tests (envtest): same-extension conflict rejected, different-extension conflict allowed (the scoping this design hinges on), substring collisions not rejected, conflict clears once the sibling is deleted, older registration stays Ready across re-reconcile
- No dedicated e2e test: this is pure controller/CRD-level validation with no Envoy or broker runtime involved, and envtest already runs a real API server exercising the same code path
- Verified against a live kind cluster (manual smoke test): confirmed the same-second tie-break behavior and the fix, and re-ran the affected e2e specs with `--procs=2` after the test updates above.

### DIFF
--- a/internal/controller/mcpserverregistration_controller.go
+++ b/internal/controller/mcpserverregistration_controller.go
@@ -44,6 +44,8 @@ const (
 	HTTPRouteIndex = "spec.targetRef.httproute"
 	// ProgrammedHTTPRouteIndex used to find programmed httproutes
 	ProgrammedHTTPRouteIndex = "status.hasProgrammedCondition"
+	// PrefixIndex used to find MCPServerRegistrations sharing a prefix, for conflict detection
+	PrefixIndex = "spec.prefix"
 
 	// conditionReasonReady is the reason used when the MCPServerRegistration is ready
 	conditionReasonReady = "Ready"
@@ -51,6 +53,9 @@ const (
 	conditionReasonNotReady = "NotReady"
 	// conditionReasonDisabled is the reason used when the MCPServerRegistration is disabled
 	conditionReasonDisabled = "Disabled"
+	// conditionReasonPrefixConflict is the reason used when another active MCPServerRegistration
+	// feeding the same MCPGatewayExtension already uses this prefix
+	conditionReasonPrefixConflict = "PrefixConflict"
 )
 
 // ServerInfo holds server information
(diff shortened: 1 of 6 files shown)


kkukadia:
hi @Patryk-Stefanski / @david-martin, wanted to share this before you review, the duplicate prefix rejection has a real e2e test. The trickier case (IMO) is two servers reusing the same prefix on different gateway listeners, which is fine since they're isolated from each other is only covered by unit-style tests, not a live cluster. Lemme know if you would like me to add a live e2e test for that before merging.

david-martin:
> Lemme know if you would like me to add a live e2e test for that before merging.

Let's leave it to unit for coverage. e2e would be overkill

david-martin (approved):
Thanks @kkukadia 

Merging

PR #1400, opened 2026-08-18
fix for 2025 stateless server

### LINKED ISSUE #1382: Broker hard-loops against a stateless streamable-HTTP upstream (no Mcp-Session-Id; GET → 405)
### Summary

With mcp-gateway **v0.9.0**, the broker enters a tight reconnect loop against an upstream MCP server that implements **stateless** streamable-HTTP — one that returns `200` to `initialize` **without** an `Mcp-Session-Id` response header and does **not** support the `GET` server→client SSE stream (`GET` → `405 Method Not Allowed`).

### Observed behavior

1. `POST initialize` → `200 text/event-stream` (no `Mcp-Session-Id`)
2. broker attempts to open the persistent listen channel (`subscriptions/listen`) with an **empty** session id
3. fails with `session not found` / `failed to list tools`, tears down, reconnects — many times per second

Broker logs (repeating):

```
level=ERROR msg="connection lost, clearing session for reconnect" ... error="sending \"subscriptions/listen\": failed to connect (session ID: ): session not found"
level=ERROR msg="failed to list tools" ... error="... calling \"tools/list\": client is closing: ... session not found"
level=INFO  msg="reconnecting after connection loss"
```

Net effect: the upstream contributes **zero** federated tools, and the loop wastes CPU on both the broker and the upstream (each cycle re-hits `initialize`).

### Reproduce

Register an upstream whose MCP endpoint is stateless streamable-HTTP:
- `initialize` → `200`, no `Mcp-Session-Id`
- a bare `tools/list` POST (no session) → `200` (tools are reachable statelessly)
- `GET <endpoint>` → `405`

### Expected

Tolerate stateless upstreams: if `initialize` yields no `Mcp-Session-Id` and/or the `GET` listen stream returns `405`, fall back to stateless request/response (which works) instead of hard-looping. At minimum, back off rather than reconnect many times per second.

### Environment

- mcp-gateway `v0.9.0`
- upstream: stateless streamable-HTTP MCP endpoint

maleck13:
Handles an empty session id from a 2025 server that responds to the initialize request but doesn't support a session. Stops the OnConnectionLost handler from firing in these cases. 

If the server is down, next time the timer triggers the manager will attempt to connect and fail as normal and backoff and retry

Fixes #1382 

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved connection handling for stateless upstream connections that do not provide a session ID.
  - Prevented unnecessary connection-loss notifications when no active session exists.
  - Ensured connection-loss handling occurs only when an established session reports an error.

- **Tests**
  - Added coverage for stateless upstream connections and rejected monitoring requests.

### DIFF
--- a/internal/broker/upstream/mcp.go
+++ b/internal/broker/upstream/mcp.go
@@ -423,15 +423,25 @@ func (up *MCPServer) notify(method string) {
 
 // OnConnectionLost registers a connection lost handler.
 // In the official SDK, connection loss is observed via session.Wait().
+// upstreams that return no Mcp-Session-Id from initialize do not need
+// connection-loss monitoring: the SDK's subscriptions/listen stream
+// will fail immediately with an empty session ID, which is expected
+// and not a real connection loss. the ticker health check and ping
+// handle failure detection for these upstreams.
 func (up *MCPServer) OnConnectionLost(handler func(err error)) {
 	session := up.currentSession()
-	if session != nil {
-		go func() {
-			if err := session.Wait(); err != nil {
-				handler(err)
-			}
-		}()
+	if session == nil {
+		return
 	}
+	if session.ID() == "" {
+		up.logger.Debug("skipping connection-lost watcher, no session ID from upstream", "upstream", up.ID())
+		return
+	}
+	go func() {
+		if err := session.Wait(); err != nil {
+			handler(err)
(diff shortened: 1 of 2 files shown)


david-martin (approved):
lgtm

PR #1402, opened 2026-08-18
refactor(resources): consolidate resource prefix inject/strip logic

kkukadia:
## Summary

Follow-up to https://github.com/Kuadrant/mcp-gateway/pull/1391#pullrequestreview-4963470610, where @david-martin  flagged `injectResourcePrefix`/`rewriteResourceURI` and the two `stripResourcePrefix` implementations were duplicated between the router and broker.

Consolidates all four into `internal/routing`, next to `EnsureSeparator`, which they already both depend on:

- `InjectResourcePrefix` — absorbs the router's `injectResourcePrefix` and the broker's `rewriteResourceURI` (identical bodies)
- `StripResourcePrefix` — promoted from the router's already-`internal/routing`-local `stripResourcePrefix`, now exported
- `StripAuthorityPrefix` — absorbs the broker's authority-string `stripResourcePrefix`

## Test plan

- `go test -race ./...` passes
-  `make lint-go` passes
- Added direct unit test coverage for `StripResourcePrefix`/`StripAuthorityPrefix` (previously only covered indirectly), plus an inject/strip round-trip test

### DIFF
--- a/internal/broker/broker.go
+++ b/internal/broker/broker.go
@@ -6,7 +6,6 @@ import (
 	"fmt"
 	"log/slog"
 	"net/http"
-	"net/url"
 	"regexp"
 	"slices"
 	"strings"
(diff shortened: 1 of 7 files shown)


david-martin (approved):
Thanks @kkukadia

PR #1408, opened 2026-08-19
test(e2e): derive resources-federation server hostnames from E2E_DOMAIN

david-martin:
The resources federation e2e suite hardcoded `*.resources-federation.127-0-0-1.sslip.io` server hostnames. On non-Kind clusters the `resources-federation` listener hostname is `*.resources-federation.<cluster-domain>`, so the per-registration HTTPRoutes never attached to the listener and the suite couldn't pass there.

Every other isolated-listener suite derives server hostnames from `E2E_DOMAIN` (`toolDiscServerHostDefault`, `protocol2026ServerHost`). This adds `resourcesFederationServerHost(subdomain)` in the same shape and uses it for all 8 registrations.

Surfaced by the OpenShift e2e pipeline nightly, which failed in `BeforeAll` on the resources federation suite. That run also needs a matching `resources-federation` listener added in the pipeline (separate change in testsuite-pipelines); this PR is the mcp-gateway side.

No behaviour change on Kind: `resourcesFederationServerHost` resolves to the same `*.resources-federation.127-0-0-1.sslip.io` under the default domain.

### DIFF
--- a/tests/e2e/commons.go
+++ b/tests/e2e/commons.go
@@ -164,6 +164,11 @@ func resourcesFederationPublicHostDefault() string {
 	return "mcp.resources-federation." + e2eDomain
 }
 
+// resourcesFederationServerHost returns a server hostname for the resources-federation listener.
+func resourcesFederationServerHost(subdomain string) string {
+	return subdomain + ".resources-federation." + e2eDomain
+}
+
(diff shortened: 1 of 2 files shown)


trepel (approved):
LGTM, thanks!

PR #1370, opened 2026-08-13
feat(guardrails) PART 1: validate guardrails Secret and add config types

christinaexyou:
## What does this PR do?

* Adds the guardrails/external/nemo Secret schema (url, configIDs, model, failMode) and EnsureNeMoConfigData to validate and parse it.
* Wires validation into the MCPGatewayExtension reconciler via the mcp.kuadrant.io/guardrails-ref annotation.
* Adds the config propagation scaffolding (GuardrailsConfig, MCPServer.GuardrailsConfigIDs, BrokerConfig.GlobalGuardrailsConfigIDs) plus the maxBodyBytes spec field and new status condition reasons.

Fixes #1302 

Issue #1372

## Pre-review checklist

Before requesting review from a maintainer, confirm you have read [CONTRIBUTING.md](../CONTRIBUTING.md) and:

- [x] Checked the CodeRabbit walkthrough for "Possibly related issues" and confirmed the PR uses `Fixes` or `Closes` syntax for any it addresses
- [x] Checked the CodeRabbit walkthrough for "Possibly related PRs" and confirmed this PR is not a duplicate
- [x] Read, understood, and addressed or dismissed (with a reason) all CodeRabbit comments
- [x] All CI checks pass, or failures have been investigated and explained below
- [x] Ran the `agent-skills:review` skill (from https://github.com/addyosmani/agent-skills) and addressed all valid recommendations

## Summary by CodeRabbit

* **New Features**
  * Added gateway-wide and per-server guardrails configuration.
  * Added validation for guardrails Secrets, including URL, model, and fail-mode settings.
  * Added configurable request and response body buffering limits, defaulting to 1 MiB.
  * Added support for updating or clearing global guardrails configuration.
  * Added status reporting for missing or invalid guardrails configuration.

* **Bug Fixes**
  * Improved detection of guardrails configuration changes, including additions, removals, and reordered configuration IDs.

PR #1409, opened 2026-08-19
feat(guardrails): add Checker and NeMo translation layer

### LINKED ISSUE #1401: Guardrails: add Checker and NeMo translation layer
Add the guardrails Checker interface and NeMo translation (Part 2 of guardrails feature).

*  The Checker interface owns HTTP transport, timeout, TLS, fail mode, and config ID merging, delegating NeMo-specific translation 
* Add a Decision type to map NeMo Guardrails RailsStatus verdicts to Router decisions

Parent issue: #1302

christinaexyou:
## What does this PR do?

* Adds `guardrails.Checker`, the component that runs `tools/call` requests and responses through an external guardrails server. 
* It owns HTTP transport, fail-mode handling, and global/per-server config ID merging, and delegates provider-specific request/response translation to a new `NeMoTransformer` for the NeMo Guardrails `/v1/guardrail/checks` schema. 
* Returns a `Decision` that distinguishes a real guardrails verdict from a `failMode` fallback, so a future caller can pick the right response (e.g. 403 vs 503).

Fixes #1401 and adds a follow-up to PR #1370  to enqueue MCPGatewayExt for secrets with the guardrails-ref annotation. 

## Pre-review checklist

## Summary by CodeRabbit

* **New Features**
  * Added guardrails checks for tool requests and responses, supporting allow, block, and modify outcomes.
  * Added NeMo Guardrails integration with configurable policies, fail modes, and response-size limits.
  * Secret updates now trigger reconciliation for extensions using related guardrails or security references.

* **Bug Fixes**
  * Prevented invalid zero or negative body-size limits.
  * Restricted guardrails configuration URLs to HTTP and HTTPS.
  * Clarified CA bundle trust behavior for upstream and protocol hairpin connections.

* **Chores**
  * Disabled serving of the deprecated v1alpha1 extension version.
  * Updated deployment images to the latest builds.

### DIFF
--- a/internal/controller/session_store.go
+++ b/internal/controller/session_store.go
@@ -50,7 +50,8 @@ func (r *MCPGatewayExtensionReconciler) validateSessionStore(ctx context.Context
 }
 
 // enqueueMCPGatewayExtForSecret maps a secret change to MCPGatewayExtension reconcile requests.
-// It enqueues extensions that reference the secret via trustedHeadersKey or sessionStore.
+// It enqueues extensions that reference the secret via sessionStore, trustedHeadersKey,
+// caCertBundleRef, the session signing key, or the guardrails-ref annotation.
 func (r *MCPGatewayExtensionReconciler) enqueueMCPGatewayExtForSecret(ctx context.Context, obj client.Object) []reconcile.Request {
 	secret := obj.(*corev1.Secret)
 
@@ -83,6 +84,13 @@ func (r *MCPGatewayExtensionReconciler) enqueueMCPGatewayExtForSecret(ctx contex
 			requests = append(requests, reconcile.Request{
 				NamespacedName: types.NamespacedName{Name: ext.Name, Namespace: ext.Namespace},
 			})
+			continue
+		}
+		if ref := ext.Annotations[labelGuardrailsReference]; ref != "" && ref == secret.Name {
+			requests = append(requests, reconcile.Request{
+				NamespacedName: types.NamespacedName{Name: ext.Name, Namespace: ext.Namespace},
+			})
+			continue
 		}
 	}
 	return requests
--- a/internal/guardrails/checker.go
+++ b/internal/guardrails/checker.go
@@ -1,0 +1,266 @@
+// Package guardrails checks tools/call requests and responses against an
+// external guardrails server. Checker owns HTTP transport, timeout, TLS,
+// fail mode, config ID merging, and provider translation.
+package guardrails
+
+import (
+	"bytes"
+	"context"
+	"crypto/tls"
+	"encoding/json"
+	"fmt"
+	"io"
+	"net"
+	"net/http"
+	"strings"
+	"time"
+
+	"github.com/Kuadrant/mcp-gateway/internal/config"
+	"github.com/Kuadrant/mcp-gateway/internal/guardrails/external/nemo"
+)
+
+// checksPath is the guardrails server endpoint all checks are sent to.
+const checksPath = "/v1/guardrail/checks"
+
+// checkTimeout bounds a single guardrails HTTP round trip, comfortably
+// inside the 10s ext_proc message_timeout.
+const checkTimeout = 3 * time.Second
+
+// dialTimeout bounds DNS/TCP connection setup so an unreachable guardrails
+// server fails fast rather than eating the full checkTimeout on dial alone.
+const dialTimeout = 1 * time.Second
+
+// defaultMaxIdleConnsPerHost is used when the caller doesn't specify a
+// concurrency hint.
+const defaultMaxIdleConnsPerHost = 100
+
+// defaultMaxBodyBytes bounds the guardrails server's check response when the
+// caller doesn't specify a limit, matching the MCPGatewayExtension
+// maxBodyBytes default (1 MiB).
+const defaultMaxBodyBytes = 1 << 20
+
+// Status is the outcome of a guardrails check.
+type Status string
+
+// Status values a Decision can carry.
+const (
+	StatusAllowed  Status = "allowed"
+	StatusBlocked  Status = "blocked"
+	StatusModified Status = "modified"
+)
+
+// Decision is the outcome of a single guardrails check, translated from the
+// NeMo Guardrails server response into a form the router acts on.
+type Decision struct {
+	Status Status
+	// Content is the text to forward: the original content unless Status
+	// is StatusModified, in which case it's the guardrails modified text.
+	Content string
+	// Reason names the triggering rail. Empty when Status is StatusAllowed.
+	Reason string
+	// Err is set when Status was resolved by failMode after a transport
+	// failure or unparseable response.
+	Err error
+}
+
+// Checker runs guardrails checks against tools/call requests and responses.
+type Checker interface {
+	CheckRequest(ctx context.Context, toolName string, arguments json.RawMessage, configIDs []string) (*Decision, error)
+	CheckResponse(ctx context.Context, toolName string, content []byte, configIDs []string) (*Decision, error)
+}
+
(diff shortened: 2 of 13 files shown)


internal/guardrails/external/nemo/transformer.go
```diff
@@ -0,0 +1,145 @@
...
+// resolved from the guardrails Secret.
+func NewTransformer(model string) *Transformer {
+	return &Transformer{Model: model}
+}
+
+// TransformRequest translates a tools/call request into a NeMo
+// /v1/guardrail/checks request body. Maps params.name to messages[0].name,
+// params.arguments (JSON-encoded) to messages[0].content, and role to
+// "user".
+func (t *Transformer) TransformRequest(toolName string, arguments json.RawMessage, configIDs []string) ([]byte, error) {
```
maleck13, line 62: TransformResponse (line 106) and TransformRequest (line 76) each copy the body twice:
  string(content) allocates, then json.Marshal re-encodes that string into the output. 

Consider Message.Content to json.RawMessage. For TransformRequest, arguments is already
  json.RawMessage — assign directly (zero copies). For TransformResponse, JSON-quote the []byte
  once into a RawMessage; the outer marshal splices it without re-encoding (one copy instead
  of two).

```go
  type Message struct {
        Role    string          `json:"role"`
        Name    string          `json:"name,omitempty"`
        Content json.RawMessage `json:"content"`
        Config  string          `json:"config,omitempty"`
  }

  func quoteBytes(b []byte) json.RawMessage {
        quoted, _ := json.Marshal(string(b))
        return json.RawMessage(quoted)
  }
```

Note I haven't tried this myself AI helped me with investigating a solution after I noticed multiple copies of the body happening.

internal/controller/session_store.go
```diff
@@ -84,6 +85,12 @@ func (r *MCPGatewayExtensionReconciler) enqueueMCPGatewayExtForSecret(ctx contex
 				NamespacedName: types.NamespacedName{Name: ext.Name, Namespace: ext.Namespace},
 			})
```
maleck13, line 86: Think we need to add a continue here

internal/guardrails/checker.go
```diff
@@ -0,0 +1,266 @@
...
+	}
+	if maxBodyBytes <= 0 {
+		maxBodyBytes = defaultMaxBodyBytes
+	}
+
+	dialer := &net.Dialer{Timeout: dialTimeout}
+	transport := &http.Transport{
+		DialContext:         dialer.DialContext,
+		TLSClientConfig:     tlsConfig,
+		MaxIdleConnsPerHost: maxIdleConnsPerHost,
```
maleck13: do we want to set an idleconnectiontimeout on this transport? Or perhaps cloning the default and overriding the default settings you want to change? We have done this a few time in the code already
```
 transport := http.DefaultTransport.(*http.Transport).Clone()

```

internal/guardrails/checker.go
```diff
@@ -0,0 +1,266 @@
...
+// only one today.
+type provider interface {
+	TransformRequest(toolName string, arguments json.RawMessage, configIDs []string) ([]byte, error)
+	TransformResponse(toolName string, content []byte, configIDs []string) ([]byte, error)
+	ParseCheckResponse(body []byte) (status Status, content, reason string, err error)
+}
+
+// nemoProvider adapts *nemo.Transformer to the provider interface,
+// translating NeMo's status strings into the transport-agnostic Status.
+type nemoProvider struct {
```
maleck13, line 84: nit: I think if we embed here we can remove some of the empty wrappers below:

```go
type nemoProvider struct{

*nemo.Transformer

}
```

internal/guardrails/checker.go
```diff
@@ -0,0 +1,266 @@
...
+	}
+	return c.check(ctx, body)
+}
+
+// check performs the guardrails HTTP round trip and maps the outcome to a
+// Decision. Non-2xx, transport errors, oversized bodies, and unparseable
+// responses all fall back to failMode rather than propagating an error —
+// only a translation failure (handled by the caller) skips failMode
+// entirely.
+func (c *nemoChecker) check(ctx context.Context, body []byte) (*Decision, error) {
```
maleck13, line 171: nit , suggestion: we might benefit from avoiding any missed errors if we split this?

```go
func (c *nemoChecker) check(ctx context.Context, body []byte) (*Decision, error)
  {
      decision, err := c.doCheck(ctx, body)
      if err != nil {
          return c.failModeDecision(err), nil
      }
      return decision, nil
  }

  func (c *nemoChecker) doCheck(ctx context.Context, body []byte) (*Decision,
  error) {
      ctx, cancel := context.WithTimeout(ctx, checkTimeout)
      defer cancel()

      req, err := http.NewRequestWithContext(ctx, http.MethodPost,
  c.baseURL+checksPath, bytes.NewReader(body))
      if err != nil {
          return nil, fmt.Errorf("guardrails: failed to build check request: %w",
  err)
      }
      req.Header.Set("Content-Type", "application/json")
      req.Header.Set("Accept", "application/json")

      resp, err := c.httpClient.Do(req)
      if err != nil {
          return nil, fmt.Errorf("guardrails: request failed: %w", err)
      }
```

maleck13 (commented):
generally looking good. A couple of items worth attention/consideration:

1. Transport config: clone http.DefaultTransport 
2. Wrap check() with failModeDecision middleware — single call site eliminates chance of missing an error response.
3. Double body copy in transformer — string(content) then json.Marshal re-encodes. Change Message.Content to json.RawMessage 
4. Embed *nemo.Transformer in nemoProvider — removes two passthrough wrapper
     methods
5. Missing continue in enqueueMCPGatewayExtForSecret — signing key match falls
     through to guardrails check, can enqueue duplicates. session_store.go:87
6. Secret URL validation accepts non-HTTP schemes — ftp:// passes validation,

internal/guardrails/external/nemo/transformer.go
christinaexyou: ack - i took your suggestion and added a help func `quoteJSONString` below

christinaexyou: additionally added a new file `internal/guardrails/external/nemo/transformer_test.go` to test the conversion between MCP Gateway traffic and NeMo

internal/guardrails/checker.go
christinaexyou: done; i split it into `Check` and `checkResponse`

christinaexyou:
thanks for your feedback @maleck13 !

internal/guardrails/checker.go
```diff
@@ -0,0 +1,265 @@
...
+
+	respBody, err := io.ReadAll(io.LimitReader(resp.Body, c.maxBodyBytes+1))
+	if err != nil {
+		return nil, fmt.Errorf("guardrails: failed to read response: %w", err)
+	}
+	if int64(len(respBody)) > c.maxBodyBytes {
+		return nil, fmt.Errorf("guardrails: response exceeds %d byte limit", c.maxBodyBytes)
+	}
+
+	if resp.StatusCode < http.StatusOK || resp.StatusCode >= http.StatusMultipleChoices {
```
maleck13: nit: could prob do this status check before reading the body?

maleck13 (dismissed):
happy to approve here. 1 nit that could be addressed but not a blocker around reading the body then check the resp.Status code rather than checking the status first then reading the body.

internal/guardrails/checker.go
christinaexyou: done, the status check now comes before reading the body and also added line 197 to discard the error msg so we're within the maxBodyBytes limit

maleck13 (approved):

PR #1411, opened 2026-08-20
test(e2e): improve wait for deployment rollout in AuthPolicy suite

trepel:
## What does this PR do?

Using more reliable wait for Deployment to get ready. Similar approach is used in the testsuite in several places already.

### Verification Steps
Eye review only, all check passing should suffice for verification.

## Pre-review checklist

## Summary by CodeRabbit

* **Tests**
  * Improved end-to-end test reliability when updating trusted-header configuration by waiting for the updated deployment generation to become available.


david-martin (approved):
Approving, though it's hard to say why this is only presenting as an issue now. Perhaps the gingko version change in https://github.com/Kuadrant/mcp-gateway/pull/1388 ? Either way, let's see if this makes a difference

trepel:
> Approving, though it's hard to say why this is only presenting as an issue now. Perhaps the gingko version change in #1388 ? Either way, let's see if this makes a difference

This is a reasonable change on its own but yes, I am not sure whether this solves the auth test failure we're seeing atm. We'll see.

PR #1413, opened 2026-08-20
fix(router): retry transient hairpin init + close idle backend sessions

david-martin:
Two router perf fixes on the hairpin/session path, both surfaced by the forge scale-out benchmark.

**1. Retry transient hairpin init failures**
Hairpin `initialize` fails under concurrency with transient connection rejections (`connection refused`, transport reset) and returns `500 failed to create session` on the first attempt. Retry non-4xx failures with bounded backoff (100ms, 200ms) inside the existing singleflight. 4xx short-circuits unchanged. Total backoff stays under the ext_proc `message_timeout` (10s). Seen as ~75k 500s in one run (s10-u200), self-recovering.

**2. Close idle backend sessions instead of holding for 24h**
Backend client handles were held open until the 24h JWT expiry, so a client that made one tool call and left kept an idle upstream connection for a full day. Now each handle has a 30m idle timer (reset on activity, mirrors the broker `SessionIdleTimeout`). On idle, close the handle and drop only the per-server mapping; the gateway session stays valid so the next tool call re-inits lazily.

### DIFF
--- a/internal/routing/router_202511.go
+++ b/internal/routing/router_202511.go
@@ -385,6 +385,7 @@ func (r *Router202511) routeToUpstream(ctx context.Context, span trace.Span, mcp
 
 	var remoteMCPServerSession string
 	if id, ok := exists[mcpReq.ServerName]; ok {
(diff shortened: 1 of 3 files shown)


david-martin:
Closing until further evidence of a need for these changes

PR #1427, opened 2026-08-25
Downgrade go version to 1.26.5

jjkiely:
## What does this PR do?

The downstream build environment does not have access to 1.26.6 yet.

Downgrading to 1.26.5 will fix build failures downstream due to go toolchain mismatch.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM golang:1.26.6 AS builder
+FROM golang:1.26.5 AS builder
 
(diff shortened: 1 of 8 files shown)


trepel:
LGTM but I would leave the final approval & merge to @david-martin , not sure if there are any security-related implications to downgrading I am not aware of (original ticket asked for v1.26.5 so it should be good though).

david-martin:
#1437

PR #1283, opened 2026-07-16
docs: umbrella operator integration design docs, test reports, and upgrade guide

Patryk-Stefanski:
## Summary

User-facing documentation for the kuadrant-operator umbrella operator integration (RFC 0019). On OLM-based clusters, MCP Gateway no longer ships a standalone operator — it is installed and managed by the Kuadrant Operator, which embeds the MCP Gateway controller and deploys it on startup. Standalone installs continue to use Helm.

Separated from the implementation PR (#1239) so docs can land and be refined independently.

Closes #1234

## What's included

- `docs/guides/olm-install.md` — rewritten for the Kuadrant Operator install path: subscribe to the operator, then create an `MCPGatewayExtension` to deploy the broker-router data plane. No `Kuadrant` CR is required for the controller to run (it deploys unconditionally on startup). Drops the removed standalone bundle/catalog/`make`-target content (see #1343).
- `docs/guides/olm-upgrade.md` — new how-to for migrating an existing OLM-installed standalone MCP Gateway to the consolidated Kuadrant Operator deployment. Zero-downtime; procedure verified end-to-end on OpenShift 4.22.
- `docs/guides/README.md` — adds both guides to the index.

## Notes

- The Helm-based standalone deployment guide (`docs/guides/how-to-install-and-configure.md`) is already on `main`; #1235 (Helm guide) was closed separately on that basis.
- The design docs / POC reports originally in this branch were removed — the umbrella integration is canonically documented by RFC 0019 in the Kuadrant architecture repo.

## Related

- Implementation PR: #1239
- Standalone OLM packaging removal: #1343
- RFC 0019: https://github.com/Kuadrant/architecture/pull/189
- Epic: #1226

## Summary by CodeRabbit

* **Documentation**
  * Added guidance for installing MCP Gateway through the Kuadrant Operator on Kubernetes and OpenShift.
  * Clarified Helm versus OLM installation options, prerequisites, catalog namespaces, verification, and uninstall steps.
  * Updated migration guidance with validation checks, authenticated-route considerations, rollback instructions, and preservation of existing gateway configuration.
  * Updated configuration examples to the latest CRD version and clarified TLS, hairpinning, and gateway CA certificate setup.

PR #1429, opened 2026-08-25
feat: runtime-managed NetworkPolicy for broker-router deployment

### LINKED ISSUE #1399: Runtime-managed NetworkPolicies for broker-router deployment
## Summary

The mcp-gateway controller should create and manage NetworkPolicy resources for the broker-router deployment to provide ingress-only network isolation. This aligns with RFC 0022 (Kuadrant/architecture#196) which establishes runtime-managed network policies across Kuadrant components.

## Context

The broker-router deployment currently accepts ingress traffic from any source in the cluster. Only the targeted Gateway (from the MCPGatewayExtension) should be able to reach the gRPC and HTTP ports. Metrics should remain open for monitoring.

## Requirements

- **Egress**: allow all (no egress restrictions)
- **Ingress deny-by-default**: a deny policy selecting broker-router pods, blocking all ingress unless explicitly allowed
- **Ingress allow rules per port**:

| Port | Purpose | Allowed Source |
|------|---------|----------------|
| 8080 | Broker HTTP/MCP | Gateway namespace only (from MCPGatewayExtension targetRef) |
| 50051 | Router gRPC ext_proc | Gateway namespace only |
| 9090 | Metrics | Any (open to all monitoring stacks) |

- The controller should derive the gateway namespace from the MCPGatewayExtension's targetRef Gateway
- Policies should update dynamically if the targeted Gateway changes
- Policies should be owned by the MCPGatewayExtension CR for automatic GC on deletion
- Use standard `networking.k8s.io/v1` NetworkPolicy (no AdminNetworkPolicy)

## Port Matrix

```
Client → Gateway (Envoy) → Router (:50051 ext_proc) → Broker (:8080 /mcp) → Upstream MCP Servers
                                                              
                                              Prometheus → :9090 (metrics)
```

## References

- RFC 0022: Kuadrant/architecture#196

david-martin:
Closes #1399

The controller now creates and owns a `networking.k8s.io/v1` NetworkPolicy for the broker-router deployment, giving it ingress-only isolation per RFC 0022. Previously the broker-router accepted ingress from any source in the cluster.

## What it does

Reconciled alongside the other broker-router resources, in the `MCPGatewayExtension` namespace, owned by the CR for automatic GC.

| Port | Purpose | Allowed source |
|------|---------|----------------|
| 8080 | Broker HTTP/MCP | Target Gateway namespace only |
| 50051 | Router gRPC ext_proc | Target Gateway namespace only |
| 9090 | Metrics | Any |

- Ingress is deny-by-default; only the rules above are allowed.
- Egress is unrestricted.
- The Gateway namespace is derived from the `targetRef` and selected via the auto-set `kubernetes.io/metadata.name` label, so it tracks dynamically if the targeted Gateway changes.

## Behavior change

The static Helm template `charts/mcp-gateway/templates/networkpolicy-broker-router.yaml` is removed, since the controller now manages this policy. The controller-pod policy (`networkpolicy-controller.yaml`, gated by `networkPolicy.enabled`) stays. On clusters whose CNI enforces NetworkPolicy, the deny-by-default ingress applies automatically; the target Gateway namespace is auto-allowed, so normal traffic keeps working. See `docs/release-notes/1399-managed-networkpolicy.md`.

## Testing

- Unit tests for `buildBrokerRouterNetworkPolicy` and `networkPolicyNeedsUpdate`
- Integration test asserting owner reference on the created policy
- `[Happy]` e2e spec verifying the policy exists, is owned by the extension, and has the expected ingress/egress

## Summary by CodeRabbit

* **New Features**
  * Broker-router network access is now managed automatically for each targeted Gateway.
  * HTTP and gRPC traffic is limited to the Gateway’s namespace, while metrics remain accessible and outbound traffic is allowed.
  * Policies are automatically created, updated, and removed with associated resources.

* **Documentation**
  * Added guidance describing managed network policies, access rules, and configuration behavior.
  * Clarified that the network policy setting applies only to the controller pod.

* **Bug Fixes**
  * Improved namespace handling when a target namespace is not specified.
  * Removed the previous static policy that permitted unrestricted ingress.


Patryk-Stefanski (commented):
A few concerns on the deny-by-default NetworkPolicy — mostly around silent-failure ingress cases and one doc mismatch. Inline below.

internal/controller/broker_router.go
```diff
@@ -308,6 +310,88 @@ func (r *MCPGatewayExtensionReconciler) buildBrokerRouterService(mcpExt *mcpv1.M
...
+						{
+							NamespaceSelector: &metav1.LabelSelector{
+								MatchLabels: map[string]string{
+									"kubernetes.io/metadata.name": gatewayNamespace,
+								},
+							},
+						},
+					},
+					Ports: []networkingv1.NetworkPolicyPort{
+						{Protocol: &tcp, Port: &httpPort},
```
Patryk-Stefanski, line 358: The readiness probe does an HTTP GET to port 8080 from the kubelet/node IP, but this ingress rule only allows 8080 from the target Gateway's namespace (+ 9090 metrics). On CNIs that enforce NetworkPolicy against host/kubelet-originated probe traffic, the probe gets dropped -> `ReadyReplicas` stays 0 -> `reconcileBrokerRouter` returns `deploymentReady=false` indefinitely -> the MCPGatewayExtension never becomes Ready.

Works out-of-the-box on Calico/Cilium (host traffic is implicitly allowed), but breaks on stricter configs. Since the old policy allowed ingress from anywhere, this is a new self-inflicted failure mode. Worth either adding an explicit ingress allowance for the probe port or documenting the CNI requirement prominently in the release note.

internal/controller/broker_router.go
```diff
@@ -308,6 +310,88 @@ func (r *MCPGatewayExtensionReconciler) buildBrokerRouterService(mcpExt *mcpv1.M
...
+				networkingv1.PolicyTypeEgress,
+			},
+			Ingress: []networkingv1.NetworkPolicyIngressRule{
+				{
+					// gateway namespace only: gRPC (ext_proc) and HTTP (MCP)
+					From: []networkingv1.NetworkPolicyPeer{
+						{
+							NamespaceSelector: &metav1.LabelSelector{
+								MatchLabels: map[string]string{
+									"kubernetes.io/metadata.name": gatewayNamespace,
```
Patryk-Stefanski, line 352: This selects the gateway namespace solely via the `kubernetes.io/metadata.name` label, which is only auto-populated when the `NamespaceDefaultLabelName` feature is enabled (GA in 1.22, but can be disabled or absent on older/customized clusters). If that label is missing on the gateway namespace, the selector matches zero namespaces and all Envoy->broker/router traffic (both 50051 and 8080) is silently blocked with no fallback -> total outage.

Given our minimum supported k8s versions this is low-probability, but the failure is silent and complete. Could be worth a guard/log if the label is absent, or at minimum calling out the requirement in the release note.

docs/reference/mcpgatewayextension.md
```diff
@@ -99,3 +100,19 @@ Trust pool hierarchy: system roots, then gateway CA bundle (if set), then per-se
...
+
+Ingress is deny-by-default except for:
+
+| **Port** | **Purpose** | **Allowed from** |
+|----------|-------------|-------------------|
+| `8080` | Broker HTTP/MCP | Only the target Gateway's namespace |
+| `50051` | Router gRPC (ext_proc) | Only the target Gateway's namespace |
+| `9090` | Metrics | Any source |
+
+The gateway namespace is selected via the auto-set `kubernetes.io/metadata.name` namespace label, derived from `spec.targetRef`. If the targeted Gateway changes, the controller updates the policy to match. Egress allows all traffic.
```
Patryk-Stefanski: "derived from `spec.targetRef`" doesn't mention the empty-namespace fallback: `targetGatewayNamespace` falls back to the MCPGatewayExtension's own namespace when `targetRef.Namespace` is empty. An operator running the gateway and extension in different namespaces who omits `targetRef.Namespace` would get a policy scoped to the extension namespace, which may not be where Envoy actually runs. The `namespace` field row above ("Defaults to the MCPGatewayExtension namespace") documents the defaulting, but this NetworkPolicy sentence reads as if it always follows the target Gateway. Suggest noting the fallback here.

internal/controller/broker_router.go
david-martin: Thinking an update to the install docs to mention a) a NetworkingPolicy is created, and b) the matcher matches on that label.
But would hold off and make that change when #1283 lands with those updated docs first.

docs/reference/mcpgatewayextension.md
david-martin: Reference doc updated.
Will update install doc to mention the NetworkPolicy (see other comment)

Patryk-Stefanski (approved):

david-martin:
Thanks @Patryk-Stefanski 

I'll merge this now while approved, and follow up with docs changes as per https://github.com/Kuadrant/mcp-gateway/pull/1429#discussion_r3913798920 once #1283 lands

PR #1438, opened 2026-08-26
feat(guardrails): wire request-side checks and propagate maxBodyBytes

### LINKED ISSUE #1403: Guardrails: add request-side tools/call
Wire `Checker` into Router and map its `Decision` onto ext_proc (Part 3 of guardrails feature). 

Depends on #1401. Parent issue: https://github.com/Kuadrant/mcp-gateway/pull/1302.

christinaexyou:
Integrate guardrails Checker into 2025/2026 request routing for tools/call and elicitation accept. Add support for guardrails-modified request content forwarding. Propagate MCPGatewayExtension.spec.maxBodyBytes through controller/config/runtime to bound router buffering and checker response size.

## What does this PR do?
* Wires request-side guardrails into routing by adding `internal/routing/guardrails_check.go` 
* Updates Router 2025/2026 by adding `GuardrailsChecker` as a dep
* Adds guardrails-specific helpers to the MCP request model in `internal/routing/mcp_request.go`
* Copies Guardrails config IDs into route metadata in `internal/broker/routing_table.go` and `internal/routing/router_202511.go`
* Propogates `maxBodyBytes` end to end. Adds new func `requestBodyLimit()` in  `internal/mcp-router/ext_proc_adapter.go` to resolve body cap between the config and CLI arg `--max-request-body-size`. The cap on the defined via the the CLI only applies if it's smaller than the config

Fixes #1403 

## Pre-review checklist

## Summary by CodeRabbit

- **New Features**
  - Added configurable maximum request body size, with a default applied when unspecified.
  - Expanded guardrails protection for tool calls and elicitation responses, including blocking, content modification, and fail-open/closed handling.
  - Added support for applying multiple guardrails configurations to server routes.

- **Bug Fixes**
  - Configuration reloads now apply atomically; invalid updates retain the previously active configuration.
  - Client-facing guardrails errors are sanitized while detailed failure information remains available for diagnostics.


cmd/mcp-broker-router/router.go
```diff
@@ -25,10 +26,22 @@ func (a *app) createRouter() {
 	}
 	a.server.RoutingConfig.Store(a.mcpConfig)
 
+	// shared across both routers so a config reload rebuilding the Checker
+	// (see ExtProcServer.OnConfigChange) is visible to whichever router
+	// handles the request, without either router importing mcp-router.
+	guardrailsChecker := func() guardrails.Checker {
```
maleck13: This reads a bit odd and took me a while to understand. I think this is being forced by the need to avoid an circular import? Suggestion move the guardrails config in to the guardrails package?

```go
internal/guardrails/config.go (new file):
package guardrails

// Config holds the resolved guardrails server config parsed from
// the guardrails Secret referenced by the MCPGatewayExtension.
type Config struct {
      URL       string   `json:"url"                 yaml:"url"`
      ConfigIDs []string `json:"configIDs,omitempty" yaml:"configIDs,omitempty"`
      Model     string   `json:"model"               yaml:"model"`
      FailMode  string   `json:"failMode,omitempty"  yaml:"failMode,omitempty"`
}
```

```go

type MCPServersConfig struct {
      lock                       sync.RWMutex
      // ...existing fields...
      GlobalGuardrails           *guardrails.Config
      MaxBodyBytes               int64
      requestChecker             guardrails.Checker // built by ext_proc adapter on config change
}

func (c *MCPServersConfig) SetRequestChecker(checker guardrails.Checker) {
      c.lock.Lock()
      defer c.lock.Unlock()
      c.requestChecker = checker
}

func (c *MCPServersConfig) GetRequestChecker() guardrails.Checker {
      c.lock.RLock()
      defer c.lock.RUnlock()
      return c.requestChecker
}

```


```go

func (s *ExtProcServer) OnConfigChange(ctx context.Context, newConfig *config.MCPServersConfig) {
      s.buildAndSetChecker(ctx, newConfig)
      s.RoutingConfig.Store(newConfig)
}

func (s *ExtProcServer) buildAndSetChecker(ctx context.Context, cfg *config.MCPServersConfig) {
      gr := cfg.GetGlobalGuardrails()
      if gr == nil {
              cfg.SetRequestChecker(nil)
              return
      }
      tlsCfg, err := tlsConfigFromCACertPEM(cfg.GetGatewayCACertPEM())
      if err != nil {
              s.Logger.ErrorContext(ctx, "failed to build guardrails TLS config", "error", err)
              cfg.SetRequestChecker(nil)
              return
      }
      checker := guardrails.NewChecker(gr, tlsCfg, 0, cfg.GetMaxBodyBytes())
      cfg.SetRequestChecker(checker)
}
```

internal/mcp-router/ext_proc_adapter.go
```diff
@@ -42,11 +45,65 @@ type ExtProcServer struct {
...
+	if cfg := s.RoutingConfig.Load(); cfg != nil {
+		limit = int(cfg.GetMaxBodyBytes())
+	}
+	if s.MaxRequestBodySize > 0 && s.MaxRequestBodySize < limit {
+		return s.MaxRequestBodySize
+	}
+	return limit
+}
+
+func tlsConfigFromCACertPEM(caCertPEM string) (*tls.Config, error) {
```
maleck13: not for this pr perhaps but I think we do this in several places
https://github.com/Kuadrant/mcp-gateway/issues/1442

internal/mcp-router/ext_proc_adapter.go
```diff
@@ -42,11 +45,65 @@ type ExtProcServer struct {
...
+	s.GuardrailsChecker.Store(&checker)
+	s.Logger.InfoContext(ctx, "guardrails checker created")
+}
+
+func (s *ExtProcServer) requestBodyLimit() int {
+	limit := int(config.DefaultMaxBodyBytes)
+	if cfg := s.RoutingConfig.Load(); cfg != nil {
+		limit = int(cfg.GetMaxBodyBytes())
+	}
+	if s.MaxRequestBodySize > 0 && s.MaxRequestBodySize < limit {
```
maleck13, line 57: not for this PR but we should prob remove this and just use the spec based value with default now:

 https://github.com/Kuadrant/mcp-gateway/issues/1443

maleck13: @david-martin we should get #1443 in before GA as it is not back compat

internal/routing/guardrails_check.go
```diff
@@ -0,0 +1,158 @@
...
+
+const guardrailsUnavailableMessage = "guardrails check unavailable"
+
+// checkGuardrailsRequest runs the guardrails check for a tools/call or
+// elicitation accept. blocked is non-nil when the request must not proceed.
+// modified is the guardrails-rewritten arguments/content when Status is
+// StatusModified; callers must apply it before forwarding. Empty IDs skip
+// the check and set guardrails.enabled=false. Non-empty IDs with no Checker
+// fail closed (503).
+func checkGuardrailsRequest(
```
maleck13: This is a pretty large function signature. Do we need to pass all of these at request time?

internal/routing/guardrails_check.go
```diff
@@ -0,0 +1,158 @@
...
+	}
+
+	switch decision.Status {
+	case guardrails.StatusBlocked:
+		if decision.Err != nil {
+			span.SetAttributes(attribute.String("guardrails.status", "error"))
+			return "", jsonRPCErrorDecision(503, requestID, guardrailsUnavailableMessage, buildToolError, contentType)
+		}
+		span.SetAttributes(attribute.String("guardrails.status", "blocked"))
+		return "", jsonRPCErrorDecision(403, requestID, guardrailsBlockedMessage(decision.Reason), buildToolError, contentType)
```
maleck13: I assume it is safe to just return the reason to the client from nemo response? As in nothing leaked from that response that would be internal?

internal/routing/guardrails_check.go
```diff
@@ -0,0 +1,158 @@
...
+	start := time.Now()
+	decision, checkErr := checker.CheckRequest(ctx, name, arguments, serverConfigIDs)
+	span.SetAttributes(attribute.Int64("guardrails.latency_ms", time.Since(start).Milliseconds()))
+	if checkErr != nil {
+		// translation failure only; always a hard deny, failMode does not apply
+		span.SetAttributes(attribute.String("guardrails.status", "error"))
+		if logger != nil {
+			logger.DebugContext(ctx, "guardrails check", "status", "error", "error", checkErr)
+		}
+		return "", jsonRPCErrorDecision(400, requestID, fmt.Sprintf("guardrails: %s", checkErr.Error()), buildToolError, contentType)
```
maleck13: If there is an error in the check, would we be better to log and not send back through to the client? Not sure what the client could do about it ? It is an internal error right?

maleck13 (commented):
Ok so I think the functionality is generally fine, but I think we should consider changing the layout somewhat. There is tension in the packages.

We are using the routing package to load the ability to do guard rails but guard rails isn't a full lib it needs pieces from other packages.  I think this needs to be the other way around to improve the composition and avoid things like:

-  Closure + atomic pointer
- Methods on MCPRequest — Arguments(), SetArgumentsJSON(), that are guard rails specific (the guard rails package could declare a wrapper that adds this functionality with adding it at the top level)

Routing should declare it needs a guard rails implementation via an interface and the guard rails package implements then it is wired together in main. 

skteched outline:

```
routing/
  guardrails.go       // interface + decision type + guardrailsCheck struct
  mcp_request.go     // untouched, no guardrails methods

guardrails/
  config.go           // Config struct (moved from config/)
  checker.go          // nemoChecker implements routing.GuardrailsChecker
  request_adapter.go  // knows how to extract. replace args from map[string]any

config/
  types.go           // stores routing.GuardrailsChecker on MCPServersConfig (typed)

cmd/
  router.go          // builds guardrails.NewChecker, sets it on config

```
The  GuardrailsConfig move from config/ to guardrails/  is key it seems. I don't see this as a blocker per se but rather something that we could probably quite easilly address WDYT?

cmd/mcp-broker-router/main.go
```diff
@@ -431,6 +433,15 @@ func (a *app) loadConfig(path string) error {
 	a.mcpConfig.SetServers(newServers, newVirtualServers)
 	a.mcpConfig.SetGatewayCACertPEM(gatewayCACertPEM)
```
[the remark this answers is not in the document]
maleck13: @christinaexyou prob does make sense to set these after the guardrails config unmarshal? But it is a nit and not a blocker

internal/routing/guardrails.go
```diff
@@ -0,0 +1,273 @@
...
+// toolCallArguments JSON-encodes params["arguments"]. Missing or nil becomes {}.
+func toolCallArguments(params map[string]any) (json.RawMessage, error) {
+	if params == nil {
+		return json.RawMessage(`{}`), nil
+	}
+	args, ok := params["arguments"]
+	if !ok || args == nil {
+		return json.RawMessage(`{}`), nil
+	}
+	raw, err := json.Marshal(args)
```
maleck13, line 250: logged  https://github.com/Kuadrant/mcp-gateway/issues/1466 as a follow on general maintenance. Can prob improve this at the MCPRequest level to avoid having to re-marshal and just pass the raw bytes through.

internal/routing/router_202511.go
```diff
@@ -186,6 +186,22 @@ func (r *Router202511) routeToolCall(ctx context.Context, table RoutingTable, mc
 	mcpReq.ReWriteToolName(upstreamToolName)
 	headers[MCPServerNameHeader] = serverInfo.Name
 
+	args, blocked := guardrailsArguments(mcpReq, mcpReq.ID, BuildSSEJSONRPCError, "")
```
maleck13: This is a lot of internal knowledge for the router about how guard rails works. I am a bit confused as to why this shouldn't / couldn't just be something like:

```
gc := newGuardrailsCheck(r.RoutingConfig.Load(), r.Logger, BuildSSEJSONRPCError, "")
argsModified, blocked := gc.checkToolCall(ctx, serverInfo.GuardrailsConfigIDs, mcpReq, upstreamToolName)
if blocked != nil {
    blocked.SetHeaders = map[string]string{SessionHeader: mcpReq.GetSessionID()}
    return blocked
}


```

Or even better you could also consider using the options pattern for the optional arguments

```go

type guardrailsOption func(*guardrailsCheck)

func withErrorBuilder(fn func(id any, msg string) []byte) guardrailsOption {
    return func(gc *guardrailsCheck) { gc.buildError = fn }
}

func withContentType(ct string) guardrailsOption {
    return func(gc *guardrailsCheck) { gc.contentType = ct }
}

func newGuardrailsCheck(cfg *config.MCPServersConfig, logger *slog.Logger, opts ...guardrailsOption) *guardrailsCheck {
    gc := &guardrailsCheck{
        cfg:         cfg,
        logger:      logger,
        buildError:  BuildJSONRPCError,       // default
        contentType: "application/json",      // default
    }
    for _, o := range opts {
        o(gc)
    }
    return gc
}
```

internal/routing/router_202511.go
```diff
@@ -479,6 +496,25 @@ func (r *Router202511) routeElicitationResponse(ctx context.Context, mcpReq *MCP
...
 
+	if isElicitationAccept(mcpReq) {
+		args, argErr := elicitationArguments(mcpReq.Result)
+		if argErr != nil {
+			blocked := jsonRPCErrorDecision(400, clientID, fmt.Sprintf("guardrails: %s", argErr.Error()), BuildSSEJSONRPCError, "")
+			blocked.SetHeaders = map[string]string{SessionHeader: mcpReq.GetSessionID()}
+			return blocked
+		}
+		gc := newGuardrailsCheck(r.RoutingConfig.Load(), r.Logger, BuildSSEJSONRPCError, "")
+		modified, blocked := gc.request(ctx, mcpServerConfig.GuardrailsConfigIDs, elicitationActionAccept, args, clientID)
```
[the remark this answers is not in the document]
maleck13: @christinaexyou seems a reasonable thing to load the config once here rather than multiple times?

internal/routing/router_202607.go
```diff
@@ -142,6 +142,34 @@ func (r *Router202607) routeToolCall(ctx context.Context, table RoutingTable, re
...
+		rewritten, err := req.Parsed.ToBytes()
+		if err != nil {
+			r.Logger.ErrorContext(ctx, "failed to marshal body to bytes", "error", err)
+			span.SetStatus(codes.Error, "body marshal failed")
+			span.SetAttributes(attribute.String("error.type", "marshal_error"))
+			return &Decision{Error: &Error{StatusCode: 500, Message: "internal error"}}
+		}
+		bodyMutation = rewritten
+	}
+
```
maleck13, line 161: same as above. If we changed as I suggest this becomes (something like):

```go
// with CheckAndModify + options pattern
gc := newGuardrailsCheck(r.RoutingConfig.Load(), r.Logger)
modified, blocked := gc.checkToolCall(ctx, serverInfo.GuardrailsConfigIDs, req.Parsed, upstreamToolName)
if blocked != nil {
    return blocked
}
if modified {
    rewritten, err := req.Parsed.ToBytes()
    if err != nil {
        r.Logger.ErrorContext(ctx, "failed to marshal body to bytes", "error", err)
        span.SetStatus(codes.Error, "body marshal failed")
        span.SetAttributes(attribute.String("error.type", "marshal_error"))
        return &Decision{Error: &Error{StatusCode: 500, Message: "internal error"}}
    }
    bodyMutation = rewritten
}

```

internal/routing/router_202607_test.go
```diff
@@ -520,3 +521,68 @@ func TestRouter202607_BrokerPassthroughReInjectsInternalHeaders(t *testing.T) {
 	require.Equal(t, "signed-jwt", decision.SetHeaders[MCPAuthorizedHeader])
 	require.Equal(t, "test/vs", decision.SetHeaders[MCPVirtualServerHeader])
 }
+
+func TestRouter202607_Guardrails(t *testing.T) {
```
maleck13, line 529: Not sure we have covered the statusmodified case here?

maleck13:
- The functionality generally looks good.
- The architectural/code layout concerns I have called out are not blockers but nice to haves
- The missing modified test (2026 router) and double config load are low-risks but again nice to have/low cost

The e2e tests are failing. I am double checking if we have an issue on main as the failure is odd. Will report back if there is an issue on main. Will wait for everything to be green before approving

internal/routing/router_202511.go
christinaexyou: good point - refactored to use the options pattern

internal/routing/router_202607_test.go
christinaexyou: you're right - updated the test to cover the `StatusModified` case

internal/config/types.go
```diff
@@ -83,6 +93,130 @@ func (config *MCPServersConfig) GetGatewayCACertPEM() string {
...
+	defer config.lock.RUnlock()
+	return config.GlobalGuardrails
+}
+
+// SetGuardrailsChecker stores the checker, or nil when guardrails is disabled.
+// Prefer SetGuardrails when updating GlobalGuardrails in the same step.
+func (config *MCPServersConfig) SetGuardrailsChecker(c guardrails.Checker) {
+	config.lock.Lock()
+	defer config.lock.Unlock()
+	config.guardrailsChecker = c
```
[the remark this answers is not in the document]
christinaexyou: Would this not add a significant amount of latency ? IIUC, this would imply that owner would have to validate every single request and wait for a reply

christinaexyou:
@maleck13 thanks so much for your feedback ! i believe i've addressed all of your comments including nits. there were quite a few changes in the most recent commit [d339074](https://github.com/Kuadrant/mcp-gateway/pull/1438/commits/d339074c5889a866ccbff457dfba01509c35195a), so for posterity's sake, the following changes were made: 
  
  * `cmd/mcp-broker-router/main.go` - we load the Checker and update the upstream servers and guardrails concurrently to prevent them from going out of sync. Also, explicitly fail for bad TLS
  * `cmd/mcp-broker-router/router.go ` - "grouped" parsing of Checker, global guardrails IDs, and per-server IDs
 * `internal/routing/guardrails.go` - simplified router API. refactored to use options format. updated `internal/routing/router_202511.go` and `internal/routing/router_202511.go` to use `checkToolCall` / `checkElicitationAccept` helpers
 * updated unit tests

christinaexyou:
there's some remaining gaps that i want to point out. i haven't addressed them yet bc they require a refactor/changes to existing files. would like your opinion on how we should proceed:
* there's still a circular dependency between the router and internal/guardrails which i circumvent using aliases. i _think_ we can address this by moving the guardrails types such as `Decision` into a separate package in `internal`. wdyt ?
* currently the hairpin goes live, then the `ApplyReload` swaps servers / CA / checker on config. this could break the gateway if the CA is bad. should we initialize the hairpin first -> do `ApplyReload` -> validate the CA -> reload the hairpin so it goes live

maleck13:
> there's some remaining gaps that i want to point out. i haven't addressed them yet bc they require a refactor/changes to existing files. would like your opinion on how we should proceed:
> 
> * there's still a circular dependency between the router and internal/guardrails which i circumvent using aliases. i _think_ we can address this by moving the guardrails types such as `Decision` into a separate package in `internal`. wdyt ?

Yes this seems reasonable 👍 

> * currently the hairpin goes live, then the `ApplyReload` swaps servers / CA / checker on config. this could break the gateway if the CA is bad. should we initialize the hairpin first -> do `ApplyReload` -> validate the CA -> reload the hairpin so it goes live

If I follow, you mean if the CA for the nemo server is correct, then updated to be bad, it will break tool calls as they will likely have a fail mode deny so you are suggesting validating against the nemo server first before reloading the clients? If bad keep the old config and log the issue? What about on start up fail fast seems appropriate?

christinaexyou:
> > * currently the hairpin goes live, then the `ApplyReload` swaps servers / CA / checker on config. this could break the gateway if the CA is bad. should we initialize the hairpin first -> do `ApplyReload` -> validate the CA -> reload the hairpin so it goes live
> 
> If I follow, you mean if the CA for the nemo server is correct, then updated to be bad, it will break tool calls as they will likely have a fail mode deny so you are suggesting validating against the nemo server first before reloading the clients? If bad keep the old config and log the issue? What about on start up fail fast seems appropriate?

> yes you're correct; the core issue is that the hairpin implicitly assumes that the CA / config is correct. it goes live without checking any of the resources that `ApplyReload deploys

maleck13:
@christinaexyou perhaps we can file a follow up issue for the CA check ? If there is a bad CA it would break things but that is an edge case and also there would be logs and a workaround of fixing the CA. I think more important right now to land the guard rails integration so we can test with it in place

cmd/mcp-broker-router/main.go
```diff
@@ -448,5 +464,48 @@ func (a *app) loadConfig(path string) error {
...
+	case nil:
+		if a.mcpConfig.GetGuardrailsChecker() != nil {
+			a.logger.InfoContext(ctx, "guardrails checker destroyed")
+		}
+	default:
+		tlsConfig, err := tlsConfigFromCACertPEM(snap.gatewayCACertPEM)
+		if err != nil {
+			return fmt.Errorf("building guardrails TLS config: %w", err)
+		}
+		checker = guardrails.NewChecker(snap.globalGuardrails, tlsConfig, 0, snap.maxBodyBytes)
```
maleck13, line 512: just spotted this. We create a new HTTPClient each time the config changes. So there could be idle connections hanging around waiting to be cleaned up and we only actually need a new client if the config has changed (which would seem infrequently) ?

```
sudo code
guardrailsUnchanged := parsed.globalGuardrails.Equal(a.currentGuardrailsConfig)
caUnchanged := parsed.gatewayCACertPEM == a.currentCACertPEM
```

if we check with something like these before creating a new client we can avoid unneeded client setup etc .  We could also implement a  `io.Closer` on Checker so we can call Close() to clean up idle connections?

internal/routing/router_202607.go
```diff
@@ -142,6 +142,23 @@ func (r *Router202607) routeToolCall(ctx context.Context, table RoutingTable, re
 	if routerErr != nil {
 		return &Decision{Error: routerErr}
 	}
+
+	gc := newGuardrailsCheck(r.RoutingConfig.Load(), serverInfo.Name, r.Logger)
```
maleck13: Sorry if I missed this before but can we check the serverIDs need to be calculated on each request? I think they are already present on `serverInfo` via the routing table build?  
We seem to do fresh iteration of the servers in GuardrailsSnapshotFor. I am not sure we need to do this snapShotFor as the routing table is rebuilt on each config change . This will happen for every tool call on the hot path.  RoutingTableFunc returns a snapshot already, the broker builds a new routing config on every change, the router calls r.Table() once per request and uses the returned value.

christinaexyou: ack - i added `GuardrailsSnapshotFor` to have a single helper function shared amongst the 2025 and 2026 router as well as elicitation. i removed this in commit [3b1259a](https://github.com/Kuadrant/mcp-gateway/pull/1438/commits/3b1259a0906377cbfff7eadd8144a11a0510c695) by reading serverIDs during broker build time

cmd/mcp-broker-router/main.go
christinaexyou: good catch - i think we should have the cleanup be transparent to the user rather than raise `io.Closer` type assertion so i added `Checker.Close()` path on the interface via commit 1b96b31

internal/routing/router_202607.go
christinaexyou: since elicitations depended on `GuardrailsSnapshotFor` and iiuc, cannot use the routing table, i added a new helper `GuardrarilsForServer` to cover the elicitation path

maleck13 (approved):
Thanks @christinaexyou this looks good to me now.

PR #1441, opened 2026-08-27
fix(e2e): skip combined Full tier2 tags in PR gate

david-martin:
Small fix for skip regex.
Saves ~1m45s on PR e2e tests. 
The intent was for them to be skipped already.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated end-to-end test filtering so exclusions correctly match standalone tags and tags combined with other labels.
  * Preserved existing filtering behavior for A2A-tagged tests.


Aman-Cool:
thanks for grabbing this @david-martin! makes sense.., `\[Full(,|\])` catching the combined `[Full,…]` tags is the clean fix.

trepel (approved):
LGTM!

Issue #1444, opened 2026-08-27
refactor: centralise config secret writes and remove --max-request-body-size flag


maleck13:
## Problem

The MCPGatewayExtension controller does multiple read-modify-write cycles against the same config Secret per reconcile. Each gateway-owned field (`gatewayCACertPEM`, `globalGuardrails`) is written independently via `WriteCACertBundle` and `WriteGlobalGuardrails`, each doing its own `readOrCreateConfigSecret` → modify → `Update()` wrapped in `retry.RetryOnConflict`. With `EnsureConfigExists` that's 3 API server round trips for one reconcile, and each write can conflict with the previous one's `resourceVersion` bump.

Adding more gateway-owned fields (e.g. `maxBodyBytes` from #1438) makes this worse.

### Centralise config writes

- Add a `GatewayConfig` struct grouping all gateway-owned fields
- Add a single `WriteGatewayConfig` method: one read, compare all fields, at most one write
- Refactor `reconcileCACertBundle` → `resolveCACertBundle` (returns value, no I/O)
- Refactor `reconcileGuardrails` → `resolveGuardrails` (returns value, no I/O)
- Collect resolved values in `reconcileActive`, call `WriteGatewayConfig` once
- Remove `EnsureConfigExists`, `WriteCACertBundle`, `WriteGlobalGuardrails` from `ConfigWriterDeleter` interface
- Simplify tests: resolve functions become pure validation, easier to test without mock writers

### Include maxBodyBytes in the single write (#1438 follow-up)

Add `maxBodyBytes` to `GatewayConfig` so it's written in the same single cycle.

### Remove --max-request-body-size CLI flag (#1443)

With `maxBodyBytes` propagated via the CRD, the CLI flag is redundant:

- Remove `--max-request-body-size` from `cmd/mcp-broker-router/main.go`
- Remove `MaxRequestBodySize` from `ExtProcServer`
- Simplify `requestBodyLimit()` to use only config
- Document migration in `docs/release-notes/`

## Breaking change

`--max-request-body-size` has been present since v0.8.0. Users passing it in container args will need to set `spec.maxBodyBytes` on their MCPGatewayExtension instead.

Closes #1443

PR #1445, opened 2026-08-27
reduce the number of config writes we do. Remove the max-request-body…

maleck13:
… flag as we are adding a spec field for this and this flag is unlikely to be used anywhere right now

## What does this PR do?

We have multiple places that write the config secret in the controller. This works fine as there is a retry but it means multiple writes and reads from the API server where they are not needed and multiple places in the code where writes are done. This centeralizes the writes and reduces the number of calls we need to make

In addition we have a flag that limits the body request size. This was never really exposed and documented or used. We are introducing a spec field to manage this instead so removing this now before we go GA

Fixes #

#1444 

## Summary by CodeRabbit

* **New Features**
  * Gateway CA certificates, guardrails, and maximum request-body size can be updated together.
  * Configuration updates avoid unnecessary writes and handle conflicts more reliably.
  * EnvoyFilter processing adapts automatically when the required custom resource is unavailable and resumes when available.
  * The default maximum request-body size is now 5 MiB.

* **Breaking Changes**
  * Removed the `--max-request-body-size` option.
  * Configure request limits through `MCPGatewayExtension.spec.maxBodyBytes`; deployments using the removed option must update their startup arguments.


docs/release-notes/1443-remove-max-request-body-size-flag.md
```diff
@@ -0,0 +1,41 @@
...
+    args:
+      - --max-request-body-size=10485760  # remove this line
+
+# After
+containers:
+  - name: mcp-broker-router
+    args: []
+```
+
+Once `spec.maxBodyBytes` is available on MCPGatewayExtension, use that to
```
david-martin, line 33: Given this doesn't exist in the spec, I'd prefer we don't even mention it as it could cause confusion.
I was confused by it, thinking it was meant to be implemented in this PR.

internal/broker/http_compat.go
```diff
@@ -117,8 +117,8 @@ const (
 const mainDefaultProtocolVersion = "2025-03-26"
 
 // maxRequestBodyBytes bounds POST body reads. in production the router's
-// ext_proc bounds bodies before they reach the broker (--max-request-body-size,
-// 5MB default); this guards direct broker access against unbounded allocation.
+// ext_proc bounds bodies before they reach the broker (5 MiB default);
+// this guards direct broker access against unbounded allocation.
```
david-martin, line 121: 5 MiB mentioned, but the const below is 10MiB

david-martin (changes requested):
comments inline.

Also, should this land before or after #1438? There's some overlap in files changed.

maleck13:
correct we need this to merge first https://github.com/Kuadrant/mcp-gateway/pull/1438

docs/release-notes/1443-remove-max-request-body-size-flag.md
maleck13: This is on the spec now. We prob don't need this doc tbh anyway?

internal/config/types.go
```diff
@@ -358,6 +361,12 @@ type BrokerConfig struct {
 	MaxBodyBytes int64 `json:"maxBodyBytes,omitempty" yaml:"maxBodyBytes,omitempty"`
 }
 
+// GatewayConfig holds gateway-level configuration fields
+type GatewayConfig struct {
+	CACertPEM  string            `json:"caCertPEM,omitempty"  yaml:"caCertPEM,omitempty"`
+	Guardrails *GuardrailsConfig `json:"guardrails,omitempty" yaml:"guardrails,omitempty"`
```
Patryk-Stefanski: Please carry MaxBodyBytes in GatewayConfig and populate it here. Removing reconcileMaxBodyBytes without adding it to this combined write drops spec.maxBodyBytes from the generated Secret.

internal/mcp-router/ext_proc_adapter.go
```diff
@@ -321,8 +309,8 @@ func (s *ExtProcServer) Process(stream extProcV3.ExternalProcessor_ProcessServer
 			s.Logger.DebugContext(ctx, "[ext_proc ] Process: ProcessingRequest_RequestBody", "request id:", requestID)
 			body := r.RequestBody.Body
 
-			if limit := s.requestBodyLimit(); limit > 0 && len(body) > limit {
-				err := fmt.Errorf("request body too large: %d bytes exceeds limit of %d", len(body), limit)
+			if len(body) > config.MaxRequestBodySize {
```
Patryk-Stefanski: This restores a fixed 5 MiB cap. Use the configured limit (GetMaxBodyBytes) here so MCPGatewayExtension.spec.maxBodyBytes remains effective.

docs/release-notes/1443-remove-max-request-body-size-flag.md
```diff
@@ -0,0 +1,41 @@
+# Remove --max-request-body-size CLI Flag
+
+The `--max-request-body-size` CLI flag on `mcp-broker-router` has been removed.
+The body size limit is now a fixed 5 MiB default, matching the previous flag
+default. It will be replaced by `MCPGatewayExtension.spec.maxBodyBytes` for
```
Patryk-Stefanski: This is stale now that spec.maxBodyBytes exists via #1438. Document the current field and default instead of saying it will be added later.

internal/config/types.go
maleck13: Thanks @Patryk-Stefanski I had intended to take a closer look at these changes as it was a rebase and wasn't fully confident it had gone the right way but other things were pulling at my attention. Will address 👍

maleck13:
@Patryk-Stefanski thanks for the review and catching those issues 👍 this should be in good shape now

internal/config/types.go
```diff
@@ -188,8 +188,8 @@ func (config *MCPServersConfig) ApplyReload(
 	config.guardrailsChecker = checker
 }
 
-// DefaultMaxBodyBytes is the MCPGatewayExtension maxBodyBytes default (1 MiB).
-const DefaultMaxBodyBytes int64 = 1 << 20
+// DefaultMaxBodyBytes is the MCPGatewayExtension maxBodyBytes default (5 MiB).
+const DefaultMaxBodyBytes int64 = 5 << 20
```
Patryk-Stefanski, line 192: The CRD still declares default: 1048576 (1 MiB) in api/v1 and generated CRDs, while this changes the runtime default to 5 MiB and the release note says 5 MiB. With spec.maxBodyBytes omitted, Kubernetes supplies 1 MiB, so these paths disagree. Keep the existing 1 MiB default or update the API markers/generated CRDs and related docs together.

maleck13: aligned those now

Patryk-Stefanski (approved):

david-martin (approved):

PR #1449, opened 2026-08-27
fix: compare only controller-owned managed volume fields

### LINKED ISSUE #1396: Harden deploymentNeedsUpdate against server-defaulted fields in managed volumes
`deploymentNeedsUpdate` compares managed volumes with `equality.Semantic.DeepEqual` over the full `SecretVolumeSource` struct, including fields Kubernetes fills in server-side like `DefaultMode`. If the controller's desired spec leaves such a field unset, the live object has it defaulted (`DefaultMode=420`), so the comparison always reports a diff and re-rolls the broker-router deployment every reconcile.

This already affected the `gateway-ca-cert-volume` (fixed in #1305 by setting `DefaultMode: 420` to match `config-volume`). That fix follows the existing convention, but it's a per-field patch: any future managed volume that omits a defaulted field reintroduces the same behaviour.

Options to harden the comparison itself:

- Normalize desired and existing through the same defaulting before `DeepEqual` (e.g. apply the API scheme defaults, or a `corev1.SetDefaults_*` pass)
- Compare only the fields the controller sets, per managed volume, rather than the whole struct
- Diff against a server-side-apply managed-fields view instead of a hand-rolled struct compare

Scope is the volume/mount/env/port comparison in `internal/controller/broker_router.go` (`deploymentNeedsUpdate` and the `filterManaged*` helpers). Not urgent, no current known-broken field beyond the one already fixed.

ifox777:
## What does this PR do?

`deploymentNeedsUpdate` compared complete managed `Volume` values, so Kubernetes defaults or user values in fields the controller left unset looked like drift and could trigger unnecessary broker-router rollouts.

This change matches managed volumes by name and only compares Secret fields that appear in desired state. It still catches changed Secret references, explicitly set fields, missing volumes, and stale `gateway-ca`, while leaving user-added volumes and controller-omitted fields alone.

Fixes #1396

## Verification

- `go test ./internal/controller -run '^(TestDeploymentNeedsUpdate|TestMergeVolumes)$' -count=1`
- `make test-unit`
- `make test-controller-integration` (45/45 specs passed)
- `git diff --check`

The local `make lint` run is currently blocked by 35 pre-existing findings (`gosec` and `noctx`); none point to lines added by this change.

## Pre-review checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved deployment update detection for managed secret volumes.
  * Prevents unnecessary updates when unrelated, user-managed volume settings change.
  * Ensures updates occur when controller-managed secret configuration is missing, stale, or changed.
  * Correctly distinguishes between controller-owned and unowned settings such as default modes, optional values, and item mappings.

### DIFF
--- a/internal/controller/broker_router.go
+++ b/internal/controller/broker_router.go
@@ -554,7 +554,7 @@ func deploymentNeedsUpdate(desired, existing *appsv1.Deployment) (bool, string) 
 	}
 	desiredVolumes := filterManagedVolumes(desired.Spec.Template.Spec.Volumes)
 	existingVolumes := filterManagedVolumes(existing.Spec.Template.Spec.Volumes)
-	if !equality.Semantic.DeepEqual(desiredVolumes, existingVolumes) {
+	if !managedVolumesEqual(desiredVolumes, existingVolumes) {
 		return true, fmt.Sprintf("volumes changed: %+v -> %+v", existingVolumes, desiredVolumes)
 	}
 	// only compare env vars the controller manages; user-added env vars are preserved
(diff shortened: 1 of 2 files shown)


david-martin:
Thanks for opening this @ifox777 .
As the linked issue is in the not yet in a specific milestone ('future'), there'll likely be a delay before a review here while we focus on v1.0.
I'd prefer to hold off merging either way until after v1.0 has been cut to limit changes this close to release

PR #1450, opened 2026-08-27
feat(broker): federate stateless upstreams on older 2025 protocol revisions

### LINKED ISSUE #1436: Broker silently excludes upstreams negotiating a non-pinned 2025-series protocol version (e.g. 2025-03-26) from all served sets
## Summary

With broker `ghcr.io/kuadrant/mcp-gateway:v0.9.0`, an upstream MCP server whose **negotiated** protocol version is a 2025-series revision **other than** the pinned `2025-11-25` (e.g. `2025-03-26` or `2025-06-18`) is **discovered and validated** but served to **zero** clients — and there is **no log or warning** explaining why. This is very hard to diagnose: the server looks healthy in every metric but is invisible on the gateway.

## Root cause

`internal/broker/upstream/mcp.go` records `supportedVersions = []string{ up.init.ProtocolVersion }` — a single element, the negotiated version. `internal/broker/protocol_filter.go` (`rebuildProtocolCaches`) then partitions served tools with `ServerSupportsVersion(id, Version2025)` / `Version2026`, where `Version2025 = "2025-11-25"` and `Version2026 = "2026-07-28"` (`internal/protocol/version.go`). A server negotiating `2025-03-26` matches **neither** exact string, so it lands in neither the stateful (2025) nor the stateless (2026) served set.

## Why it is common

The MCP TypeScript SDK negotiates the client-requested version only if it is in `SUPPORTED_PROTOCOL_VERSIONS`, else it returns its own `LATEST_PROTOCOL_VERSION`. Older SDKs are widespread: `@modelcontextprotocol/sdk` ≤ 1.11 tops out at `2025-03-26`, and 1.12–1.23 at `2025-06-18`; only 1.25+ knows `2025-11-25`. Any such upstream is silently unfederatable, even though `2025-03-26`/`2025-06-18` are valid stateful revisions whose `tools/list` + `tools/call` wire format the broker already speaks.

## Evidence (per-upstream broker metrics vs served output)

For the affected upstream:
```
mcp_broker_tools_discovered{server_name=".../affected"}          44
mcp_broker_tools_list_response_bytes{server_name=".../affected"} 34589
mcp_broker_discovery_total{server_name=".../affected",status="success"} >0
```
…yet the broker's own `POST /mcp` `tools/list` returns **0** tools for that server, while peers negotiating `2025-11-25` serve fully. Pointing the same upstream at an SDK/shim that negotiates `2025-11-25` makes all 44 tools appear immediately — confirming the negotiated-version string is the sole gate.

## Expected

Either:
- treat any `2025-*` stateful revision (`2025-03-26`, `2025-06-18`, `2025-11-25`) as belonging to the 2025/stateful served set (version-range membership, not exact-string match); and/or
- at minimum, emit a `WARN` when an upstream is excluded from all served sets because its negotiated protocol version matches no served bucket, so this is diagnosable instead of silent.

## Version

Broker `v0.9.0`.

rwlove:
Closes #1436

## Problem

The broker only federates an upstream whose advertised protocol versions contain **exactly** `2025-11-25` (`Version2025`) or `2026-07-28` (`Version2026`). `rebuildProtocolCaches` partitions each upstream's tools/prompts into the stateful or stateless served set on that exact match; an upstream that advertises neither lands in **neither set and is silently served to no client** — no error, no log.

In practice this drops any upstream built on an older MCP SDK. The broker requests `2025-11-25` at initialize, the upstream down-negotiates to the newest revision *it* knows (e.g. `2025-03-26` on `@modelcontextprotocol/sdk` ≤ 1.24.x), and its tools vanish. Separately, such upstreams frequently run the streamable-HTTP transport in **stateless** mode (`sessionIdGenerator` disabled → no `Mcp-Session-Id`); the broker treats any non-2026 upstream as stateful and (a) sends a session ping and (b) opens a session-keyed GET SSE watcher, both of which a session-less server rejects — de-federating it after the failure threshold.

Net effect: a perfectly functional, spec-compliant stateless `2025-xx` MCP server cannot be used through the gateway.

## Change

Decouple two things the broker currently conflates: which upstreams are *served*, and which are *stateless*.

**1. Serve the whole 2025 family, not just `2025-11-25`.** An upstream is routed onto the stateful set if it advertises **any** `2025-` revision (`serverServesOnStatefulRoute`). `tools/list` and `tools/call` payloads are wire-compatible across 2025 revisions, so they are served on the existing stateful route. Applied consistently at every stateful decision point so tools don't just *list* but also *route* and *fetch*:
- `rebuildProtocolCaches` (tools + prompts served-set partition)
- `buildRoutingTable` (`Stateful` flag → so calls route)
- `FetchUserSpecificTools` (per-user fetch for `userSpecificList` upstreams)

The stateless (`2026-07-28`) route stays an exact match. (Refactored the version lookup into `serverProtocolVersions`, shared by a lock-free `supportsStatefulRoute` helper.)

**2. Detect statelessness by transport, not by version.** Statelessness is "the upstream issued no `Mcp-Session-Id`" — independent of protocol version (`IsSessionless()`). For a session-less upstream:
- `Ping()` returns nil — a session ping is rejected by strict stateless servers; a successful `Connect` is the connectivity proof, matching the existing 2026 path.
- the GET SSE notification watcher is skipped — there is no session-keyed stream to watch.
- the connection is **recycled on every health tick** (same as 2026 upstreams), so liveness/recovery is preserved: a dead upstream is caught by the next `Connect` failure → failure threshold → tools dropped + reconnect. (Without this, a session-less upstream had *no* working liveness signal once the session-ping was dropped — it would keep serving stale tools indefinitely.)

## Tests (all green: `go test ./internal/broker/...`, `go vet` clean, gofmt clean)

- `TestBroker_serverServesOnStatefulRoute` — the family predicate (exact `2025-11-25`, older `2025-03-26`/`2025-06-18`, `2026`-only → not stateful, `2024`-only, mixed, empty, unknown).
- `TestRebuildProtocolCaches` — two new cases: an older-2025 upstream's tools land in `statefulTools`; a pre-2025 upstream stays dropped.
- `TestMCPManager_manage_RecyclesSessionlessUpstream` — a session-less upstream is recycled on a timer tick (liveness); a session-ful stateful upstream is not.
- Existing `TestBroker_ServerSupportsVersion` still passes through the `serverProtocolVersions` refactor.

## Verification scope

Unit/component-tested against the mock upstream. **Not** yet exercised end-to-end against a real session-less `2025-03-26` server over the wire — happy to add an integration test if you point me at the preferred harness. Verified locally with Go 1.26.6 (`go.mod` toolchain).

## Motivating case

A stock `baruchiro/paperless-mcp` (stateless streamable-HTTP, `@modelcontextprotocol/sdk` 1.11 → negotiates `2025-03-26`) currently connects but is served to no client. With this change its tools federate on the stateful route.

## Summary by CodeRabbit

- **New Features**
  - Expanded support for all 2025 protocol revisions, improving routing for stateful connections.
  - Added support for sessionless 2025 upstream connections without unnecessary session handling.
- **Bug Fixes**
  - Corrected tool and prompt routing for older 2025 protocol versions.
  - Improved connection health checks and recycling for sessionless upstreams.
  - Prevented unsupported session pings and notification watchers for sessionless connections.
- **Tests**
  - Added coverage for mixed, older, newer, and unknown protocol configurations.

### DIFF
--- a/internal/broker/broker.go
+++ b/internal/broker/broker.go
@@ -988,16 +988,14 @@ func (m *mcpBrokerImpl) IsReady() bool {
 	return false
 }
 
-// ServerSupportsVersion returns true if the given upstream server supports the specified protocol version.
-// Returns false if the server is not found or version info is unavailable. Lazily populates the cache
-// from the manager on first access.
-func (m *mcpBrokerImpl) ServerSupportsVersion(id config.UpstreamMCPID, version string) bool {
+// serverProtocolVersions returns the protocol versions the given upstream
+// supports, lazily populating the cache from the manager on first access.
+// Returns nil if the server is unknown or has not reported versions yet.
+func (m *mcpBrokerImpl) serverProtocolVersions(id config.UpstreamMCPID) []string {
 	// try cached value first
-	val, ok := m.serverVersions.Load(id)
-	if ok {
-		versions, ok := val.([]string)
-		if ok {
-			return slices.Contains(versions, version)
+	if val, ok := m.serverVersions.Load(id); ok {
+		if versions, ok := val.([]string); ok {
+			return versions
 		}
 	}
 
@@ -1006,15 +1004,22 @@ func (m *mcpBrokerImpl) ServerSupportsVersion(id config.UpstreamMCPID, version s
 	mgr, found := m.mcpServers[id]
 	m.mcpLock.RUnlock()
 	if !found {
-		return false
+		return nil
 	}
 
 	versions := mgr.SupportedVersions()
 	if versions == nil {
-		return false
+		return nil
 	}
 
 	// cache for next time
 	m.serverVersions.Store(id, versions)
-	return slices.Contains(versions, version)
+	return versions
+}
+
+// ServerSupportsVersion returns true if the given upstream server supports the specified protocol version.
+// Returns false if the server is not found or version info is unavailable. Lazily populates the cache
+// from the manager on first access.
+func (m *mcpBrokerImpl) ServerSupportsVersion(id config.UpstreamMCPID, version string) bool {
+	return slices.Contains(m.serverProtocolVersions(id), version)
 }
--- a/internal/broker/protocol_filter.go
+++ b/internal/broker/protocol_filter.go
@@ -3,11 +3,43 @@ package broker
 import (
 	"maps"
 	"slices"
+	"strings"
 
 	"github.com/Kuadrant/mcp-gateway/internal/config"
 	"github.com/Kuadrant/mcp-gateway/internal/protocol"
 	"github.com/modelcontextprotocol/go-sdk/mcp"
 )
+
+// statefulVersionPrefix matches every revision in the stateful protocol family
+// (2025-03-26, 2025-06-18, 2025-11-25, ...). The broker requests Version2025
+// (2025-11-25) at initialize, but upstreams built on older MCP SDKs downgrade
+// to an earlier 2025 revision. Their tools/list and tools/call payloads are
+// wire-compatible with what the broker serves downstream, so an upstream that
+// advertises any 2025-series revision is served on the stateful route rather
+// than silently dropped from every served set.
+const statefulVersionPrefix = "2025-"
+
+// supportsStatefulRoute reports whether any of the given advertised protocol
+// versions belongs to the stateful (2025) family. Lock-free so it can be
+// called from paths that already hold mcpLock (e.g. buildRoutingTable).
+func supportsStatefulRoute(versions []string) bool {
+	for _, v := range versions {
+		if strings.HasPrefix(v, statefulVersionPrefix) {
+			return true
+		}
+	}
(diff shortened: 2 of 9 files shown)


rwlove:
Scoping / follow-up note: this PR covers the **2025 branch**. #1483 is the same root cause on the **2026 stateless route** — a `2026-07-28` stateless-only upstream whose tools are fetched but never surface to clients, because `isStatelessProtocol` requires the client's header to equal `2026-07-28` exactly and those tools live only in the stateless set. This PR intentionally kept the 2026 route an exact match, so it doesn't cover that branch.

If useful, I can extend the same "`tools/list` is wire-compatible, so serve it on the stateful route" argument to the 2026 stateless set and fold #1483 into the fix.

tommyc2:
restored PR:

https://github.com/Kuadrant/mcp-gateway/pull/1490

PR #1462, opened 2026-09-01
fix: exclude private-scope server tools when no prefix configured

### LINKED ISSUE #1385: Broker: exclude private-scope server tools when no prefix configured
## Problem

Servers with `cacheScope: "private"` (or `userSpecificList: true`) return per-user tool lists. These tools may have names not present in the shared routing table, which is populated from the broker's initial `credentialRef` fetch. The router resolves `tools/call` via `LookupTool` (exact match) then falls back to `LookupPrefix` (prefix match). Without a prefix configured on the MCPServerRegistration, neither lookup succeeds for user-specific tool names — the tool is listed but unroutable.

All existing e2e tests for user-specific servers use a prefix, so this gap is not caught today.

### 1. Broker: exclude tools from private-scope servers with no prefix

When the broker detects `cacheScope: "private"` (from a 2026 upstream) or `userSpecificList: true` (from the CRD) on a server that has no prefix configured, it should **not include that server's tools in `tools/list` responses**. Listing unroutable tools is worse than omitting them — clients will attempt `tools/call` and get a routing error.

### 2. Broker: log a warning when private scope + no prefix

When the broker encounters this configuration, it should emit a structured warning log:

```
logger.Warn("private-scope server has no prefix configured, tools excluded from listing",
    "server", serverName, "reason", "tools would be unroutable without prefix for LookupPrefix fallback")
```

This helps operators diagnose why tools from a specific server are missing.

### 3. Document the constraint

Add to the broker design doc (`docs/design/broker-2026-07-28`) constraints section:

> Private-scope servers require a prefix. Servers with `cacheScope: "private"` or `userSpecificList: true` must configure a prefix on their MCPServerRegistration. Without a prefix, per-user tools not in the shared routing table cannot be routed by the router's `LookupPrefix` fallback, so the broker excludes them from `tools/list`.

### 4. Add e2e test coverage

- Negative test: private-scope server with no prefix — verify its tools are excluded from `tools/list`

## Context

- Router design: `docs/design/router-2026-07-28/router-2026-07-28-design.md` (LookupPrefix fallback)
- Broker design: `docs/design/broker-2026-07-28/broker-2026-07-28-design.md` (cacheScope, userSpecificList)
- Existing e2e tests: `tests/e2e/user_specific_list_test.go` (all use prefix)

tommyc2:
## What does this PR do?

Excludes private scope server tools when no prefix is configured

Fixes https://github.com/Kuadrant/mcp-gateway/issues/1385

## Pre-review checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Private-scope servers without a configured prefix are excluded from shared tool listings when their tools cannot be routed.
  * Per-user tool requests now skip unroutable private-scope servers.
  * Unroutable tools are consistently excluded from cached and per-request results.
  * Private-scope servers with prefixes continue to function normally.
  * A2A passthrough now handles protocol metadata headers consistently, rejects unlabelable requests safely, and preserves GET behavior.

* **Documentation**
  * Documented the requirement for private-scope servers to configure a prefix.

### DIFF
--- a/internal/broker/protocol_filter.go
+++ b/internal/broker/protocol_filter.go
@@ -67,7 +67,7 @@ func (m *mcpBrokerImpl) privateScopeServersWithoutPrefix() map[config.UpstreamMC
 		}
 		excluded[id] = struct{}{}
 		m.logger.Warn("private-scope server has no prefix configured, tools excluded from listing",
-			"server", mgr.Config().Name,
+			"server", mgr.MCPName(),
 			"reason", "tools would be unroutable without prefix for LookupPrefix fallback")
 	}
 	return excluded
--- a/internal/broker/protocol_filter.go
+++ b/internal/broker/protocol_filter.go
@@ -44,11 +44,7 @@ type protocolCacheEntry[T any] struct {
 }
 
 // isPrivateScopeWithoutPrefix reports whether an upstream's tools are
-// unroutable and must be excluded from tools/list. A server whose list is
-// per-user — cacheScope:"private" (2026 upstream) or userSpecificList (CRD) —
-// returns tool names absent from the shared routing table; without a prefix the
-// router's LookupPrefix fallback has nothing to match, so the tools cannot be
-// routed.
+// unroutable and must be excluded from tools/list.
 func isPrivateScopeWithoutPrefix(s upstream.ActiveMCPServer) bool {
 	cfg := s.Config()
 	if cfg.Prefix != "" {
--- a/internal/broker/protocol_filter.go
+++ b/internal/broker/protocol_filter.go
@@ -4,6 +4,7 @@ import (
 	"maps"
 	"slices"
 
+	"github.com/Kuadrant/mcp-gateway/internal/broker/upstream"
 	"github.com/Kuadrant/mcp-gateway/internal/config"
 	"github.com/Kuadrant/mcp-gateway/internal/protocol"
 	"github.com/modelcontextprotocol/go-sdk/mcp"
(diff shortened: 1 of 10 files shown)


internal/broker/protocol_filter.go
```diff
@@ -42,11 +43,48 @@ type protocolCacheEntry[T any] struct {
 	freshFetchServers []userSpecificServer
 }
 
+// isPrivateScopeWithoutPrefix reports whether an upstream's tools are
+// unroutable and must be excluded from tools/list. A server whose list is
+// per-user — cacheScope:"private" (2026 upstream) or userSpecificList (CRD) —
+// returns tool names absent from the shared routing table; without a prefix the
+// router's LookupPrefix fallback has nothing to match, so the tools cannot be
+// routed.
```
david-martin: we can remove this comment. the important bit is the first sentence, i.e. what the func does (and is readable from the code in my opinion, and the func name).
The latter part is hinting at how the func is used elsewhere in the code.

david-martin (changes requested):
1 small suggestion.
The rest looks reasonable.

Needs DCO (see the CI failure).

Also, I'd like to get @maleck13 review on this as the issue author.

internal/broker/protocol_filter.go
tommyc2: was going to mention the commit signing but i didnt want to change anything while you were reviewing, so ill fix that

internal/broker/protocol_filter.go
```diff
@@ -42,11 +43,44 @@ type protocolCacheEntry[T any] struct {
...
+	for id, mgr := range m.mcpServers {
+		if !isPrivateScopeWithoutPrefix(mgr) {
+			continue
+		}
+		if excluded == nil {
+			excluded = make(map[config.UpstreamMCPID]struct{})
+		}
+		excluded[id] = struct{}{}
+		m.logger.Warn("private-scope server has no prefix configured, tools excluded from listing",
+			"server", mgr.Config().Name,
```
maleck13: nit: this actually copies and allocates a new copy of the config as they are read only methods. So not ideal just to log the name. Can be a follow up but could expose something like GetName()string that just reads the name

internal/broker/user_specific_tools_test.go
```diff
@@ -650,6 +650,127 @@ func TestFetchUserSpecificTools_StatelessFetch(t *testing.T) {
...
+				assert.Contains(t, names, tt.wantTool, "included server's tool must be merged")
+			}
+		})
+	}
+}
+
+// the per-request path must skip excluded servers silently: the warning is
+// emitted once at rebuild, never per request (hot-path rule).
+func TestFetchUserSpecificTools_NoWarnOnExclusion(t *testing.T) {
+	const msg = "private-scope server has no prefix configured, tools excluded from listing"
```
maleck13: Nit: This test seems quite brittle as it depends on specific log message ? If the message chaged in the code this would fail silently. Not sure it has any real value as a test

maleck13 (dismissed):
small nits but ok to merge

david-martin (approved):

PR #1476, opened 2026-09-03
fix: handle missing Istio EnvoyFilter CRD

### LINKED ISSUE #1465: Controller panics on startup when Istio CRDs are not installed
Found while integrating mcp-gateway as an embedded child component of kuadrant-operator ([RFC 0019](https://github.com/Kuadrant/architecture/blob/main/rfcs/0019-olmv1-operator-consolidation.md) consolidation work).

## Problem

`MCPGatewayExtensionReconciler.SetupWithManager` unconditionally registers a watch on the Istio `EnvoyFilter` CRD:

https://github.com/Kuadrant/mcp-gateway/blob/main/internal/controller/mcpgatewayextension_controller.go#L935

```go
Watches(&istionetv1alpha3.EnvoyFilter{}, handler.EnqueueRequestsFromMapFunc(r.enqueueMCPGatewayExtForEnvoyFilter)).
```

If the Istio CRDs aren't installed on the cluster, controller-runtime's RESTMapper can't resolve `EnvoyFilter` to a kind, so `Complete(r)` returns an error. `cmd/main.go` wraps every setup error in a bare `panic`:

```go
if err = (&controller.MCPGatewayExtensionReconciler{...}).SetupWithManager(ctx, mgr); err != nil {
    panic("unable to start manager : " + err.Error())
}
```

So the controller crashes on startup, and since nothing about the failure is transient, it crash-loops indefinitely rather than starting up in a degraded/waiting state.

## Suggested fix

Detect whether the Istio API group/CRDs are present (e.g. via a discovery client check, similar to dependency-detection patterns elsewhere in the Kuadrant ecosystem) before registering the `EnvoyFilter` watch, and skip/defer EnvoyFilter-related reconciliation gracefully if they're absent, rather than panicking the whole manager.

Patryk-Stefanski:
Closes #1465

## Summary
- discover the Istio `EnvoyFilter` API resource before registering the typed watch
- register a safe `CustomResourceDefinition` lifecycle watch even when Istio is absent
- report `Ready=False` and stop MCPGatewayExtension reconciliation before creating Deployments, Services, HTTPRoutes, or EnvoyFilters when the CRD is unavailable
- when the CRD is missing at startup, cancel the manager after it becomes Established so the Kubernetes-managed controller restarts and registers the typed EnvoyFilter watch safely
- when an already-initialized CRD is deleted, keep the controller running, report `Ready=False`, and avoid normal dependent-resource updates until the CRD returns
- resume reconciliation when the CRD is recreated; no periodic discovery polling or custom dynamic EnvoyFilter watcher remains
- add CRD scheme registration, generated RBAC permissions, bundle RBAC metadata, discovery tests, and controller wiring tests

## Review guide
Review `internal/controller/istio_discovery.go` first, then the early availability guard and CRD lifecycle watch in `internal/controller/mcpgatewayextension_controller.go`, followed by `cmd/main.go`, generated RBAC manifests, and the bundle permission update.

## Verification
- `go test ./...` — 21 packages passed, 9 packages had no tests
- `make test-controller-integration` — 45/45 specs passed
- `make check` — CRDs, bundle manifests, and RBAC rules synchronized
- all shell blocks in the local Kind verification guide pass `bash -n`

The local Kind procedure is available in the untracked `kind-istio-envoyfilter-verification.md` file and is intentionally not part of the commit. It verifies startup with a missing CRD, safe restart when the CRD first appears, live EnvoyFilter events, and unchanged Deployment, Service, and HTTPRoute desired state while the CRD is unavailable.

### DIFF
--- a/api/v1/mcpgatewayextension_types.go
+++ b/api/v1/mcpgatewayextension_types.go
@@ -35,6 +35,8 @@ const (
 	ConditionReasonRefGrantRequired = "ReferenceGrantRequired"
 	// ConditionReasonDeploymentNotReady is the reason when the broker-router deployment is not ready
 	ConditionReasonDeploymentNotReady = "DeploymentNotReady"
+	// ConditionReasonIstioUnavailable is the reason when the Istio EnvoyFilter CRD is unavailable
+	ConditionReasonIstioUnavailable = "IstioEnvoyFilterUnavailable"
 
 	// ConditionReasonSecretNotFound is the reason when the trusted headers secret is missing
 	ConditionReasonSecretNotFound = "SecretNotFound"
--- a/bundle/manifests/mcp-gateway.clusterserviceversion.yaml
+++ b/bundle/manifests/mcp-gateway.clusterserviceversion.yaml
@@ -63,6 +63,14 @@ spec:
                 - update
                 - watch
             - apiGroups:
+                - apiextensions.k8s.io
+              resources:
+                - customresourcedefinitions
+              verbs:
+                - get
+                - list
+                - watch
+            - apiGroups:
                 - apps
               resources:
                 - deployments
--- a/charts/mcp-gateway/templates/rbac.yaml
+++ b/charts/mcp-gateway/templates/rbac.yaml
@@ -29,6 +29,14 @@ rules:
       - update
       - watch
   - apiGroups:
+      - apiextensions.k8s.io
+    resources:
+      - customresourcedefinitions
+    verbs:
+      - get
+      - list
+      - watch
+  - apiGroups:
       - apps
     resources:
       - deployments
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -19,6 +19,7 @@ This package contains the main of the mcp controller
 package main
 
 import (
+	"context"
 	"flag"
 	"log/slog"
 	"os"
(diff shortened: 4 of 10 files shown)


maleck13:
Without EnvoyFilter the gateway wont work. Is it a problem to panic in this situation? I guess if it is installed but not being used it would be annoying for it to flap around?

If there is no envoy filter api, I don't think the controller should do anything. Any resource created should just get a not ready status and a message of no envoy installed and exit. Don't create any deployments,httproutes etc

david-martin:
> Is it a problem to panic in this situation? I guess if it is installed but not being used it would be annoying for it to flap around

Yeah, the flapping/crashing constantly could be annoying. I'm +1 on doing something about that.

> If there is no envoy filter api, I don't think the controller should do anything

interesting. I think this is reasonable rather than partially provisioning an extension.
It would require moving the check earlier.

I'm +1 on this too.
It gets tricky if the crd disappears when somethings already provisioned. I was thinking the controller could panic and go back to the startup logic, but that could disrupt running reconciler logic outside the envoyfilter reconciler. however, if we agree nothing should be done if that crd is missing, then panicing to reset is possibly safer

david-martin (approved):

PR #1478, opened 2026-09-03
fix(controller): decouple config sync from extension readiness

### LINKED ISSUE #1446: MCPServerRegistration never re-reads a corrected credentialRef Secret while its MCPGatewayExtension is NotReady
Environment: v0.7.1 (`f3a5a32f`), cross-namespace `MCPGatewayExtension` (extension in a different namespace than its target `Gateway`, via `ReferenceGrant`), one upstream MCP server registered.

**What happened**

My `credentialRef` Secret ended up missing the `Bearer ` prefix the broker expects verbatim (`internal/broker/upstream/mcp.go:40-41`). The broker's upstream connection failed instantly with `authorization required` on every pod start. I fixed the Secret, but it never took effect — the broker kept failing until I patched the derived runtime config directly and restarted its pods.

**Why it doesn't self-heal**

`IsReady()` (`internal/broker/broker.go:538-550`) ties broker readiness to its upstream servers' health, so with one server configured, a bad credential fails `/readyz` and the Deployment goes `NotReady`. That flips the `MCPGatewayExtension` to `NotReady` too (`mcpgatewayextension_controller.go:236,245`), and a cross-namespace extension in that state gets excluded by `FindValidMCPGatewayExtsForGateway` (`mcpgatewayextension.go:100-107`). `MCPServerRegistration.Reconcile` then exits early on `len(mcpGatewayExtensions) == 0` (`mcpserverregistration_controller.go:201-211`) — before it ever reaches `buildMCPServerConfig`, the only place that rereads the `credentialRef` Secret (`mcpserverregistration_controller.go:434-458`).

So it's a closed loop: bad credential → unhealthy broker → `NotReady` extension → the one reconcile path that would pick up a corrected credential stops running. Fixing the Secret does nothing until something outside the controller breaks the loop.

(Same-namespace extensions skip this — they're included unconditionally without a Ready check, `mcpgatewayextension.go:96-99`.)

**Reproducing it**

I haven't run this exact sequence in kind, only hit the equivalent live on OpenShift, so flag anything that doesn't match:

1. `Gateway` in namespace B, `MCPGatewayExtension` in namespace A targeting it via `ReferenceGrant`, `MCPServerRegistration` in A with one upstream server, `credentialRef` → Secret S with a valid `"Bearer <token>"`.
2. Confirm both resources `Ready` and `tools/list` works.
3. `kubectl patch secret S -n A --type=merge -p '{"stringData":{"credential":"<token, no Bearer prefix>"}}'` — nothing validates this, so the next reconcile (extension still Ready) propagates it into the runtime config.
4. `kubectl rollout restart deployment/<broker> -n A` so it re-initializes against the bad credential and fails to authenticate.
5. `/readyz` returns 503, the Deployment goes `0/1`, the extension's `Ready` condition flips `False`, and the registration reports `"no valid mcpgatewayextensions configured"`.
6. Patch Secret S back to the valid value. Wait a few reconcile/check-interval cycles.
7. Diff the runtime config against Secret S — it's still stale. The broker stays unauthenticated indefinitely.
8. Restart the broker pod again without touching the runtime config — same stale credential, still broken. Confirms this isn't just a restart/caching issue.

**Suggested fix**

Sync `credentialRef` independent of the extension's Ready gate, so a corrected credential can reach the broker config even while the extension is recovering. Happy to send a PR if a maintainer agrees with the approach.

Related: #386 is adjacent (broker doesn't detect a live credential change without a config nudge) but that's about the broker's runtime cache, not this reconcile ordering.

rajatrsrivastav:
This PR resolves #1446 by fixing a fatal reconciliation deadlock in cross-namespace Kuadrant deployments.

### The Problem
Previously, a bad credential caused the broker to crash, turning the `MCPGatewayExtension` to `NotReady`. The `FindValidMCPGatewayExtsForGateway` function was explicitly filtering out cross-namespace extensions if they were not `Ready`. Because the extension was filtered out, `MCPServerRegistration` exited its reconciliation loop early and never reached `buildMCPServerConfig` to re-read the corrected Secret. This created a permanent deadlock.

### Changes Made
*   **Decoupled Config Sync from Readiness:** Modified `internal/controller/mcpgatewayextension.go` to remove the `mcpv1.ConditionTypeReady` requirement for cross-namespace extensions.
*   **Import Cleanup:** Removed the now-unused `"k8s.io/apimachinery/pkg/api/meta"` import.

### Why This Works
By removing the `Ready` gate, a bad credential will still accurately transition the extension to `NotReady`, but the reconciler will now successfully proceed to `buildMCPServerConfig`, sync the corrected Secret to the runtime config, and restart the broker—breaking the deadlock.

*Note: Existing security boundaries remain intact (ReferenceGrants are still strictly validated), and this change brings cross-namespace extensions to parity with same-namespace extensions, which already bypassed the `Ready` check.*

### Related Issues
Resolves [MCPServerRegistration never re-reads a corrected credentialRef Secret while its MCPGatewayExtension is NotReady #1446](#1446)

### Checklist
- [x] Configuration syncing is no longer blocked by extension readiness.
- [x] Cross-namespace `ReferenceGrant` validation remains fully enforced.
- [x] Code compiles and `go test` passes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Cross-namespace gateway extensions are now recognized as valid when an appropriate reference grant exists, even if their Ready status condition is not set to true.

### DIFF
--- a/internal/controller/mcpgatewayextension.go
+++ b/internal/controller/mcpgatewayextension.go
@@ -7,7 +7,6 @@ import (
 	"strings"
 
 	mcpv1 "github.com/Kuadrant/mcp-gateway/api/v1"
-	"k8s.io/apimachinery/pkg/api/meta"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
 	gatewayv1 "sigs.k8s.io/gateway-api/apis/v1"
@@ -102,7 +101,7 @@ func (r *MCPGatewayExtensionValidator) FindValidMCPGatewayExtsForGateway(ctx con
 			// we have to exit here
 			return validExtensions, fmt.Errorf("failed to check if mcpgatewayextension is valid %w", err)
 		}
-		if has && meta.IsStatusConditionTrue(mg.Status.Conditions, mcpv1.ConditionTypeReady) {
+		if has {
 			validExtensions = append(validExtensions, &mg)
 		}
 	}


maleck13 (commented):
this look reasonable to me please move to ready for review so we can have the tests run

rajatrsrivastav:
@maleck13 Thanks for reviewing! I've marked the PR as "Ready for review". 

All CI checks (including E2E and Conformance tests) have just finished running and passed successfully. Let me know if you need anything else!

maleck13 (approved):

maleck13:
thanks @rajatrsrivastav

PR #1469, opened 2026-09-02
fix some minor nits in the helm chart definitions

maleck13:
## What does this PR do?

- Removed dead broker.discoveryToolsEnabled and broker.discoveryToolThreshold from charts/mcp-gateway/values.yaml (never referenced in any template)
- Added restricted securityContext to the controller container in three deployment manifests:
  - charts/mcp-gateway/templates/deployment-controller.yaml
  - config/mcp-gateway/components/controller/deployment-controller.yaml
  - config/mcp-system/deployment-controller.yaml
- Added restricted securityContext to both Helm test pods:
  - charts/mcp-gateway/templates/tests/test-connection.yaml
  - charts/mcp-gateway/templates/tests/test-status.yaml

## Summary by CodeRabbit

- **Security**
  - Controller and test containers now run as non-root users with read-only root filesystems.
  - Privilege escalation is disabled, Linux capabilities are dropped, and the controller uses the default runtime seccomp profile for stronger isolation.

- **Configuration**
  - Removed broker settings for enabling discovery meta-tools and configuring the discovery tool threshold.
  - Added an optional setting to override the controller-discovered internal Gateway host.

PR #1482, opened 2026-09-07
Harden controller deployment securityContext

jjkiely:
## What

Add a `securityContext` to the mcp-gateway controller deployment:

- pod-level `runAsNonRoot: true`
- container-level `allowPrivilegeEscalation: false` and drop `ALL` capabilities

## Why

Other Kuadrant component controllers (e.g. dns-operator) already ship this hardening. Adding it here keeps the controllers consistent and lets the controller run under the OpenShift `restricted-v2` SCC without requesting extra privileges.

The chart is consumed downstream via the kuadrant-operator component-chart sync, so this is the correct source of truth for the change.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

-------------------------------------------------------------------------------------------------------------------------------------------------------------------------
The idea itself came from reviewing the downstream consolidation work: https://github.com/rh-api-management/rhcl-operator-product-build/pull/659 
Thanks @Patryk-Stefanski

### DIFF
--- a/charts/mcp-gateway/templates/deployment-controller.yaml
+++ b/charts/mcp-gateway/templates/deployment-controller.yaml
@@ -18,10 +18,17 @@ spec:
         {{- include "mcp-gateway.controllerSelectorLabels" . | nindent 8 }}
     spec:
       serviceAccountName: {{ include "mcp-gateway.controllerServiceAccountName" . }}
+      securityContext:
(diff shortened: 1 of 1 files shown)


jjkiely:
Looks like this is duplicating https://github.com/Kuadrant/mcp-gateway/pull/1469/changes

So closing this

PR #1489, opened 2026-09-08
fix: keep non-cacheable tools out of shared cache

### LINKED ISSUE #1488: [Bug] Duplicate tools from private 2026 upstreams
## Description
The v0.9.0 broker can return duplicate tools for a 2026-only upstream whose `tools/list` response declares `cacheScope: private` or `ttlMs: 0`.

The broker keeps the upstream tools in its stateless protocol cache and also performs the expected per-request fresh fetch, then appends the fresh tools without deduplicating or replacing the cached entries.

## Steps to Reproduce
1. Create a fresh Kind cluster.
2. Install mcp-gateway v0.9.0.
3. Deploy mcp-searxng 2.1.0 with `MCP_HTTP_STATELESS=true`.
4. Register it with a `search_` prefix.
5. Connect a 2026-07-28 MCP client and call `tools/list`.

## Expected Behavior
Each upstream tool appears once in the gateway response.

## Actual Behavior
The upstream returns four unique tools, but the gateway returns eight entries: the same four cached tools plus the same four freshly fetched tools.

Broker evidence from the reproduction:
```text
statelessTools=4
fetched user-specific tools toolCount=4
FilterTools called input_tools_count=8
```

## Environment
- mcp-gateway: v0.9.0
- mcp-searxng: 2.1.0
- protocol: 2026-07-28
- cluster: Kind

## Acceptance Criteria
- [ ] Private/zero-TTL 2026 upstream tools are returned once per `tools/list` response.
- [ ] Fresh per-request fetching remains enabled for upstreams that require it.
- [ ] A regression test covers the cached-plus-fresh merge path.
- [ ] Existing 2025 and 2026 protocol behavior remains unchanged.

## Additional Context
The upstream itself returns four unique tools and supports only 2026-07-28. The duplication is introduced by the gateway broker, not by the upstream server or router.

Patryk-Stefanski:
## Summary

- Keep 2026 tools with `cacheScope: private` or `ttlMs: 0` out of the shared broker tool cache.
- Continue fetching those tools fresh per request, as required for per-user data.
- Load the cached tool set and fresh-fetch server list from one cache snapshot so metadata transitions cannot combine stale cached tools with a fresh fetch.
- Preserve aggregate cache metadata and existing 2025 behavior.
- Add regression coverage for private scope, zero TTL, and cache-snapshot consistency.

Fixes #1488

## Why

The original failure was not that the broker needed to deduplicate two valid lists. Private and zero-TTL tools were incorrectly inserted into the shared cache even though they were already classified for fresh per-request fetching. A subsequent `tools/list` therefore started with a cached copy and appended a fresh copy.

The fix prevents those tools from entering the shared cache. Fresh fetching remains the source of truth for those servers.

## Scope

This PR addresses the tool-list caching issue only. Prompt and resource protocol behavior are intentionally unchanged.

## How to review

1. Review `rebuildProtocolCaches` in `internal/broker/protocol_filter.go` and confirm private/zero-TTL 2026 tools are excluded from `statelessTools` but their servers remain available for fresh fetching and cache metadata aggregation.
2. Review `toolsAndFreshFetchServersForProtocol` and the `tools/list` middleware path for the single-snapshot guarantee.
3. Review the regression tests in `protocol_filter_private_scope_test.go` and `protocol_filter_test.go`.

## Verification

- `go test ./internal/broker -run 'Test(RebuildProtocolCaches_ExcludesNonCacheableTools|ToolsAndFreshFetchServersForProtocol_UsesOneSnapshot)$' -count=1`
- `go test ./internal/broker/... -count=1`
- `git diff --check`

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved tool listings by checking servers requiring fresh data at request time.
  - Ensured tool listings use a consistent snapshot, reducing mismatches between available tools and server routing.
  - Restored expected handling for tools returned by multiple sources, preventing valid results from being inadvertently omitted.

- **Tests**
  - Added coverage for fresh-data handling, cache behavior, and consistent tool-list responses.

### DIFF
--- a/internal/broker/broker.go
+++ b/internal/broker/broker.go
@@ -451,14 +451,15 @@ func (m *mcpBrokerImpl) filteringMiddleware() mcp.Middleware {
 				}
 				isStateless := isStatelessProtocol(headers)
 
-				// filter by protocol version before user-specific fetches
-				toolsResult.Tools = m.toolsForProtocol(isStateless)
+				// load tools and fresh-fetch servers from one protocol-cache snapshot
+				var freshFetchServers []userSpecificServer
+				toolsResult.Tools, freshFetchServers = m.toolsAndFreshFetchServersForProtocol(isStateless)
 
 				var sessionID string
 				if s := req.GetSession(); s != nil {
 					sessionID = s.ID()
 				}
-				m.FetchUserSpecificTools(ctx, headers, toolsResult)
+				m.fetchUserSpecificTools(ctx, headers, toolsResult, freshFetchServers)
 
 				// collect cache metadata before FilterTools strips kuadrant/id
 				if isStateless {
--- a/internal/broker/protocol_filter.go
+++ b/internal/broker/protocol_filter.go
@@ -81,9 +81,44 @@ func (m *mcpBrokerImpl) rebuildProtocolCaches() {
 	// per-user servers with no prefix serve unroutable tools; exclude them
 	excluded := m.privateScopeServersWithoutPrefix()
 
+	// precompute 2026 servers needing per-request fetching (cacheScope:"private"
+	// or ttlMs:0). Their tools remain out of the shared cache and are merged
+	// into each request after a fresh upstream fetch.
+	userSpecific := make(map[config.UpstreamMCPID]bool, len(m.userSpecificServers))
+	for _, s := range m.userSpecificServers {
+		userSpecific[s.id] = true
+	}
+	freshToolServers := make(map[config.UpstreamMCPID]struct{})
+	var freshFetchServers []userSpecificServer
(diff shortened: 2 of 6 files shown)


maleck13 (dismissed):
Looks reasonable. Do we have the same issue with prompts?

maleck13:
@Patryk-Stefanski these tools shouldn't be in the cache. If a tool list or prompt list returns a cache scope private or a ttlms of 0 we should cache them. They are already fetched fresh as they are treated as per user servers when scope is private or ttlms is 0. So the bug is that they are cached at all

maleck13 (approved):

PR #1490, opened 2026-09-09
fix(broker): federate stateless upstreams on older 2025 protocol revisions 

### LINKED ISSUE #1451: Broker silently drops upstreams that negotiate an older 2025 revision or run stateless (no Mcp-Session-Id)
## Summary

The broker silently serves **no** tools/prompts for an upstream that (a) negotiates a **2025 protocol revision other than `2025-11-25`** (e.g. `2025-03-26`) and/or (b) runs the streamable-HTTP transport in **stateless** mode (no `Mcp-Session-Id`). Such an upstream connects successfully but is dropped from every served set with no error, and — because the broker treats any non-`2026` upstream as stateful — its session ping is rejected, de-federating it.

## Why it happens

- `rebuildProtocolCaches` partitions an upstream's tools into the stateful/stateless served set on an **exact** `Version2025` (`2025-11-25`) / `Version2026` (`2026-07-28`) match. An upstream advertising `2025-03-26` matches neither → served to no client.
- The broker requests `2025-11-25` at initialize; an upstream on an older MCP SDK down-negotiates to the newest revision *it* knows (e.g. `2025-03-26`), so this is the common case for any not-freshly-updated server.
- Statelessness is inferred from `protocol version >= 2026-07-28`, not from the transport. A session-less `2025` server therefore gets a session ping (rejected by strict stateless servers) and a session-keyed GET SSE watcher (nothing to watch).

## Reproduction

Stock [`baruchiro/paperless-mcp`](https://github.com/baruchiro/paperless-mcp) (stateless streamable-HTTP, `@modelcontextprotocol/sdk` 1.11 → negotiates `2025-03-26`): registers behind the broker, connects, and lists zero tools downstream. `mcp-searxng` 2.0.0 exhibits the same class of interop failure (cf. #1419).

## Proposed fix

Decouple "served" from "stateless":
- serve any `2025-` family upstream on the stateful route (consistently in `rebuildProtocolCaches`, `buildRoutingTable`, and `FetchUserSpecificTools` so tools list, route, and fetch);
- detect statelessness by transport (`session.ID() == ""`), skipping the session ping and GET SSE watcher and recycling the connection each health tick (as 2026 upstreams already do) so liveness/recovery is preserved.

A complete, unit-tested implementation (9 files, `go vet`/gofmt clean, broker + upstream suites green) is ready on a branch and can be opened as a PR once this is triaged:
https://github.com/rwlove/mcp-gateway/tree/feat/serve-stateless-2025-upstreams

tommyc2:
Closes #1451

## Changes

- serve any 2025-family upstream on the stateful route
(`serverServesOnStatefulRoute`), applied consistently in
`rebuildProtocolCaches`, `buildRoutingTable` and `FetchUserSpecificTools`.
- Widen `UsesStatelessProtocol` to mean "negotiated >= 2026-07-28 or no
server-assigned session ID
- Add standard tests to reflect changes
- session-less 2025 upstream skips the session ping and the GET SSE watcher
- `adjustTickerFromTTL` stays gated on the
negotiated version: a 2025 upstream sends no TTL hint, and widening there
would override the configured poll interval.
- PR Updated to reflect changes in #1485 

## Motivating case

A stock `baruchiro/paperless-mcp` (stateless streamable-HTTP, `@modelcontextprotocol/sdk` 1.11 → negotiates `2025-03-26`) currently connects but is served to no client. With this change its tools federate on the stateful route.

## Summary by CodeRabbit

* **Bug Fixes**
  * Upstreams using any 2025 protocol revision are now correctly routed through the stateful path.
  * Older 2025-compatible upstreams remain available for tools, prompts, and personalized tool results.
  * Sessionless connections now follow the correct notification and health-check behavior.
  * Sessionless upstreams no longer receive unsupported session pings.
  * Excluded upstreams are skipped before stateful route evaluation.
  * 2026-protocol upstream tools are refreshed according to their advertised cache timing.


tommyc2:
@rwlove recreated your PR as #1490 (originally #1450).

See code rabbit suggestions also + commit signature issue

internal/broker/protocol_filter.go
```diff
@@ -108,6 +140,7 @@ func (m *mcpBrokerImpl) rebuildProtocolCaches() {
 		}
 		serverID := config.UpstreamMCPID(serverIDStr)
 
+		if m.serverServesOnStatefulRoute(serverID) {
```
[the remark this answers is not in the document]
tommyc2: @coderabbitai this issue has been resolved.

internal/broker/upstream/manager.go
```diff
@@ -490,11 +493,14 @@ func (man *MCPManager) manage(ctx context.Context, event eventType) {
...
-	// closing the session, so the broker never learns notifications stopped.
-	// force a fresh connection on each health tick to re-establish the stream.
-	if event == eventTypeTimer && man.mcp.UsesStatelessProtocol() {
+	// Stateless upstreams are recycled on each health tick: a fresh Connect is
+	// the liveness+recovery signal for them (Ping is a no-op and there is no
+	// session whose loss the SDK reports), so a dead upstream is caught by the
+	// next Connect failure rather than lingering with stale tools. This covers
+	// both 2026 upstreams (which also rely on re-establishing the SDK's
+	// subscriptions/listen stream, silently dropped on restart) and session-less
+	// 2025 upstreams (stateless streamable-HTTP transport, no Mcp-Session-Id).
```
tommyc2: Reduce comment length here

tommyc2:
I would probably suggest adding an e2e to test all cases.

Also, I think `protocol_filter_stateful_route_test.go` can be merged into ` protocol_filter_test.go`, rather than creating a separate file

internal/broker/upstream/manager.go
maleck13: This needs a rebase as it is no longer the case

maleck13 (commented):
This needs  a rebase on main. The subscription listen and connection recycling is gone. I think we can fold this check fully into ```UsesStatelessProtocol()``` so it  means ``` negotiated >= 2026 || session.ID() == ""```

A quick scan of the code shows that the above should be fine:

| # | File:Line | Usage | Safe to widen? |
|---|-----------|-------|----------------|
| 1 | `mcp.go:327` | Debug log `"uses-stateless"` | Cosmetic — fine |
| 2 | `mcp.go:332` | Skip notification watcher (2026), check `session.ID() == ""` (session-less 2025), else start watcher | Yes — session-less 2025 hits the first branch, skips watcher correctly. Middle branch becomes dead code, collapse to two-way if. Fix log message. |
| 3 | `mcp.go:469` | Definition: `init.ProtocolVersion >= Version2026` | This is what you'd change |
| 4 | `mcp.go:505` | `Ping()` — if `UsesStatelessProtocol()`, return nil | Yes — session-less 2025 also can't be pinged |

rwlove:
The failing `Unit Tests` check (`TestMCPManager_manage_RecyclesSessionlessUpstream`) isn't the mock contract CodeRabbit flagged earlier — that suggestion predates the rebase and won't make it pass. The rebase onto main removed connection-recycling, so that test asserts behavior that no longer exists (`disconnectCount == 0`).

The real gap is that the session-less fold is only half-done: `UsesStatelessProtocol()` still checks version only, so a session-less 2025 upstream (negotiates e.g. `2025-03-26`, no `Mcp-Session-Id`) returns `false` and still gets the session-keyed GET SSE watcher + ping at `mcp.go:331`/`:495` — the exact case this PR is meant to fix.

I've opened a PR against this branch with the fix → tommyc2/mcp-gateway#1. It completes the fold: widens `UsesStatelessProtocol()` (`>= 2026 || session.ID() == ""`), drops the now-dead `IsSessionless()`, and replaces the obsolete recycle test with a regression assertion in the existing real-connect stateless test (which negotiates 2025, so it exercises the session branch). Verified locally green: `go build/vet/test ./...` (whole repo) + `-race` on the broker.

Merging that into `restore-serve-stateless-2025-upstreams` should clear the unit-test check. If you'd rather push it yourself, add me to the fork or grab the patch — whatever's least friction. Still open separately: the e2e all-cases test and merging `protocol_filter_stateful_route_test.go` into `protocol_filter_test.go` — happy to take those next.

tommyc2:
HI @rwlove,
Thanks for the contribution. I've already incorporated these changes and just have to push them here on this PR. I'll take a look at your PR on my fork to see if i can cherry pick anything. I appreciate the follow up!

tommyc2:
@coderabbit review recent changes in https://github.com/Kuadrant/mcp-gateway/pull/1490/commits/fc55f4a78470f8dcc0bbd2b41356cc0e1d7d3a35

tommyc2:
@coderabbit review

tommyc2:
@coderabbitai full review

internal/broker/protocol_filter.go
```diff
@@ -3,13 +3,43 @@ package broker
...
+// supportsStatefulRoute reports whether any of the given advertised protocol
+// versions belongs to the stateful (2025) family. The broker requests
+// Version2025 at initialize, but upstreams built on older MCP SDKs downgrade to
+// an earlier 2025 revision; their tools/list and tools/call payloads are
+// wire-compatible with what the broker serves downstream, so any 2025-series
+// revision is served on the stateful route rather than dropped from every
+// served set. Lock-free so it can be called from paths that already hold
+// mcpLock (e.g. buildRoutingTable).
+func supportsStatefulRoute(versions []string) bool {
+	for _, v := range versions {
```
maleck13: nit this could use the slices package: 

```go
 return slices.ContainsFunc(versions, func(v string) bool {
              return strings.HasPrefix(v, statefulVersionPrefix)
      })
```
keeps it consistent or at least somewhat consistent with the rest of the code base

internal/broker/user_specific_tools.go
```diff
@@ -80,7 +80,15 @@ func (broker *mcpBrokerImpl) fetchUserSpecificTools(ctx context.Context, headers
 	seen := make(map[config.UpstreamMCPID]bool, len(crdServers))
 	var matching []userSpecificServer
 	for _, srv := range crdServers {
-		if broker.ServerSupportsVersion(srv.id, clientVersion) {
+		// stateful clients are served by any upstream in the 2025 family;
+		// the stateless (2026) route stays an exact match.
```
maleck13, line 84: any reason here not to have a 2026 prefix match as versions beyond the current one are unlikely to revert to a stateful protocol

internal/broker/user_specific_tools.go
```diff
@@ -80,7 +80,18 @@ func (broker *mcpBrokerImpl) fetchUserSpecificTools(ctx context.Context, headers
...
+		var matches bool
+		if clientVersion == protocol.Version2026 {
+			matches = broker.ServerSupportsVersion(srv.id, protocol.Version2026)
+		} else {
+			matches = broker.serverServesOnStatefulRoute(srv.id)
+		}
+		if matches {
 			matching = append(matching, srv)
 		}
 		seen[srv.id] = true
```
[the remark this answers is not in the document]
maleck13: @tommyc2 seems a reasonable test to add?

internal/broker/routing_table.go
```diff
@@ -21,9 +21,9 @@ func (m *mcpBrokerImpl) buildRoutingTable() *routing.Table {
 			UserSpecificList:    cfg.UserSpecificList,
 			GuardrailsConfigIDs: cfg.GuardrailsConfigIDs,
 			// dual-protocol servers have both set to true; each router
-			// checks its own flag independently
+			// checks its own flag independently.
 			Stateless: up.SupportsVersion(protocol.Version2026),
```
maleck13, line 25: wondering again here if we should be a bit more future proof and add a prefix match for stateless version

maleck13 (commented):
Couple of comments. Generally though LGTM

- main issues is if we want to future proof 2026 with a prefix match?

internal/broker/routing_table.go
tommyc2: @maleck13 agreed, for this PR i focused mainly on the 2025 family not necessarily 2026 prefix matching. That can be a follow up as im conscious that this PR is becoming very large and i think its technically out of scope for issue #1451 . 

for this PR specifically, ill add the regression test. i can create a separate issue for 2026 checks.

maleck13: ok add a new issue for 2026 👍

maleck13 (approved):

PR #1498, opened 2026-09-11
docs: align OpenTelemetry documentation with implementation

david-martin:
## Summary

Aligns OpenTelemetry documentation with the current implementation.

- Updates routing spans, metrics, endpoints, and namespace guidance.
- Keeps the guide installation-independent.
- Makes the guide canonical; the local example now documents only stack setup and workflows.
- Removes stale tracing references without changing runtime code.

## Verification

- `git diff --check origin/main..HEAD`
- Confirmed the example no longer duplicates span or configuration reference sections.

### DIFF
--- a/.claude/otel-tracing-and-logging.md
+++ b/.claude/otel-tracing-and-logging.md
@@ -1,146 +1,68 @@
 # OpenTelemetry Tracing and Logging
 
-## Current Setup
+## Canonical references
 
-OTel is configured via environment variables. When no `OTEL_EXPORTER_OTLP_ENDPOINT` is set,
-OTel is fully disabled (zero overhead). The setup lives in `internal/otel/`:
+Keep the user-facing span, attribute, and metric reference in the
+[OpenTelemetry integration guide](../docs/guides/opentelemetry.md). Keep local
+stack setup and traffic examples in the [observability example](../examples/otel/README.md).
+Do not duplicate their tables here; update those documents when instrumentation
+changes.
 
-- `config.go` -- reads env vars, determines which signals are enabled
-- `otel.go` -- `SetupOTelSDK()` wires up propagator, tracer provider, logger provider, and shutdown
-- `provider.go` -- creates trace exporter (HTTP or gRPC based on URL scheme)
-- `logs.go` -- creates log exporter (same scheme dispatch)
-- `logging.go` -- `TracingHandler` injects `trace_id`/`span_id` into slog records; `MultiHandler` fans out to stdout + OTLP
-- `resource.go` -- shared resource with `service.name`, `service.version`, `vcs.revision`
+## Implementation map
 
-The global tracer provider is set in `cmd/mcp-broker-router/main.go` via `otel.SetTracerProvider()`.
-The logger is created with `NewTracingLogger()` which wraps slog with trace context injection.
+OTel setup lives in `internal/otel/`:
 
-## Tracing Conventions
+- `config.go` reads environment variables and determines which signals are enabled.
+- `otel.go` defines `SetupOTelSDK`, the broker tracer name, error helper, propagator, providers, and metrics handler.
+- `provider.go` creates the trace exporter, selecting HTTP or gRPC from the URL scheme.
+- `logs.go` creates the log exporter with the same scheme dispatch.
+- `logging.go` provides `TracingHandler` and `MultiHandler` for trace-correlated stdout and OTLP logs.
+- `resource.go` creates shared `service.name`, `service.version`, and VCS resource attributes.
 
-### Span Naming
-Spans are prefixed by component:
+The global tracer provider is set in `cmd/mcp-broker-router/main.go`. The logger
+is created with `NewTracingLogger()`.
 
-**Router** (`mcp-router.`):
-- `mcp-router.process` -- root span for the full ext_proc stream lifecycle
-- `mcp-router.route-decision` -- routing decision (tool-call, prompt-get, elicitation-response, or broker)
-- `mcp-router.tool-call` -- full tool call handling
-- `mcp-router.prompt-get` -- full prompt get handling
-- `mcp-router.elicitation-response` -- elicitation response routing
-- `mcp-router.broker-passthrough` -- pass-through to broker
-- `mcp-router.broker.get-server-info` -- broker lookup for tool server info
-- `mcp-router.broker.get-server-info-by-prompt` -- broker lookup for prompt server info
-- `mcp-router.session-cache.get` -- session cache read
-- `mcp-router.session-cache.store` -- session cache write
-- `mcp-router.session-init` -- hairpin initialize to backend
+## Agent-facing invariants
 
-**Broker** (`mcp-broker.`):
-- `mcp-broker.handle-request` -- wraps every MCP request to the broker
-- `mcp-broker.tools-list` -- tools/list filtering (authorization, virtual server, gateway metadata)
-- `mcp-broker.prompts-list` -- prompts/list filtering
-- `mcp-broker.upstream-manage` -- periodic backend health check and tool/prompt discovery
+- The router has separate stateful, body-based 2025-11-25 and stateless, header-based 2026-07-28 paths. See the [trace hierarchy](../docs/guides/opentelemetry.md#traces) before changing either path.
+- The broker publishes a cached `routing.RoutingTable`; router lookups happen inside route spans and do not create separate lookup spans.
+- Use `internaljwt.LogSafeSessionID` for session IDs in telemetry. Do not record raw session IDs in attributes or logs.
+- Follow the [OpenTelemetry MCP semantic conventions](https://opentelemetry.io/docs/specs/semconv/gen-ai/mcp/) before adding attributes. Keep protocol version fields distinct: incoming header, selected MCP protocol, and JSON-RPC envelope version are different values.
+- Trace context must be extracted and propagated through the OpenTelemetry propagator. Do not parse `traceparent` manually.
+- Metrics remain enabled independently of OTLP trace and log export.
 
-When adding new spans, follow the naming pattern: `<component>.<action>`.
+## Tracer and error conventions
 
-### Tracer Names
-Tracer names are defined as constants:
-- `mcpotel.BrokerTracerName` (`internal/otel/otel.go`) -- used by broker and upstream packages
-- `tracerName` (`internal/mcp-router/tracing.go`) -- router-local constant
+- Use `mcpotel.BrokerTracerName` for broker and upstream tracers.
+- Use the router-local `tracerName` constant in `internal/mcp-router/tracing.go`.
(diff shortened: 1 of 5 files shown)


.claude/otel-tracing-and-logging.md
```diff
@@ -1,146 +1,68 @@
 # OpenTelemetry Tracing and Logging
 
-## Current Setup
```
david-martin, line 3: removing duplicate content that's in docs. can be pulled in by agents from there as needed rather than duplicate

docs/guides/opentelemetry.md
```diff
@@ -1,134 +1,192 @@
...
 ```bash
-kubectl set env deployment/mcp-gateway -n mcp-system \
+kubectl set env deployment/mcp-gateway -n <namespace> \
   OTEL_EXPORTER_OTLP_TRACES_ENDPOINT="http://traces-collector:4318" \
   OTEL_EXPORTER_OTLP_LOGS_ENDPOINT="http://logs-collector:4318" \
   OTEL_EXPORTER_OTLP_INSECURE="true"
 ```
 
-To enable only traces (without log export), set only `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT`. To enable only log export, set only `OTEL_EXPORTER_OTLP_LOGS_ENDPOINT`.
+To enable only traces, set `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT` without a logs endpoint. To enable only log export, set `OTEL_EXPORTER_OTLP_LOGS_ENDPOINT` without a traces endpoint. Prometheus metrics do not use these endpoint variables.
```
Patryk-Stefanski: Please also unset `OTEL_EXPORTER_OTLP_ENDPOINT` for single-signal export. Both exporters fall back to it (`internal/otel/config.go:35–57`), so after Step 1 a traces-only override still exports logs. Unset the base endpoint and the other signal-specific endpoint before setting the desired one.

docs/guides/opentelemetry.md
```diff
@@ -1,134 +1,192 @@
...
-- `mcp.method.name` -- MCP method (initialize, tools/call, tools/list)
-- `gen_ai.operation.name` -- same as `mcp.method.name`
-- `gen_ai.tool.name` -- tool name (for tools/call requests)
-- `mcp.session.id` -- gateway session ID
-- `mcp.server` -- resolved backend server name
-- `mcp.route` -- routing decision (`tool-call`, `broker`, or `elicitation-response`)
-- `http.method`, `http.path`, `http.request_id`, `http.status_code`
-- `jsonrpc.request.id`, `jsonrpc.protocol.version`
-- `client.address` -- from x-forwarded-for header
+Attributes follow [OpenTelemetry MCP Semantic Conventions](https://opentelemetry.io/docs/specs/semconv/gen-ai/mcp/#server). Attributes below are recorded only on the listed span and when the source value exists. All router spans carry `component=mcp-router`.
```
Patryk-Stefanski: Header-derived attributes are set even when their headers are absent; their values are empty strings (`ext_proc_adapter.go:227–240`). Also, `mcp.response.protocol_version` copies the incoming request header, not the response version (`:512–516`). Remove the blanket presence guarantee and clarify these empty-value and source semantics in the table.

docs/guides/opentelemetry.md
```diff
@@ -1,134 +1,192 @@
...
+| 2026-07-28 `mcp-router.prompt-get` | `mcp.prompt.name`, `mcp.server`, and `mcp.server.hostname`; this stateless span does not set `mcp.session.id` |
+| `mcp-router.broker-passthrough` | `mcp.method.name` |
+
+#### Broker spans
+
+The MCP Broker emits spans for request handling, capability filtering, resource filtering, and upstream management:
+
+| Span | When | Attributes |
+|------|------|------------|
+| `mcp-broker.handle-request` | Every MCP request handled by the broker | `mcp.method` and `protocol.version`; `mcp.session.id` when the SDK request has a stateful session |
```
Patryk-Stefanski: This span starts in the SDK receiving middleware, not at the broker HTTP entry point. The legacy compatibility handler answers requests such as `server/discover` before SDK dispatch (`http_compat.go:239–245`), so those responses emit no broker handler span. Replace “Every MCP request” with “MCP requests dispatched through the SDK receiving middleware” and note that compatibility-layer rejections bypass it.

examples/otel/README.md
```diff
@@ -348,18 +266,38 @@ echo "Search for trace: $TRACE_ID"
...
+4. Select the provisioned **Prometheus** datasource and choose a `server_name` when
+   inspecting per-upstream panels.
+5. The **Tools Discovered per Server** panel reads `mcp_broker_tools_discovered`;
+   see the [OpenTelemetry integration guide](../../docs/guides/opentelemetry.md#broker-metrics)
+   for the metric definition and behavior.
+
+To inspect the raw gateway metrics without Grafana, use a separate port-forward:
+
+```bash
+kubectl port-forward -n mcp-system deployment/mcp-gateway 9090:9090
```
Patryk-Stefanski, line 303: `kubectl port-forward` stays in the foreground, so the following `curl` cannot run in the same terminal while the tunnel is active. Split this into two code blocks: keep the port-forward running in one terminal and run `curl` in a second.

Patryk-Stefanski (commented):
Documentation accuracy and workflow comments.

docs/guides/opentelemetry.md
david-martin: Makes sense if OTEL_EXPORTER_OTLP_ENDPOINT is set.
Updating...

docs/guides/opentelemetry.md
david-martin: Will update wording to make this clear.

docs/guides/opentelemetry.md
david-martin: > MCP requests dispatched through the SDK receiving middleware

This sounds too vague (what does this mean in terms of mcp requests) and exposes implementation detail.
I'm inclined to keep it more like the existing wording, with specific methods as examples
e.g.

> MCP method requests processed by the broker, such as `initialize`, `ping`, `tools/list`, `tools/call`, `prompts/list`, `prompts/get`, and `resources/list`

david-martin:
@Patryk-Stefanski feedback addressed

and 1 small change to the Istio CR to keep the shell of it for CI purposes

docs/guides/a2a-passthrough.md
```diff
@@ -235,4 +235,4 @@ kubectl -n gateway-system logs deployment/mcp-gateway-istio --tail=50 | grep wea
 ## Next steps
 
 - [Authorization](./authorization.md) — the per-capability authorization pattern this builds on
-- [Auditing MCP Tool Calls](./auditing.md) — the Istio Telemetry access-log approach in depth
+- [Auditing MCP Tool Calls](./auditing.md) — the router structured audit log
```
Patryk-Stefanski: Step 4 still tells readers to define `a2a-access-log-provider` from the Auditing guide, but this patch deletes that provider procedure and the `mcp-json-access-log` provider. The new link therefore points to a guide that cannot complete the A2A audit workflow. Keep the provider setup self-contained here, or remove/rewrite the provider-based Step 4.

docs/guides/opentelemetry.md
```diff
@@ -1,134 +1,210 @@
 # OpenTelemetry Integration
 
-This guide covers enabling OpenTelemetry (OTel) on the MCP Gateway for distributed tracing, log export, and Prometheus metrics. Tracing and log export require an OTLP endpoint to be configured. Prometheus metrics are always enabled and require no configuration.
+This guide covers enabling OpenTelemetry (OTel) on the MCP Gateway for distributed tracing, log export, and Prometheus metrics. Tracing and log export require a base or signal-specific OTLP endpoint. Prometheus metrics are always enabled and require no OTLP endpoint.
```
Patryk-Stefanski: `Prometheus metrics are always enabled` is not true for an invalid or unsupported OTLP endpoint: `SetupOTelSDK` returns before `NewMetricsProvider` (`internal/otel/otel.go:45-67`), and `/metrics` is then served as 503 (`cmd/mcp-broker-router/broker.go:109-116`). Initialize metrics independently, or qualify this claim.

docs/guides/opentelemetry.md
```diff
@@ -1,134 +1,210 @@
...
-- `jsonrpc.request.id`, `jsonrpc.protocol.version`
-- `client.address` -- from x-forwarded-for header
+The router propagates W3C trace context on forwarded requests. The broker extracts that context, so router and broker work can be correlated in one trace. Broker filtering spans are children of `mcp-broker.handle-request`, not direct children of router route spans.
 
-On error, spans include `error.type`, `error_source`, and `http.status_code`.
+#### Error attributes
+
+Error recording is component-specific. An error span does not automatically contain every error attribute:
+
+- ext_proc processing errors record `error.type`, `error_source=ext-proc`, and `http.status_code`.
```
Patryk-Stefanski: `ext_proc processing errors` is too broad: several `stream.Send` error paths return without calling `recordError` (for example, `internal/mcp-router/ext_proc_adapter.go:298-305` and `341-370`). Say “errors passed to `recordError`” or describe the covered paths.

docs/guides/opentelemetry.md
```diff
@@ -1,134 +1,210 @@
...
+- Broker handler errors record `error.type` and `error_source=broker`.
+- Backend connection or discovery errors recorded by `mcp-broker.upstream-manage` record `error.type`, `error_source=backend`, and `mcp.server`.
+- Router route errors can set span status and `error.type`. The resulting HTTP status is recorded on `mcp-router.process` when response headers are processed.
 
 ### Logs
 
-When log export is enabled, all `slog` log lines are sent to the collector via OTLP in addition to stdout. Log lines emitted within a traced request automatically include `trace_id` and `span_id` fields, enabling log-to-trace correlation in backends like Grafana (Loki to Tempo).
+When log export is enabled, all `slog` log lines are sent to the collector through OTLP in addition to stdout. Log lines emitted within a traced request automatically include `trace_id` and `span_id` fields, enabling log-to-trace correlation in backends such as Grafana (Loki to Tempo).
```
Patryk-Stefanski: `all slog log lines` overstates coverage. The process starts with a stdout-only logger and installs the OTLP handler only after `SetupOTelSDK` returns (`cmd/mcp-broker-router/main.go:181-196`), so setup/startup records are not exported. Say “subsequent log lines.”

david-martin: Updated wording

docs/guides/opentelemetry.md
david-martin: Updated wording, but avoiding giving implementation detail

docs/guides/opentelemetry.md
david-martin: Updated wording

Patryk-Stefanski (approved):

Issue #1499, opened 2026-09-11
Router trusts unverified JWT claims for identity and audit logs


david-martin:
## Summary

PR #1116 fixed the broker's raw JWT parsing for URL-elicitation identity binding, but the router still derives `x-mcp-verified-sub` by decoding the client-supplied `Authorization` JWT. `ExtractSubClaim` does not verify the signature; it assumes AuthPolicy ran before ext_proc.

Without AuthPolicy, an unsigned JWT can choose the audit identity and internal verified-sub value.

## Reproduction

In a local `make local-env-setup` environment with no AuthPolicy, initialize a session and set `SESSION_ID` to its `Mcp-Session-Id`:

```bash
curl -sS -X POST 'http://mcp.127-0-0-1.sslip.io:8001/mcp' \
  -H 'Authorization: Bearer eyJhbGciOiJub25lIn0.eyJzdWIiOiJ2aWN0aW0ifQ.' \
  -H "Mcp-Session-Id: $SESSION_ID" \
  -H 'Content-Type: application/json' \
  -H 'Accept: application/json, text/event-stream' \
  -d '{"jsonrpc":"2.0","id":3,"method":"tools/call","params":{"name":"everything_echo","arguments":{"message":"spoof-check"}}}'
```

The call returns HTTP 200:

```text
data: {"result":{"content":[{"type":"text","text":"Echo: spoof-check"}]},"jsonrpc":"2.0","id":3}
```

The router records the forged identity:

```text
level=INFO msg="tool call" component=router audit=true user=victim tool=echo status=200 request_id=<uuid> session=jti:<uuid>
```

## Impact

An unauthenticated client can make audit records attribute activity to an arbitrary subject. The same trust assumption also leaves URL-elicitation identity binding dependent on AuthPolicy rather than enforcing the boundary in the router.

## Proposed fix

Source identity only from AuthPolicy/Envoy-verified metadata, or require AuthPolicy on every route that consumes `x-mcp-verified-sub`. If no verified identity exists, do not set the header or an authenticated audit user. Add an end-to-end regression test with a forged Authorization header and no AuthPolicy.

Related: #1083, #1116

david-martin:
Discussed on backlog refinement 15th Sep 2026 - this should be a known limitation i.e. the mcp gateway does not verify the jwt, and it's up to something else i.e authorino, to do that verification.

PR #1508, opened 2026-09-14
docs: clarify audit identity requirements

Patryk-Stefanski:
Relates to #1499

Document that audit `user` comes from JWT `sub` but is meaningful only after AuthPolicy validation. Without AuthPolicy, callers can choose `sub`; do not use `user` as identity evidence.

## Summary by CodeRabbit

- **Documentation**
  - Updated the auditing guide to clarify how the `user` field is populated after authentication.
  - Removed outdated references to the previous identity header and direct JWT claim extraction.


david-martin (approved):
This feels like the right approach rather than trying to validate the header in the mcp gateway.
Auth identity is decoupled from the mcp-gateway.

Issue #1301, opened 2026-07-24
Add NeMo guardrails integration


david-martin:
Allow NeMo guardrails API calls to be made per MCP tool call request. This should be configurable via the existing MCP gateway CRD(s). Proposal to follow first on approach.

See also https://redhat.atlassian.net/browse/CONNLINK-1378

christinaexyou:
hi @maleck13, can I please be assigned to this issue ?

david-martin:
@christinaexyou assigned. 
We're also looking at removing the restriction on requiring an assigned issue to open a PR, so the PR open/close flip flopping shouldn't be an issue next time

david-martin:
PRs so far:

* #1302 
* #1370 
* #1409 
* #1438

maleck13:
@christinaexyou after #1438 merges what remains here?

christinaexyou:
@maleck13 response-side tools/call checks (#1439, #1440), closing the guardrails per-server loop (#1507), and docs are still remaining. i'll open up issues for the latter two

PR #1514, opened 2026-09-16
feat(guardrails): 2025 tools/call responses

### LINKED ISSUE #1440: feat(guardrails): add response-side tools/call for 2025 SSE/Event-Stream
Add response-side guardrails for `2025-11-25` tool-call responses that return text/event-stream. 

Sibling issue to #1439. Depends https://github.com/Kuadrant/mcp-gateway/issues/1403. Parent issue: https://github.com/Kuadrant/mcp-gateway/pull/1302.

christinaexyou:
## What does this PR do?

Response-side guardrails for tools/call on the `2025-11-25` protocol path. Previously, guardrails only inspected requests before they went upstream. This PR enables the guardrails to also inspect the upstream response before it reaches the client.

Fixes #1440, parent issue #1301

## Pre-review checklist

## Summary by CodeRabbit

- **New Features**
  - Added response guardrails for tool-call results in JSON and SSE formats.
  - Allowed responses pass through unchanged; blocked responses return safe tool errors.
  - Modified responses can be redacted and returned as successful results.
  - Elicitation events continue forwarding immediately during guarded responses.
  - Response buffering respects configured body-size limits.

- **Bug Fixes**
  - Prevented guarded content from being truncated before inspection.
  - Preserved streaming when buffering is unnecessary.
  - Corrected response-size enforcement for listener traffic.
  - Improved handling of multiline and incomplete SSE events.


internal/mcp-router/ext_proc_adapter.go
```diff
@@ -139,11 +139,15 @@ func responseDecisionToResponse(d *routing.ResponseDecision) []*extProcV3.Proces
 	responses := rb.WithResponseHeaderResponse(headers).Build()
 
 	if d.StreamBody && len(responses) > 0 {
+		bodyMode := extprochttp.ProcessingMode_STREAMED
+		if d.BufferResponseBody {
+			bodyMode = extprochttp.ProcessingMode_BUFFERED
```
Patryk-Stefanski: Buffering the whole SSE response stalls `elicitation/create`: the upstream waits for the client's reply, but Envoy withholds the event until the response ends. Reproduced with a minimal Envoy 1.39.1 fixture: `STREAMED` completes; `BUFFERED` stalls. Forward elicitation/control events and check the tool result before releasing it.

internal/routing/response.go
```diff
@@ -120,9 +125,16 @@ func (h *ResponseHandler202511) HandleResponse(ctx context.Context, input *Respo
...
 	// resource URI rewriting - either gate is sufficient on its own, tool calls
 	// to servers with no prefix and no elicitation stay pass-through
-	if req != nil && req.IsToolCall() && input.StatusCode == strconv.Itoa(http.StatusOK) &&
-		(req.ClientElicitation || req.ServerPrefix != "") {
-		decision.StreamBody = true
+	if req != nil && req.IsToolCall() && input.StatusCode == strconv.Itoa(http.StatusOK) {
+		if req.ClientElicitation || req.ServerPrefix != "" {
+			decision.StreamBody = true
+		}
+		if len(req.GuardrailsConfigIDs) > 0 {
```
Patryk-Stefanski: This checks only per-server IDs, so global-only policies never activate response guardrails. Activate inspection when either global or per-server IDs are configured. The checker already merges both sets.

internal/mcp-router/response_body_extract.go
```diff
@@ -0,0 +1,72 @@
...
+)
+
+// extractToolResponseText extracts and concatenates the text from all
+// type:"text" content items in a tools/call result body. The body may be
+// plain JSON or SSE. Returns nil when no text content is found.
+func extractToolResponseText(body []byte) []byte {
+	var texts []string
+	remaining := body
+	for len(remaining) > 0 {
+		idx := bytes.IndexByte(remaining, '\n')
```
Patryk-Stefanski, line 29: Decoding physical lines skips valid pretty-printed JSON and SSE events split across multiple `data:` fields. Extraction returns `nil`, which skips guardrails and forwards the original response. Decode complete JSON documents or assembled SSE events.

internal/mcp-router/response_body_extract.go
```diff
@@ -0,0 +1,72 @@
...
+
+func extractTextFromResultJSON(data []byte) []string {
+	// reuse jsonRPCMessage from elicitation.go
+	var msg jsonRPCMessage
+	if err := json.Unmarshal(data, &msg); err != nil || msg.Method != "" || len(msg.Result) == 0 {
+		// skip requests (have Method), notifications, and error responses (no Result)
+		return nil
+	}
+	// fast path: skip the second unmarshal when there are no text-type items
+	if !bytes.Contains(msg.Result, []byte(`"text"`)) {
```
Patryk-Stefanski: `{"type":"\u0074ext","\u0074ext":"secret"}` is valid text content, but the raw substring check misses it and skips guardrails. Both the type value and property name are escaped. Remove the shortcut and inspect decoded JSON.

internal/mcp-router/ext_proc_adapter.go
```diff
@@ -577,14 +588,37 @@ func (s *ExtProcServer) Process(stream extProcV3.ExternalProcessor_ProcessServer
...
+				text := extractToolResponseText(body)
+				replacement := routing.CheckToolResponseGuardrails(
+					ctx,
+					s.RoutingConfig.Load(),
+					mcpRequest.ToolName(),
+					mcpRequest.GuardrailsConfigIDs,
+					text,
+					mcpRequest.ID,
+					s.Logger,
+					routing.BuildSSEToolError,
```
Patryk-Stefanski: Blocked or modified `application/json` responses get an SSE body without a matching `Content-Type` change, so clients cannot parse them. Select JSON or SSE replacement builders using the response media type.

internal/controller/mcpgatewayextension_controller.go
```diff
@@ -768,13 +768,21 @@ func (r *MCPGatewayExtensionReconciler) enqueueMCPGatewayExtForReferenceGrant(ct
...
 	// build the ext_proc filter config as a structpb.Struct
 	extProcConfig, err := structpb.NewStruct(map[string]any{
 		"name": "envoy.filters.http.ext_proc",
 		"typed_config": map[string]any{
 			"@type":               "type.googleapis.com/envoy.extensions.filters.http.ext_proc.v3.ExternalProcessor",
 			"failure_mode_allow":  false,
 			"allow_mode_override": true,
+			// max_request_bytes bounds BUFFERED mode body buffering.
+			// must match maxBodyBytes so guardrails response checks are not silently truncated.
+			"max_request_bytes": maxBodyBytes,
```
Patryk-Stefanski: `max_request_bytes` is not an ext_proc field. Istio normally discards it, so this does not configure Envoy's response buffer limit. Use a supported buffer setting; `spec.maxBodyBytes` still applies to existing application-side checks.

Patryk-Stefanski (changes requested):
Requesting changes for the response guardrails activation/parsing bypasses, in-band elicitation stall, replacement media type, and ineffective Envoy buffer setting described inline.

Verification against `83bd092d5185`: isolated Go probes using the PR's extractor, builders and checker, plus pinned Envoy protobufs; a real Envoy 1.39.1 MCP/ext_proc reproduction comparing streamed and buffered elicitation. This was focused verification, not a full Kubernetes deployment.

internal/mcp-router/ext_proc_adapter.go
christinaexyou: ack - the controller now forward elicitations immediately. added a new file (+ test file), `internal/mcp-router/guardrails_response_buffer.go` to handle SSE parsing and wired it back into `internal/mcp-router/ext_proc_adapter.go`

internal/routing/response.go
christinaexyou: added new functions, `configured` and `GuardrailsConfigured` in `guardrails.go` to verify whether either per-server IDs and global-only policies are configured.

internal/mcp-router/response_body_extract.go
christinaexyou: tackled both via `internal/mcp-router/guardrails_response_buffer.go` we check if data is JSON-encoded and everything else is treated as SSE events.

internal/mcp-router/response_body_extract.go
christinaexyou: done - removed

internal/controller/mcpgatewayextension_controller.go
christinaexyou: done -  patched Listener.PerConnectionBufferLimitBytes with a second EnvoyFilter, carrying the same maxBodyBytes value. however, i'm unsure how kosher this is since it creates another resource - all existing tests pass but do let me know if there is a better way of implementing this

internal/mcp-router/guardrails_response_buffer.go
```diff
@@ -0,0 +1,186 @@
...
+		return false
+	}
+	if msg.Method != "" {
+		return false // request or notification, not a response
+	}
+	if len(msg.Result) == 0 && len(msg.Error) == 0 {
+		return false
+	}
+	return idsEqual(msg.ID, g.requestID)
+}
```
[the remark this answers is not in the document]
christinaexyou: added a `respId` return var to `internal/mcp-router/guardrails_response_buffer.go` to track responses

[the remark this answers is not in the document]
christinaexyou: check again

internal/mcp-router/guardrails_response_buffer.go
```diff
@@ -0,0 +1,262 @@
...
+func (g *guardrailsResponseBuffer) withLimit(maxBytes int, oversized []byte) *guardrailsResponseBuffer {
+	g.maxBytes = maxBytes
+	g.oversized = oversized
+	return g
+}
+
+// Process appends chunk and returns any bytes now safe to forward.
+func (g *guardrailsResponseBuffer) Process(ctx context.Context, chunk []byte) []byte {
+	if g.done {
+		return chunk
```
david-martin: could this be called after guardrails ran on some previous chunks, and result in unchecked chunks being returned to the client?

internal/mcp-router/guardrails_response_buffer.go
```diff
@@ -0,0 +1,262 @@
...
+// overLimit reports whether currently buffered bytes exceed maxBytes.
+func (g *guardrailsResponseBuffer) overLimit() bool {
+	return g.maxBytes > 0 && len(g.unconsumed)+len(g.eventBytes) > g.maxBytes
+}
+
+// reject discards all buffered bytes and returns oversized in their place,
+// marking the buffer done so any further chunks in this response stream pass
+// through unchecked rather than continuing to accumulate.
+func (g *guardrailsResponseBuffer) reject() []byte {
+	g.done = true
```
david-martin, line 162: same question for oversized responses, will future chunks pass through because g.done is true from here on?

internal/mcp-router/guardrails_response_buffer.go
christinaexyou: ack thanks for flagging - updated s.t we discard any unchecked chuncks

Patryk-Stefanski:
Rebased onto the latest `main` (`cf26aeb`, including #1522) to remove branch drift and ensure the fixes are reviewed against the current base. The updated branch keeps response guardrails compatible with in-band elicitation: control events remain streamed while terminal tool results are checked. It also includes the global-only activation, JSON/SSE parsing and encoding, and supported listener buffer-limit fixes.

internal/mcp-router/guardrails_response_buffer.go
christinaexyou: yes - mitigated by discarding any unchecked chunks

Patryk-Stefanski:
@david-martin @CraigBrooks — I grouped the two findings that may depend on upstream behavior:

1. `Content-Type: text/event-stream` with a bare JSON-RPC body. The gateway detects raw JSON for parsing but still selects the SSE replacement path; the modified response reproduced as an empty HTTP 500.
2. An SSE response ending at EOF without the blank-line event delimiter. The gateway flush path skips terminal ID normalization; an upstream `id: 999` reached the client instead of the request ID.

Both are malformed/non-compliant upstream cases. Should the gateway reject or normalize these defensively, or should we treat them as MCP server misconfiguration and out of scope? The JSON-RPC error and CR-only SSE findings are separate and involve valid protocol behavior.

Patryk-Stefanski (approved):

PR #1515, opened 2026-09-16
feat: override upstream protocol versions via MCPServerRegistration

### LINKED ISSUE #1483: Broker fetches tools from a 2026-07-28 stateless upstream but never surfaces them in tools/list
## Summary

On mcp-gateway **0.9.0**, an upstream that negotiates MCP **`2026-07-28`** in **stateless** mode connects successfully and the broker fetches its tools into internal state (`updating gateway tools adding=N`, `internal tools total=N`, stable across health-check cycles) — but **none of those tools appear in the aggregated `tools/list` served to clients**. Upstreams that negotiate `2025-11-25` surface their tools normally through the same broker.

Net effect: the upstream is effectively de-federated from a client's perspective, while broker logs look healthy.

## Environment

- **Broker:** `ghcr.io/kuadrant/mcp-gateway:v0.9.0`, single replica, `--mcp-check-interval=60`.
- **Affected upstream:** `docker.io/isokoliuk/mcp-searxng:2.1.0` (`MCP_HTTP_HOST=0.0.0.0`). Registered via `MCPServerRegistration` with `prefix: search_`.
- The broker connects directly to the upstream service over plain HTTP (no mesh/mTLS on this hop).
- **Reproduces regardless of the upstream's session mode** — with `MCP_HTTP_STATELESS=true` (server accepts session-less requests) *and* with it unset (server requires a session / returns `400` to session-less requests). In both cases the broker reports `negotiated-protocol=2026-07-28 uses-stateless=true` and serves 0 of its tools.

## What the broker debug logs show (`--log-level=-4`)

```
msg="upstream connected"  negotiated-protocol=2026-07-28  uses-stateless=true
msg="using subscriptions/listen for notifications (2026 upstream)"
msg="fetching tools"
msg="updating gateway tools"  adding=4  removing=0
msg="internal tools"  total=4
# next health-check tick (+60s): recycles the stateless connection, reconnects,
# adding=0 removing=0, total=4  — i.e. the 4 tools remain in internal state, stable
```

So the broker successfully lists 4 tools from the upstream and holds them internally.

## What the client sees

`initialize` + `tools/list` against the broker returns the full aggregate (all other upstreams present) but **zero** of this upstream's 4 tools (`searxng_web_search`, `searxng_search_suggestions`, `searxng_instance_info`, `web_url_read`) — verified by exact name. Not truncation (full response body), not client-protocol-dependent (tried `2026-07-28`, `2025-11-25`, `2025-06-18` — all omit them).

## The discriminator

In a cluster of ~17 enabled upstreams, this is the **only** one negotiating `2026-07-28` + `uses-stateless=true`, and the **only** one taking the `using subscriptions/listen for notifications (2026 upstream)` path. Every other upstream negotiates `2025-11-25` (`uses-stateless=false`) and its tools are served correctly. That strongly localizes the bug to the broker's handling of the **2026-07-28 stateless / subscriptions-listen** code path: tools are fetched into internal state but not published to the served tool set.

## Expected

Tools fetched from a healthy `2026-07-28` stateless upstream (`adding=N`, `total=N`) should appear in the client-facing `tools/list`, same as for `2025-11-25` upstreams.

## Minimal repro

1. Deploy broker 0.9.0.
2. Deploy `isokoliuk/mcp-searxng:2.1.0` with `MCP_HTTP_STATELESS=true`, `MCP_HTTP_HOST=0.0.0.0`; register it (any `prefix`).
3. `initialize` + `tools/list` against the broker → the searxng tools are absent, despite broker debug logs showing `adding=4 / total=4`.

Happy to provide the full debug log or test a patch.

rwlove:
## Problem

The broker classifies an upstream's tools by the versions it reports in
`server/discover`. Some MCP SDKs (e.g. the TS SDK v2) advertise only modern
revisions there by design, even though the server still serves 2025-era
requests. Such an upstream is federated only to 2026 clients and is invisible
to every 2025 client — including Claude Code and mcp-inspector's default.

Verified against a live searxng upstream (`MCP_HTTP_STATELESS=true`): it
advertises `["2026-07-28"]` in `server/discover`, yet a session-less request at
`2025-11-25` returns `initialize` → 200, `tools/list` → 200, and `tools/call`
→ 200 (executes). It is a dual-era server that the gateway de-federates for
2025 clients.

## Change

Adds an optional `spec.supportedProtocolVersions` to `MCPServerRegistration`.
When set, the broker treats the upstream as supporting exactly those versions
(the negotiated version is always included) rather than trusting
`server/discover`. This feeds the same `supportsStatefulRoute` /
`ServerSupportsVersion` classification #1490 added, so the upstream's tools are
served on each listed version. Unset preserves today's behavior.

Follows the CRD-override direction discussed in #1483.

## Notes

- Only versions the upstream genuinely serves should be listed — a client calls
  a listed tool over its own negotiated version.
- Version resolution is extracted into a pure `resolveSupportedVersions` helper
  with unit tests; docs added to `multi-protocol-support.md`.

Closes #1483

## Summary by CodeRabbit

- **New Features**
  - Added support for configuring protocol versions supported by registered MCP servers.
  - Configured versions take precedence over automatically discovered versions, while the negotiated initialization version remains included.
  - Protocol version settings support up to 10 date-formatted entries.
  - Calls using versions unsupported by the upstream server may fail.

- **Documentation**
  - Added guidance on configuring protocol version overrides, their limitations, and their effects on server classification and tool federation.


docs/guides/multi-protocol-support.md
```diff
@@ -17,12 +17,45 @@ When a client connects, the gateway detects the protocol version:
...
 | Both | `["2025-11-25", "2026-07-28"]` | 2026-07-28 (highest) |
 
+### Overriding an upstream's advertised versions
+
+The gateway learns which versions an upstream serves from that upstream's
+`server/discover` response. Some MCP SDKs advertise **only** modern (2026-era)
+revisions there, even though the server still serves legacy (2025-era) requests.
+Such an upstream is classified as 2026-only and its tools are federated only to
+2026 clients — invisible to the entire 2025 client base — despite serving those
+clients correctly.
```
david-martin: I don't like this, but I understand it.
I'd prefer wording more along the lines of 'Some MCP servers may not advertise supported protocols correctly...' rather than calling out the cause of this particular case.
This makes it clear there is a problem with the MCP server implementation, and we're trying to accomodate that at the gateway.

docs/guides/multi-protocol-support.md
```diff
@@ -17,12 +17,45 @@ When a client connects, the gateway detects the protocol version:
...
+  # This upstream serves both eras but advertises only 2026-07-28 in
+  # server/discover; declare both so 2025 clients can reach its tools.
+  supportedProtocolVersions:
+    - "2025-11-25"
+    - "2026-07-28"
+```
+
+Only list versions the upstream genuinely serves: a client calls a listed tool
+over its own negotiated version, so declaring a version the upstream does not
+serve will cause those calls to fail.
```
david-martin, line 52: Let's extend this caveat to include wording that setting this field does not guarantee any future protocol discovery or compatibility. 
Use it only as a last resort sort of thing

david-martin (approved):
Thanks @rwlove

